PR #49, opened 2023-03-14
feat: DataLoader Impl

amitksingh1490:
[Draft]


runtime/src/test/scala/tailcall/runtime/ConfigSpec.scala
```diff
@@ -143,6 +144,13 @@ object ConfigSpec extends ZIOSpecDefault {
           val expected =
             """{"post":{"title":"sunt aut facere repellat provident occaecati excepturi optio reprehenderit","user":{"name":"Leanne Graham"}}}"""
           assertZIO(program)(equalTo(expected))
+        },
+        test("posts user") {
+          val program = for {
+            _   <- execute(JsonPlaceholderConfig.config)(""" query {posts { title user { name } } }""")
+            opt <- TestConsole.output
+          } yield opt.count(_.contains("Request"))
+          assertZIO(program)(equalTo(11))
```
tusharmath, line 153: We can drop this test. Instead have a test for DataLoader which checks if the side effect is executed only once.

PR #108, opened 2023-04-22
feat: global request timeout support

amitksingh1490:


runtime/src/main/scala/tailcall/runtime/model/Blueprint.scala
```diff
@@ -29,7 +29,10 @@ import scala.annotation.tailrec
  * is clearly defined. Once the IR is ready we will directly
  * compile IR to Caliban's Step ADT.
  */
-final case class Blueprint(definitions: List[Blueprint.Definition] = Nil) {
+final case class Blueprint(
+  definitions: List[Blueprint.Definition] = Nil,
+  server: Blueprint.Server = Blueprint.Server(),
```
tusharmath, line 34:
```suggestion
  definitions: List[Blueprint.Definition],
  server: Blueprint.Server,
```

tusharmath: If we remove the default values, we won't need unit tests.

server/src/main/scala/tailcall/server/GenericServer.scala
```diff
@@ -14,13 +14,20 @@ object GenericServer {
     Http.collectZIO[Request] { case req @ Method.POST -> !! / "graphql" / id =>
       for {
         schema <- if (id == "0") SchemaRegistry.list(0, 1).map(_.headOption) else SchemaRegistry.get(Digest.fromHex(id))
-        result <- schema match {
+        timeout     <- ZIO.succeed(
+          (for {
+            blueprint <- schema
+            timeout   <- blueprint.server.globalResponseTimeout
+          } yield timeout).getOrElse(10000)
+        )
```
tusharmath, line 22: Can further delay this check, until it's confirmed that the blueprint was found.

tusharmath (approved):

PR #123, opened 2023-04-28
fix: should fail if http and unsafe are marked as required

amitksingh1490:


runtime/src/test/scala/tailcall/runtime/transcoder/Config2BlueprintSpec.scala
```diff
@@ -32,5 +34,21 @@ object Config2BlueprintSpec extends ZIOSpecDefault {
...
+            .withTypes("Query" -> Config.Type("foo" -> Config.Field.string.asRequired.withHttp(Http(Path.empty))))
+          assertZIO(Transcoder.toBlueprint(config).toZIO.flip)(equalTo(
+            Chunk("""`Query.foo` has http operation hence can not be non-nullable""")
+          ))
+        },
+        test("unsafe with required") {
+          val config = Config.default
+            .withTypes("Query" -> Config.Type("foo" -> Config.Field.string.asRequired.resolveWith(100)))
+          assertZIO(Transcoder.toBlueprint(config).toZIO.flip)(equalTo(
+            Chunk("""`Query.foo` has unsafe operation hence can not be non-nullable""")
```
tusharmath, line 49:
```suggestion
            Chunk("""`Query.foo` has an unsafe operation hence can not be non-nullable""")
```

runtime/src/test/scala/tailcall/runtime/transcoder/Config2BlueprintSpec.scala
```diff
@@ -32,5 +34,21 @@ object Config2BlueprintSpec extends ZIOSpecDefault {
 
         assertTrue(Transcoder.toBlueprint(config).nonEmpty)
       },
+      suite("required")(
+        test("http with required") {
+          val config = Config.default
+            .withTypes("Query" -> Config.Type("foo" -> Config.Field.string.asRequired.withHttp(Http(Path.empty))))
+          assertZIO(Transcoder.toBlueprint(config).toZIO.flip)(equalTo(
+            Chunk("""`Query.foo` has http operation hence can not be non-nullable""")
```
tusharmath, line 42:
```suggestion
            Chunk("""`Query.foo` has an http operation hence can not be non-nullable""")
```

tusharmath (approved):

PR #128, opened 2023-05-04
[WIP] rust reimplementation of server

mayant15:
Trying to re-implement the server in Rust for blazingly fast speeds.

**Why reimplement it entirely and not just FFI it in as a library?**
Easier and more incremental yes but the entry point would still be in Scala, and at some point there will need to be a "flip" into a Rust main function.  I don't know how that will work with `tokio`'s async main. Can be looked into. [Reference](https://tokio.rs/tokio/topics/bridging)


tusharmath:
Great! This is a fantastic start. I think we should think about implementing Lambda initially. We need a high-performance Lambda evaluator. Another thing is — we can create the ADTs for Rust using Scala? So that way we are always in sync between scala and rust.

PR #166, opened 2023-06-08
Feature: Support baseURL directly thru the `@http` directive

tusharmath:
Tests required :

- [x] Config -> SDL -> Config should maintain identity when baseURL is set in the HTTP operation.
- [x] baseURL set in the HTTP operation should be used to make HTTP calls when executed.
- [x] ~~Http Directive schema should encode/decode baseURL correctly~~
- [x] If no root base URL is defined and HTTP also doesn't have a baseURL then trigger a failure.

### DIFF
--- a/runtime/src/main/scala/tailcall/runtime/model/UnsafeSteps.scala
+++ b/runtime/src/main/scala/tailcall/runtime/model/UnsafeSteps.scala
@@ -65,6 +65,8 @@ object UnsafeSteps {
         self.copy(method = method, query = query, groupBy = groupBy)
       }
 
+      def withBaseURL(url: URL): Http = copy(baseURL = Option(url))
+
       def withBatchKey(batchKey: String): Http = copy(batchKey = Option(batchKey))
 
       def withBody(body: Option[String]): Http = copy(body = body)
--- a/runtime/src/main/scala/tailcall/runtime/model/UnsafeSteps.scala
+++ b/runtime/src/main/scala/tailcall/runtime/model/UnsafeSteps.scala
@@ -1,6 +1,7 @@
 package tailcall.runtime.model
 
 import tailcall.runtime.http.Method
+import tailcall.runtime.internal.JsonCodecImplicits._
 import tailcall.runtime.lambda.~>>
 import tailcall.runtime.model.Mustache.MustacheExpression
 import tailcall.runtime.model.UnsafeSteps.Operation
@@ -10,7 +11,7 @@ import zio.json.ast.Json
 import zio.schema.annotation.caseName
 import zio.schema.{DynamicValue, Schema}
 
-import java.net.{URI, URL}
+import java.net.URL
 
 @caseName("unsafe")
 final case class UnsafeSteps(steps: List[Operation]) {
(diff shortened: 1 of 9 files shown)


runtime/src/main/scala/tailcall/runtime/model/UnsafeSteps.scala
```diff
@@ -92,6 +95,13 @@ object UnsafeSteps {
     }
 
     object Http {
+      implicit val urlCodec: JsonCodec[URL] = JsonCodec[String].transformOrFail[URL](
+        string =>
+          try Right(URI.create(string).toURL)
+          catch { case _: Throwable => Left(s"Malformed url: ${string}") },
+        _.toString,
+      )
```
tusharmath, line 103: Find a common place to put this duplicate logic.

Can also create a PR in ZIO-JSON for JsonCodec[URL], if it doesn't exist.

runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
```diff
@@ -111,9 +111,10 @@ object Config2Blueprint {
       config.server.baseURL match {
         case Some(baseURL) => TValid.succeed {
             val steps    = field.unsafeSteps.getOrElse(Nil)
-            val host     = baseURL.getHost
-            val port     = if (baseURL.getPort > 0) baseURL.getPort else 80
-            val scheme   = if (baseURL.getProtocol.toLowerCase == "https" || port == 443) Scheme.Https else Scheme.Http
+            val url      = http.baseURL.getOrElse(baseURL)
```
tusharmath, line 114: Tests required :

- Config -> SDL -> Config should maintain the baseURL set in the HTTP operation.
- baseURL set in the HTTP operation should be used to make HTTP calls when executed.

runtime/src/main/scala/tailcall/runtime/model/UnsafeSteps.scala
tusharmath: We can add an `object JsonCodecImplicits` inside of `tailcall.runtime.internal`.
This would contain the implicit defs and then we could import it wherever required.

runtime/src/test/scala/tailcall/runtime/transcoder/Config2SDLSpec.scala
```diff
@@ -69,6 +72,26 @@ object Config2SDLSpec extends TailcallSpec {
...
+                          |}
+                          |
+                          |type Query {
+                          |  bar: String @http(path: "/bar")
+                          |  foo: String @http(path: "/foo",baseURL: "http://foo.com")
+                          |}""".stripMargin.trim
+
+        assertSDL(config, expected, true)
+
+      },
```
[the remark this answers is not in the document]
tusharmath: - [ ] Drop `asConfig` from `assertSDL`. `assertSDL` doesn't use it in any of the tests.

sujeetsr:
@tusharmath Addressed the comments, PR looks good to merge now.

PR #204, opened 2023-06-25
better response cache headers

amitksingh1490:
Intelligently calculate the cache headers from the downstream requests made by each query and forward the minimum to the client


runtime/src/main/scala/tailcall/runtime/service/HttpCache.scala
```diff
@@ -31,6 +35,27 @@ private[tailcall] object HttpCache {
     }
   }
 
+  import scala.util.Try
+
+  def calculateMinimumHeader(headers: List[(String, String)]): Option[String] = {
```
tusharmath, line 40: This function is unused!

tusharmath (approved):

Issue #216, opened 2023-06-27
Using `modify` on a field that's implemented via an interface


tusharmath:
`@modify` should not be possible on fields that are being implemented via an interface. Compiler should throw an error when `modify` is used this way.

PR #219, opened 2023-06-29
Don't allow modify on fields that are implemented via an interface

sujeetsr:
Fixes https://github.com/tailcallhq/tailcall/issues/216

@tusharmath - I've added a section for the error message in the graphql spec file. If the error message section is present, then the test checks for error while generating the client SDL. If not, it checks for success against the given client SDL

### DIFF
--- a/runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
+++ b/runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
@@ -72,7 +72,7 @@ object Config2Blueprint {
         val dblUsage = inputTypes.contains(typeName) && outputTypes.contains(typeName)
         for {
           _      <- TValid.fail(s"$typeName cannot be both used both as input and output type").when(dblUsage)
-          fields <- toFieldList(typeName, typeInfo, config.graphQL.types.toList)
+          fields <- toFieldList(typeName, typeInfo)
         } yield {
           val definition = Blueprint.ObjectTypeDefinition(
             name = typeName,
@@ -94,11 +94,7 @@ object Config2Blueprint {
       } yield Blueprint.FieldDefinition(name = fieldName, args = args, ofType = ofType, description = field.doc)
     }
 
-    private def toFieldList(
-      typeName: String,
-      typeInfo: Type,
-      types: List[(String, Type)],
-    ): TValid[String, List[Blueprint.FieldDefinition]] = {
+    private def toFieldList(typeName: String, typeInfo: Type): TValid[String, List[Blueprint.FieldDefinition]] = {
       TValid.foreach(typeInfo.fields.toList) { case (fieldName, field) =>
         {
           for {
@@ -108,8 +104,7 @@ object Config2Blueprint {
               else for {
                 bField      <- updateUnsafeField(field, bField).trace("@" + UnsafeSteps.directive.name)
                 bField      <- updateFieldHttp(field, bField).trace("@" + Http.directive.name)
-                mayBeBField <- updateModifyField(fieldName, field, bField, typeName, typeInfo, types)
-                  .trace("@" + ModifyField.directive.name)
+                mayBeBField <- updateModifyField(field, bField, typeInfo).trace("@" + ModifyField.directive.name)
                 bField      <- mayBeBField match {
                   case Some(bField) => updateInlineField(typeName, typeInfo, fieldName, field, bField).some
                       .trace("@" + InlineType.directive.name)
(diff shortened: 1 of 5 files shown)


runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
```diff
@@ -94,23 +94,62 @@ object Config2Blueprint {
...
+        {
+          for {
+            bField <- toFieldDefault(fieldName, field)
+            bField <-
+              if (inputTypes.contains(typeName)) TValid.succeed(List(bField))
+              else for {
+                _           <- validateField(fieldName, field, typeName, typeInfo, types)
+                bField      <- updateUnsafeField(field, bField).trace("@" + UnsafeSteps.directive.name)
+                bField      <- updateFieldHttp(field, bField).trace("@" + Http.directive.name)
+                mayBeBField <- updateModifyField(field, bField).trace("@" + ModifyField.directive.name)
```
tusharmath, line 144: We should implement the validation logic inside of `updateModifyField`

runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
```diff
@@ -300,15 +307,28 @@ object Config2Blueprint {
     }
 
     private def updateModifyField(
+      fieldName: String,
```
tusharmath, line 310: This isn't required. use `bfield.name`

runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
```diff
@@ -300,15 +307,28 @@ object Config2Blueprint {
...
       field: Field,
       bField: Blueprint.FieldDefinition,
+      typeName: String,
+      typeInfo: Type,
+      types: List[(String, Type)],
```
tusharmath, line 315: `types` are available everywhere, you don't need to pass as an argument.

runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
```diff
@@ -300,15 +307,28 @@ object Config2Blueprint {
     }
 
     private def updateModifyField(
+      fieldName: String,
       field: Field,
       bField: Blueprint.FieldDefinition,
+      typeName: String,
```
tusharmath, line 313: `typeName` is only used for errors. We don't need it because we set it via `trace` at the callsite. See how we fail here https://github.com/tailcallhq/tailcall/pull/219/files#diff-d6df7e22117627a7b52f3ec8ea3c2063effd3dc30928b528a9371ada608e34f7R332

runtime/src/test/resources/tailcall/runtime/graphql/test-modify-interface-field.graphql
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type B implements IA {
+  b: String
+  x: String
+}
+
+type Query {
+  bar: B
+}
```
tusharmath, line 35: Will the SDL be checked if there is a compilation error?

runtime/src/test/resources/tailcall/runtime/graphql/test-modify-interface-field.graphql
```diff
@@ -0,0 +1,39 @@
...
+
+#> client-error
+# Cannot use modify on field a on type B because it implements interface IA with the same field
+# B,a,@modify
```
tusharmath, line 39: May be we use this format to depict, the location and error message.
```graphql
# - [<location>] <message>
```
For example —
```suggestion
#> client-error
# - [B, a, @modify] Cannot use modify on field a on type B because it implements interface IA with the same field
```

tusharmath:
(nitpick) Maybe we can split the PR into two separate ones, one for writing specs in .graphql files where we also migrate the existing test cases, and one just to validate the modify error.

runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
```diff
@@ -302,13 +304,22 @@ object Config2Blueprint {
...
         case Some(ModifyField(Some(newName), None)) =>
-          val resolverPath = if (bField.resolver.isEmpty) List("value", bField.name) else List()
-          val resolver     = Lambda.identity[DynamicValue].path(resolverPath: _*).toDynamic
-          TValid.succeed(appendResolver(bField, resolver).copy(name = newName)).some
+          val interfaceWithSameField = config.graphQL.types.toList.filter(_._2.interface.isDefined)
+            .filter(interface => typeInfo.implements.toList.flatten.contains(interface._1))
+            .find(interface => interface._2.fields.find { case (iFieldName, _) => iFieldName == bField.name }.isDefined)
+          interfaceWithSameField match {
+            case Some(value) => TValid
+                .fail(s"Cannot use modify on field ${bField.name} because it is implemented from interface ${value._1}")
```
tusharmath, line 317:
```suggestion
                .fail(s"Implemented fields are unmodifiable")
```
More concise error message IMO. The field name and type name is already available in the trace.

runtime/src/test/resources/tailcall/runtime/graphql/test-modify-interface-field.graphql
tusharmath: I think we can drop the client-SDL here. Since we are asserting on errors thrown.

runtime/src/test/resources/tailcall/runtime/graphql/test-modify-interface-field.graphql
```diff
@@ -0,0 +1,38 @@
...
+type B implements IA {
+  b: String
+  x: String
+}
+
+type Query {
+  bar: B
+}
+
+#> client-error
```
tusharmath, line 37: Not sure if we should call them as client-errors. I think those should be issued only at run time, when we issue a real graphql query. Something else, perhaps - "validations"?

runtime/src/test/resources/tailcall/runtime/graphql/test-modify-interface-field.graphql
```diff
@@ -0,0 +1,38 @@
...
+# - [B, a, @modify] Cannot use modify on field a because it is implemented from interface IA
```
tusharmath, line 38: - [ ] Can we move the error to the top of the file?
- [ ] Can we make sure that invalid graphql specs are in a separate folder and are not mixed with the valid ones, where are are asserting server-SDL with produced client-SDL.

sujeetsr:
Closing this. Opened [223](https://github.com/tailcallhq/tailcall/pull/223) to fix the modify issue. Adding validation error messages will be done in a separate PR, along with moving the execution specs to graphql.

PR #223, opened 2023-07-02
compiler error on modifying inherited fields

### LINKED ISSUE #216: Using `modify` on a field that's implemented via an interface

sujeetsr:
fixes #216

### DIFF
--- a/runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
+++ b/runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
@@ -308,7 +308,7 @@ object Config2Blueprint {
         case Some(ModifyField(None, Some(true)))    => TValid.none
         case Some(ModifyField(Some(newName), None)) =>
           val matchingInterfaceName = typeInfo.implements.toList.flatten
-            .find(iName => config.findType(iName).exists(_.fields.get(bField.name).isDefined))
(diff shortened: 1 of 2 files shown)


runtime/src/main/scala/tailcall/runtime/transcoder/Config2Blueprint.scala
```diff
@@ -302,13 +302,20 @@ object Config2Blueprint {
...
+      typeInfo: Type,
     ): TValid[String, Option[Blueprint.FieldDefinition]] = {
       field.modify match {
         case Some(ModifyField(None, Some(true)))    => TValid.none
         case Some(ModifyField(Some(newName), None)) =>
-          val resolverPath = if (bField.resolver.isEmpty) List("value", bField.name) else List()
-          val resolver     = Lambda.identity[DynamicValue].path(resolverPath: _*).toDynamic
-          TValid.succeed(appendResolver(bField, resolver).copy(name = newName)).some
+          val matchingInterfaceName = typeInfo.implements.toList.flatten
+            .find(iName => config.findType(iName).exists(_.fields.get(bField.name).isDefined))
```
tusharmath, line 311: Use contains instead of get and isDefined

tusharmath (changes requested):

PR #274, opened 2023-08-06
chore: registry related commands and related types

girdharshubham:
- [x]  Delete CLI command - publish, list, show, drop
- [x] Remove unnecessary dependencies
- [ ]  Remove registry package
- [x]  Remove registry cli


girdharshubham:
Continues on from where @arkkade left off #272

girdharshubham:
> - [ ] Remove registry package

@tusharmath The registry package would still be needed, right?

What I have done is remove references for remote from the codebase.

tusharmath:
I think we can drop it. We will only support static configuration

PR #307, opened 2023-09-12
CI benchmark

sujeetsr:
Run benchmark with wrk and comment on the PR with the results.


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,116 @@
...
+      
+      - name: Run tailcall
+        run: |
+          ./target/release/tailcall start ci-benchmark/benchmark.graphql&
+
+      - name: Install Nginx
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y nginx
+
```
alankritdabral, line 57: - name: Install iai-callgrind
        runcargo install --version 0.7.3 iai-callgrind-runner

tusharmath: @alankritdabral Can you create a separate issue for whatever you are trying to communicate?

PR #328, opened 2023-09-23
fix: headers forwarding

amitksingh1490:


src/http/data_loader.rs
```diff
@@ -59,7 +61,16 @@ impl HttpDataLoader {
       .iter()
       .map(|key| async {
         let url = key.url.clone();
-        let req = reqwest::Request::new(reqwest::Method::from(&key.method), url);
+        let mut headers = HeaderMap::new();
+        for (k, v) in key.headers.iter() {
+          //TODO: remove expects and handle errors
+          headers.insert(
+            HeaderName::try_from(k.as_str()).expect("invalid header name"),
+            HeaderValue::from_str(v.as_str()).expect("invalid header value"),
```
tusharmath, line 69: We should not panic, can't we use question mark?

amitksingh1490: It was WIP, I have cleaned up now

Issue #333, opened 2023-09-26
`ValidationError` doesn't fail with exit code 1


tusharmath:
### Prerequisites

- [x] I have checked the existing issues to ensure this is not a duplicate.
- [x] I am running the latest version of Tailcall.

### Describe the bug

TC server when pointed to an invalid file, throws an validation error but exits with status code 0.

### Steps to reproduce

1. Create the following file `test.graphql`

   ```graphql
   schema @schema {
     query: Query
   }

   type Query {
     users: [User] @http(path: "/users")
   }
   ```

2. Run the check command on the file

   ```
     tc check ./assets/jsonplaceholder.graphql
     Validation Error
       • No base URL defined [Query, users, @http]

     echo $?
     0 # should be 1
   ```

### Expected behavior

Fail with exit code Err.

### Actual behavior

Fails with exit code OK.

### Environment information:

- **Operating System**: All
- **Tailcall Version**: 0.9.0
- **GraphQL Client/version** NA

### Additional context

Learn more about exit codes here — https://rust-cli.github.io/book/in-depth/exit-code.html
\*\*

ayewo:
@tusharmath what exit code would you like it to emit?

tusharmath:
@ayewo we could use `78`. We could use the exit code crate for this also — https://docs.rs/exitcode/latest/src/exitcode/lib.rs.html#89

ayewo:
@tusharmath 
I'm curious which style of working is your project adopting with Algora:
1. let the first contributor that uses the `/attempt` command to work on an issue; or
2. let anyone attempt an issue right away. Multiple contributors can use the `/attempt` command. 

The upside of style 2 is that you get multiple contributions on a single issue (as is the case here) but you only pay one person. The downside for contributors obviously is that if their work is not accepted they get nothing for their time.

FWIW, the Remotion project uses [style 1](https://github.com/remotion-dev/remotion/issues/2495#issuecomment-1683640507) which avoids the scenario described in style 2.

tusharmath:
I would prefer one.

kremedev:
Bounties at this price range and difficulty generally are style 2

tusharmath:
Agree with @kremedev

ayewo:
> Bounties at this price range and difficulty generally are style 2

Nope, this is not codified anywhere. The style of working is entirely up to each project to decide which is why I asked to confirm.

kremedev:
I'm just telling what I have seen so far

[...]

PR #335, opened 2023-09-26
fix: ValidationError doesn't fail with exit code 1

kremedev:
fix: #333 
/claim #333

### DIFF
--- a/src/blueprint/from_config.rs
+++ b/src/blueprint/from_config.rs
@@ -278,7 +278,10 @@ fn update_http(field: &config::Field, b_field: FieldDefinition, config: &Config)
 
           Valid::Ok(b_field)
         }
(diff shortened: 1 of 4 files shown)


src/blueprint/from_config.rs
```diff
@@ -278,7 +278,10 @@ fn update_http(field: &config::Field, b_field: FieldDefinition, config: &Config)
 
           Valid::Ok(b_field)
         }
-        None => Valid::fail("No base URL defined".to_string()),
+        None => {
+          std::process::exit(1);
```
tusharmath, line 282: I think we need a more generic approach instead of exiting with status code 1 for this specific error.

tusharmath:
closing in favour of #336

PR #336, opened 2023-09-26
Fix: validation should fail with non-zero exit code.

ayewo:
@tusharmath

This is my first PR in Rust :) so please let me know if I need to make any changes.

/claim #333

### DIFF
--- a/src/cli/tc.rs
+++ b/src/cli/tc.rs
@@ -32,7 +32,7 @@ pub async fn run() -> Result<()> {
           let err_str = format!("{:?}", e);
           let formatted_err = Fmt::error(&err_str);
           Fmt::display(formatted_err);
-          Ok(())
+          std::process::exit(exitcode::CONFIG);
         }
       }
     }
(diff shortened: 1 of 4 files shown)


src/cli/tc.rs
```diff
@@ -32,7 +32,7 @@ pub async fn run() -> Result<()> {
           let err_str = format!("{:?}", e);
           let formatted_err = Fmt::error(&err_str);
           Fmt::display(formatted_err);
-          Ok(())
+          std::process::exit(exitcode::CONFIG);
```
tusharmath, line 35: We perform a process::exit inside of `main.rs`. Though we exit with status `1` and we could change it to 78 or `exitcode::Config` but we should only do that at one place ie. inside of `main.rs` only.

ayewo: @tusharmath when I put `process::exit` in `main.rs` and replace `std::process::exit(exitcode::CONFIG);` with a `return Err(e)` statement, the error message is output twice, which is of course not what you want. 
So, should I remove the error output in `src/cli/tc.rs`?

```bash
Validation Error
• No base URL defined [Query, users, @http]                  <--- printed by src/cli/tc.rs

Error: Validation Error
• No base URL defined [Query, users, @http]                  <--- printed by src/main.rs
```

tusharmath (approved):

tusharmath (approved):

Issue #334, opened 2023-09-26
Bailout on undeclared types


tusharmath:
### Prerequisites

- [x] I have checked the existing issues to ensure this is not a duplicate.
- [x] I am running the latest version of Tailcall.

### Describe the bug
At the time of schema validation, whenever there is a type that's not defined in the schema/config, tailcall server should bailout and display and error message with a proper trace.

### Steps to reproduce
Use the following graphql config 

```graphql
schema @schema {
  query: Query
}

type Query {
  users: [User] @http(baseURL: "https://jsonplaceholder.typicode.com", path: "/users")
}
```

Perform a check command 

```bash
tc check ./assets/jsonplaceholder.graphql
No errors found
N + 1: 0
```

### Expected behavior
- [ ] A validation error should be shown — `Undeclared type 'User' was found [@ Query, users]
- [ ] `tc start` should not panic

### Actual behavior
- No validation error is thrown
- Server start panics and fails.

### Environment information:
- All envs.

tusharmath:
fixed in #338

PR #338, opened 2023-09-26
Fix: Add validation for undeclared types and unions

meskill:
/claim #334


src/config/config.rs
```diff
@@ -131,7 +135,7 @@ impl Type {
 pub struct GraphQL {
   pub schema: RootSchema,
   pub types: BTreeMap<String, Type>,
-  pub unions: Option<Vec<Union>>,
+  pub unions: BTreeMap<String, Union>,
```
tusharmath, line 138: I think if we wish to store the name of the Union type as a Key in BTreeMap, we should remove the field `name` from `Union`

tusharmath: BTW I am wondering if the change in union should be in the scope of this PR. However this is a good improvement so happy to merge :)

src/blueprint/from_config.rs
```diff
@@ -168,13 +170,19 @@ fn to_interface_type_definition(definition: ObjectTypeDefinition) -> Valid<Defin
 }
 fn to_fields(type_of: &config::Type, config: &Config) -> Valid<Vec<blueprint::FieldDefinition>> {
   let fields: Vec<Option<blueprint::FieldDefinition>> = type_of.fields.iter().validate_all(|(name, field)| {
+    let field_type = &field.type_of;
+
+    if !is_scalar(field_type) && config.find_type(field_type).is_none() && config.find_union(field_type).is_none() {
+      return Valid::fail(format!("Undeclared type '{field_type}' was found")).trace(&name);
```
tusharmath, line 176: Performing a `return` will short circuit the execution ignoring all potential errors that could arise in the later part of the checks. For example — if a field has two errors viz no base url and an undeclared type, because we return directly from here, we won't get to know about the base url error until we fix the undeclared type error and then retry.

I would recommend using `validate_or` operator between this logic and the else part so that errors can be combined and shown to the user.

tests/graphql/errors/undeclared-type.graphql
```diff
@@ -0,0 +1,12 @@
+#> server-sdl
+schema @schema {
+  query: Query
+}
+
+type Query {
+  users: [User] @http(baseURL: "https://jsonplaceholder.typicode.com", path: "/users")
```
tusharmath, line 7:
```suggestion
  users0: [User] @http(baseURL: "https://jsonplaceholder.typicode.com", path: "/users")
  users1: [User] @http(path: "/users")
```
Ideally we should see two errors for `users1` field:
1. `The type 'User' was not found`.
2. `No base URL defined`.

src/blueprint/from_config.rs
```diff
@@ -168,13 +170,19 @@ fn to_interface_type_definition(definition: ObjectTypeDefinition) -> Valid<Defin
 }
 fn to_fields(type_of: &config::Type, config: &Config) -> Valid<Vec<blueprint::FieldDefinition>> {
   let fields: Vec<Option<blueprint::FieldDefinition>> = type_of.fields.iter().validate_all(|(name, field)| {
+    let field_type = &field.type_of;
+
+    if !is_scalar(field_type) && config.find_type(field_type).is_none() && config.find_union(field_type).is_none() {
```
tusharmath, line 175:
```suggestion
    if !is_scalar(field_type) && config.contains(field_type) {
```
Can add a helper method `contains`.

tests/graphql/errors/undeclared-type.graphql
meskill: I've put this test case into separate test

src/blueprint/from_config.rs
meskill: done

tusharmath:
@meskill this is a fmt error can you have a look.

src/blueprint/from_config.rs
tusharmath: Thanks for catching the bug!

tusharmath: /tip $100

meskill:
> @meskill this is a fmt error can you have a look.

Could you rerun the workflow? I've fixed formatting

tusharmath (approved):

PR #354, opened 2023-09-27
Fix: Header forwarding

amitksingh1490:


src/request_template.rs
```diff
@@ -110,49 +112,76 @@ impl TryFrom<Endpoint> for RequestTemplate {
...
+  use std::borrow::Cow;
+
   use hyper::HeaderMap;
   use pretty_assertions::assert_eq;
   use serde_json::json;
 
   use crate::mustache::Mustache;
   use crate::request_template::RequestTemplate;
-
+  struct TestContext {
```
tusharmath, line 123: can add a default and setters.

tusharmath, line 123:
```suggestion
  struct Context {
```

tusharmath (approved):

Issue #342, opened 2023-09-27
tc check incorrectly identifies n + 1


sujeetsr:
### Description
`tc check -n` includes query and relationships that are not lists in `N+1` information

### Steps to reproduce
Create a `jsonplaceholder.graphql` config

```graphql
schema
  @server(
    baseURL: "http://jsonplaceholder.typicode.com"
    enableGraphiql: "/graphiql"
  ) {
  query: Query
}

type User {
  id: ID!
  name: String
  comments: [Comment] @http(path: "/users/{{value.id}}/comments")
}

type Comment {
  id: ID!
  body: String
}

type Query {
  user(id: ID!): User @http(path: "/users/{{args.id}}")
}
```
Run the following command 

```
cargo run -- check -n  ./jsonplaceholder.graphql
```

### Expected behavior
```bash
No errors found
N + 1: 0
```

### Actual behaviour
Incorrectly identifies the query as a `n + 1` query.
```bash
No errors found
N + 1: 1
  query { user { comments } }
```

tusharmath:
fixed in #359

PR #359, opened 2023-09-28
fix: fix false-positive case for n + 1 checker

meskill:
/claim #342

### DIFF
--- a/src/config/n_plus_one.rs
+++ b/src/config/n_plus_one.rs
@@ -20,7 +20,7 @@ pub fn n_plus_one(config: &Config) -> Vec<Vec<(String, String)>> {
           } else if field.has_resolver() && !field.has_batched_resolver() && is_list {
             vec![new_path]
           } else {
-            find_fan_out(config, &field.type_of, new_path, field.list.is_some() || is_list)
+            find_fan_out(config, &field.type_of, new_path, field.list.unwrap_or(false) || is_list)
           }
         })
         .collect(),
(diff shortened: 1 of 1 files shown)


src/config/n_plus_one.rs
```diff
@@ -20,7 +20,7 @@ pub fn n_plus_one(config: &Config) -> Vec<Vec<(String, String)>> {
           } else if field.has_resolver() && !field.has_batched_resolver() && is_list {
             vec![new_path]
           } else {
-            find_fan_out(config, &field.type_of, new_path, field.list.is_some() || is_list)
+            find_fan_out(config, &field.type_of, new_path, field.list.unwrap_or(false) || is_list)
```
meskill, line 23: As for example from #342 the User gets `.list = Some(false)` and that was wrongly handled by n+1 check code.

I've a little bit confused of using Option<bool> for such kind of flag instead of pure bool, but there are several usages of it and I don't want to break anything in case it is important

tusharmath: Config can read directly from a .json or a yaml file. That allows A user to directly write Config and we build a blueprint directly from it. Option of bool is used so that the user doesn't need to specify every boolean option and we can assume certain defaults.

tusharmath: Hope this makes sense, happy to answer more questions if you have any.

meskill: Thanks for the explanation, then current changes should be ok to fix the issue.

Speaking of reading from config - using [serde's default](https://serde.rs/attr-default.html) might be helpful for this. But it is out of scope for the current issue

tusharmath: @meskill Thanks! Would you mind creating an issue for it?

tusharmath (approved):

src/config/n_plus_one.rs
tusharmath, line 23:
```suggestion
            find_fan_out(config, &field.type_of, new_path, field.isList() || is_list)
```
Would really appreciate if we could add a `isList` operator on config. This will keep the code clean.

meskill: Created https://github.com/tailcallhq/tailcall/issues/367

meskill: The branch is already merged. Do you want to see this as a separate PR, or would you prefer it to be covered by another approach from the https://github.com/tailcallhq/tailcall/issues/367?

PR #362, opened 2023-09-28
fix: add validation that root resolvers are existing in schema for query and mutation

### LINKED ISSUE #207: Validation for missing resolvers at Root level
**Description:**
When a tailcall server is configured without a resolver at the root level, it starts successfully. However, it doesn't respond to any requests, which can be misleading for developers. This behavior should be addressed by enforcing the presence of a resolver at compile time.

**Reproduction Steps:**
1. Configure the tailcall server without a root-level resolver.
2. Start the server.
3. Observe that the server starts successfully without errors.
4. Make a request to the server.
5. Notice the server provides no response.

**Example Configuration with Issue:**
```graphql
type Query {
  user: [User]! # Unresolved root level configuration
}

type User {
  id: ID
  name: String
  posts: [Post] @http(path: "/posts/{{value.id}}") 
}
```

**Proposed Solution:**
If a configuration lacks a resolver at the root level and none can be inferred, a compile-time error should be thrown, clearly indicating the absence of a necessary resolver. This ensures that developers are aware of the missing resolver before runtime.

**Update 1**
This should only be applicable for scenarios where the type of the field is **required**. In a scenario where the field is not required, it's totally fine if there are no resolvers.

meskill:
**Summary:**  
Add validation that root resolvers are present both for query and mutation

/claim #207

**Issue Reference(s):**  
Fixes #207 

**Build & Testing:**

- [x] The build is passing.

**Checklist:**


src/blueprint/from_config.rs
```diff
@@ -223,6 +225,56 @@ fn to_type(name: &str, list: &Option<bool>, required: &Option<bool>, list_type_r
...
+
+  let Some(query) = config.find_type(query_type_name) else {
+    return Valid::fail("Query type is not defined".to_owned()).trace(query_type_name);
+  };
+
+  query
+    .fields
+    .iter()
+    // validate only required fields
+    // TODO: somehow when field is required it has a value of Option<false> due to https://github.com/tailcallhq/tailcall/blob/49e39e12013cf837ce0cc78562c4207f4a9782ad/src/config/from_document.rs#L195
```
meskill, line 244: this is related to https://github.com/tailcallhq/tailcall/issues/371

I didn't find when the value could be `Some(true)` in this case so this could be a little messy

tusharmath: - We could implement `is_required()` on field which unwrap_or(false)
- You could possibly attempt #371 first? I still don't have a PR yet.

What do you want to do?

meskill: I prefer to fix it once and for all by removing option wrappers as it already hit me couple times for the last couple days 😄 

If I can find some time before someone solves it, I would like to attempt.

tusharmath: Perfect! Looking forward to your PR for #371 :)

src/blueprint/from_config.rs
```diff
@@ -223,6 +225,56 @@ fn to_type(name: &str, list: &Option<bool>, required: &Option<bool>, list_type_r
...
+    .filter(|(_, field)| field.required.is_some())
```
tusharmath, line 245:
```suggestion
```

On second thoughts we should throw this error irrespective of the nullability.

tusharmath: Apologies on going to and fro on this 🙈

meskill: no problem, I've reverted these changes

tusharmath (approved):

PR #382, opened 2023-09-29
refactor: replace Option wrappers with serde(default) attribute

### LINKED ISSUE #371: Remove unnecessary `Option` wrapping in `Config.rs`
Expanding on #367 

The following fields have type `Option<bool>` and `Option<Vec<_>>` in various types in config.rs. 
We can drop them all and use [serde::default] for all of them and make our Config a lot more compact.

[serde::default]: https://serde.rs/attr-default.html

```rust
pub interface: Option<bool>
pub scalar: Option<bool>
pub list: Option<bool>
pub required: Option<bool>
pub list_type_required: Option<bool>
pub omit: Option<bool>
pub list: Option<bool>
pub required: Option<bool>

pub implements: Option<Vec<String>>
pub variants: Option<Vec<String>>
pub match_path: Option<Vec<String>>
```

meskill:
**Summary:**  

Instead of `Option<bool>` etc use `#[serde(default)]` to mark fields as optional with some default value.

These changes simplifies the code and prevents possible bugs when utilizing Option

**Issue Reference(s):**  
Fixes #371

/claim #371

**Build & Testing:**

- [ ] I have run `./lint.sh` to address and fix linting issues.

**Checklist:**

### DIFF
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -11,7 +11,7 @@ use crate::batch::Batch;
 use crate::http::Method;
 use crate::json::JsonSchema;
 
-fn skip_if_default<T: Default + Eq>(val: &T) -> bool {
+fn is_default<T: Default + Eq>(val: &T) -> bool {
   *val == T::default()
 }
 
@@ -176,7 +176,7 @@ impl Field {
   }
   pub fn has_batched_resolver(&self) -> bool {
     if let Some(http) = self.http.as_ref() {
-      http.match_key.is_some() || http.match_path.is_some()
+      http.match_key.is_some() || !http.match_path.is_empty()
     } else {
       false
     }
@@ -196,7 +196,7 @@ pub struct Unsafe {
 pub struct ModifyField {
   pub name: Option<String>,
   #[serde(default)]
-  #[serde(skip_serializing_if = "skip_if_default")]
+  #[serde(skip_serializing_if = "is_default")]
   pub omit: bool,
 }
 
@@ -227,17 +227,23 @@ pub struct Union {
 pub struct Http {
   pub path: String,
   #[serde(default)]
-  #[serde(skip_serializing_if = "skip_if_default")]
+  #[serde(skip_serializing_if = "is_default")]
   pub method: Method,
-  pub query: Option<BTreeMap<String, String>>,
+  #[serde(default)]
+  #[serde(skip_serializing_if = "is_default")]
+  pub query: BTreeMap<String, String>,
   pub input: Option<JsonSchema>,
   pub output: Option<JsonSchema>,
   pub body: Option<String>,
-  pub match_path: Option<Vec<String>>,
+  #[serde(default)]
+  #[serde(skip_serializing_if = "is_default")]
+  pub match_path: Vec<String>,
   pub match_key: Option<String>,
   #[serde(rename = "baseURL")]
   pub base_url: Option<String>,
-  pub headers: Option<BTreeMap<String, String>>,
+  #[serde(default)]
+  #[serde(skip_serializing_if = "is_default")]
+  pub headers: BTreeMap<String, String>,
 }
 
 impl Http {
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -53,10 +53,8 @@ impl Config {
     for (_, type_of) in self.graphql.types.iter() {
       if !type_of.interface {
         for (_, field) in type_of.fields.iter() {
-          if let Some(ref args) = field.args {
-            for (_, arg) in args.iter() {
-              types.insert(&arg.type_of);
-            }
+          for (_, arg) in field.args.iter() {
+            types.insert(&arg.type_of);
           }
         }
       }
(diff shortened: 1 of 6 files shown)


tests/graphql/test-modify.graphql
```diff
@@ -8,7 +8,7 @@ input Foo {
 }
 
 type Query {
-  foo(input: Foo): String @modify(name: "data")
+  foo(input: Foo): String @modify(name: "data", omit: false)
```
meskill, line 11: had to change this part in test file as conversion `SDL -> config -> SDL` now may differ due to fact that defaults are now adding to the config

src/config/config.rs
```diff
@@ -112,17 +112,16 @@ impl Config {
...
   pub doc: Option<String>,
-  pub interface: Option<bool>,
-  pub implements: Option<Vec<String>>,
-  #[serde(rename = "enum")]
+  #[serde(default)]
+  pub interface: bool,
+  #[serde(default)]
+  pub implements: Vec<String>,
+  #[serde(rename = "enum", default)]
   pub variants: Option<Vec<String>>,
```
meskill, line 120: Note that I haven't changed this. It is because `variants` are using as discriminated value to separate enums from other types and to remove it we need some other flag/switch to handle it as enum properly

meskill:
I've tried to replace as much properties as possible but changing some fields (like [in Http](https://github.com/tailcallhq/tailcall/blob/ef20f2028c594a17ac008df91bcc3206fde3c268/src/config/config.rs#L224)) breaks tests as now deserialization of config to sdl outputs more types that were initially provided. That requires either changing tests itself or how identity is compared

tests/graphql/test-modify.graphql
tusharmath: so the default is `true`?

meskill: no, default is false, but in this test it fails comparing input `@modify(name: "data")` to the output `@modify(name: "data", omit: false)`

tusharmath:
Identity tests are required because there are use cases where we generate a new config using open API or Adobe other format. Some times we want to merge two config into one. I don't think we should change it.

meskill:
> Identity tests are required because there are use cases where we generate a new config using open API or Adobe other format. Some times we want to merge two config into one. I don't think we should change it.

Could you please point out to some examples or documentation about this to better understand current usage?

Identity tests should stay I agree, but to be able to replace Option to default we could:
- leave that pr as is with replaced entries that do not affect actual output (like in `Field`) while leave other entries as is (that are related to directives)
- improve identity tests that they are validate that `output == input + defaults` and not just `input == output`

tusharmath:
Think about `Config` as a reduced form of input from the user. The user (a human) can write a Json, Yaml, GraphQL or use a GUI to create an input to tailcall. Let's say the input is of type "I". Now here is what happens next:

- Convert "I" into a consistent input format which in our case is called as "Config". All inputs should be converted into a "Config".
- From Config, we convert to Blueprint and then into a GraphQL executor.
- The reason why we need to have a consistent input format — Config, is that we should be able to convert from one input format to another. So a user could write in .json format, but could allow it to be converted into a .graphql format. Or a user could use a GUI to generate .graphql config, download it and start the server. 

I hope this makes sense. Let me know if it doesn't, happy to talk more about the architecture and the reason behind it.

Feel free to connect with us on [discord] if you prefer a more real time collaboration :)

tusharmath:
The identity tests, guarantee that we always generate the most minimal and compact configuration that's possible. This is one of the reasons why we started with the `Option` wrapping.

tusharmath:
> Identity tests are required because there are use cases where we generate a new config using open API or some other format. Some times we want to merge two config into one. I don't think we should change it.

It's not `Adobe` I meant `some`

meskill:
> The identity tests, guarantee that we always generate the most minimal and compact configuration that's possible. This is one of the reasons why we started with the `Option` wrapping.

In case we want to generate as minimal configuration as possible from Config I have a suggestion to use [serde's skip_serializing](https://serde.rs/attr-skip-serializing.html). In that case fields with that attribute won't appear in the output if they have a default value.

Please, look at my latest commit that shows the example of usage. 

One possible side effect of this is that some input settings could disappear from the output if they are equal to the default value

tusharmath:
I think this is going in the right right direction. We have a few pending `Option<Vec<_>>`.

src/config/config.rs
```diff
@@ -190,7 +195,9 @@ pub struct Unsafe {
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct ModifyField {
   pub name: Option<String>,
-  pub omit: Option<bool>,
+  #[serde(default)]
+  #[serde(skip_serializing_if = "skip_if_default")]
```
tusharmath, line 199:
```suggestion
  #[serde(skip_serializing_if = "is_default")]
```

meskill:
> I think this is going in the right right direction. We have a few pending `Option<Vec<_>>`.

Updated other fields that has clear defaults. Please, note [the comment](https://github.com/tailcallhq/tailcall/pull/382#discussion_r1341430735) why some `Option<Vec>` wasn't replaced

tusharmath (approved):

PR #388, opened 2023-09-30
feat: support to configure http settings

### LINKED ISSUE #383: Support to configure http setting via server
Add support to configure http-client settings on the server

```graphql
schema @server(http: {
  keepAliveInterval: 1,
  keepAliveTimeout: 1
  keepAliveWhileIdle: true,
  poolIdleTimeout: 1024
  poolMaxIdlePerHost: 100
}) {
  query: Query
  mutation: Mutation
}
```

Use these settings while configuring `http_client.rs`

akhill10:
**Summary:**  
Support to configure http settings in the server operator

**Issue Reference(s):**  
Fixes #383 

**Build & Testing:**

**Checklist:**

/claim #383


src/config/server.rs
```diff
@@ -47,3 +48,13 @@ impl Server {
 pub struct Proxy {
   pub url: String,
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct HttpSettings {
```
tusharmath, line 54:
```suggestion
pub struct Upstream {
```
On second thoughts, we should call this as upstream

src/config/server.rs
```diff
@@ -19,6 +19,7 @@ pub struct Server {
   pub port: Option<u16>,
   pub proxy: Option<Proxy>,
   pub vars: Option<BTreeMap<String, String>>,
+  pub http: Option<HttpSettings>,
```
tusharmath, line 22:
```suggestion
  pub upstream: Option<Upstream>,
```

src/config/server.rs
```diff
@@ -47,3 +48,13 @@ impl Server {
...
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct HttpSettings {
+  pub keep_alive_interval: u64,
+  pub keep_alive_timeout: u64,
+  pub keep_alive_while_idle: bool,
+  pub pool_idle_timeout: u64,
+  pub pool_max_idle_per_host: usize,
```
tusharmath, line 59: I think we should try to support all the settings provided over here —  https://github.com/seanmonstar/reqwest/blob/master/src/async_impl/client.rs#L87


```rust
struct Config {
    // NOTE: When adding a new field, update `fmt::Debug for ClientBuilder`
    accepts: Accepts,
    headers: HeaderMap,
    #[cfg(feature = "native-tls")]
    hostname_verification: bool,
    #[cfg(feature = "__tls")]
    certs_verification: bool,
    #[cfg(feature = "__tls")]
    tls_sni: bool,
    connect_timeout: Option<Duration>,
    connection_verbose: bool,
    pool_idle_timeout: Option<Duration>,
    pool_max_idle_per_host: usize,
    tcp_keepalive: Option<Duration>,
    #[cfg(any(feature = "native-tls", feature = "__rustls"))]
    identity: Option<Identity>,
    proxies: Vec<Proxy>,
    auto_sys_proxy: bool,
    redirect_policy: redirect::Policy,
    referer: bool,
    timeout: Option<Duration>,
    #[cfg(feature = "__tls")]
    root_certs: Vec<Certificate>,
    #[cfg(feature = "__tls")]
    tls_built_in_root_certs: bool,
    #[cfg(feature = "__tls")]
    min_tls_version: Option<tls::Version>,
    #[cfg(feature = "__tls")]
    max_tls_version: Option<tls::Version>,
    #[cfg(feature = "__tls")]
    tls_info: bool,
    #[cfg(feature = "__tls")]
    tls: TlsBackend,
    http_version_pref: HttpVersionPref,
    http09_responses: bool,
    http1_title_case_headers: bool,
    http1_allow_obsolete_multiline_headers_in_responses: bool,
    http1_ignore_invalid_headers_in_responses: bool,
    http1_allow_spaces_after_header_name_in_responses: bool,
    http2_initial_stream_window_size: Option<u32>,
    http2_initial_connection_window_size: Option<u32>,
    http2_adaptive_window: bool,
    http2_max_frame_size: Option<u32>,
    http2_keep_alive_interval: Option<Duration>,
    http2_keep_alive_timeout: Option<Duration>,
    http2_keep_alive_while_idle: bool,
    local_address: Option<IpAddr>,
    nodelay: bool,
    #[cfg(feature = "cookies")]
    cookie_store: Option<Arc<dyn cookie::CookieStore>>,
    trust_dns: bool,
    error: Option<crate::Error>,
    https_only: bool,
    #[cfg(feature = "http3")]
    tls_enable_early_data: bool,
    #[cfg(feature = "http3")]
    quic_max_idle_timeout: Option<Duration>,
    #[cfg(feature = "http3")]
    quic_stream_receive_window: Option<VarInt>,
    #[cfg(feature = "http3")]
    quic_receive_window: Option<VarInt>,
    #[cfg(feature = "http3")]
    quic_send_window: Option<u64>,
    dns_overrides: HashMap<String, Vec<SocketAddr>>,
    dns_resolver: Option<Arc<dyn Resolve>>,
}
```

src/config/server.rs
```diff
@@ -19,6 +19,7 @@ pub struct Server {
   pub port: Option<u16>,
   pub proxy: Option<Proxy>,
```
tusharmath, line 20: proxy should move to `upstream`

tests/graphql/passed/test-https-settings.graphql
```diff
@@ -0,0 +1,26 @@
...
+type Query {
+    firstUser: User @http(path: "/users/1" baseURL: "http://jsonplaceholder.typicode.com")
+}
+
+#> client-query
+query @expect(json: {data:{firstUser:{name:"Leanne Graham"}}}) {
+    firstUser {
+        name
+    }
+}
```
tusharmath, line 26: I don't think we need these tests.

akhill10: @tusharmath Are there any other ways to test?

src/config/server.rs
akhill10: @tusharmath  Are all the fields optional? And as we are using a builder pattern for initializing the client options.  So I have to make a check for each and every field and instaniate? Or is there any other way? 

something like this 
```rust
if let Some(ref upstream) = server.upstream {
      if let Some(ref keep_alive_interval) = upstream.keep_alive_interval {
        builder = builder.http2_keep_alive_interval(Duration::from_millis(keep_alive_interval.clone()))
      }
     // similarly for other fields
 }

```

tusharmath: You could do something as follows — 

```rust
// config::Server.rs
Upstream {
  keep_alive_timeout: u64
  // and the rest of the fields ...
}

impl From<Upstream> for reqwest::Config {
  fn from(value: Upstream) -> Self {
    reqwest::Config {
      // ... now set fields one by one
    }
  }
}
```

Once you have the reqwest::Config, you can use that directly to create a client.

tests/graphql/passed/test-https-settings.graphql
tusharmath: I think we can ignore the tests, as this is something that we can rely on for the compiler for. It is not possible to create an invalid client type.

src/config/server.rs
akhill10:
```
 cannot find type 'Config' in crate 'reqwest' not found in 'reqwest'rustc
```

Getting this compilation error.

akhill10: I think only way to set custom configuration is through builder() ( Maybe I m wrong , I m new to rust )

https://rcos.io/static/internal_docs/reqwest/struct.Client.html
> The Client has various configuration values to tweak, but the defaults are set to what is usually the most commonly desired value. To configure a Client, use Client::builder().

tusharmath: Just realized, that Config is private inside of reqwest, we will have to take another approach.

src/http/client.rs
```diff
@@ -27,6 +27,15 @@ impl HttpClient {
...
       .user_agent("Tailcall/1.0");
 
+    if let Some(ref http_settings) = server.http {
+      builder = builder
+        .http2_keep_alive_interval(Some(Duration::from_millis(http_settings.keep_alive_interval)))
+        .http2_keep_alive_timeout(Duration::from_millis(http_settings.keep_alive_timeout))
+        .http2_keep_alive_while_idle(http_settings.keep_alive_while_idle)
+        .pool_idle_timeout(Some(Duration::from_millis(http_settings.pool_idle_timeout)))
+        .pool_max_idle_per_host(http_settings.pool_max_idle_per_host)
+    }
```
tusharmath, line 37: I think this implementation is good for now, we should just add the options that the client is already configured with right above your changes — https://github.com/tailcallhq/tailcall/pull/388/files#diff-949af4887508b0c9ff8c0ba77a47a53fe8d3cc46c940dd8218463517e052b12fR23

That way we build the builder in one go.

akhill10: Got it.  Thanks
Should I support all fields or only the above fields would suffice for now?

tusharmath: Just the above fields for now. I think we will have to go thru each one of them to see what makes sense and what doesn't. Let's keep it simple for now.

src/http/client.rs
```diff
@@ -27,17 +27,20 @@ impl HttpClient {
       .connect_timeout(Duration::from_secs(60))
       .user_agent("Tailcall/1.0");
```
tusharmath, line 28: We should set them also via upstream settings. No hard coded values should be set over here, all should come from Upstream.

akhill10: @tusharmath Changes done. No hard coded values. But if the client doesn't pass any values then set default values.

src/http/client.rs
```diff
@@ -20,14 +20,31 @@ impl Default for HttpClient {
 
 impl HttpClient {
   pub fn new(server: Server) -> Self {
+    let upstream_settings = &server.upstream.clone().unwrap_or(Upstream {
```
meskill, line 23: Please, note with current code defining upstream in config require settings all the fields of it, as if you omit property it won't be properly parsed by serde and the whole `server.upstream` will be None. I.e. there is no way to override specific settings without defining the whole set through config.

To mitigate the issue I suggest to use [serde default](https://serde.rs/attr-default.html) attribute for fields of upstream that will provide defaults properly. 

Another suggestion it's actually write some basic test to catch the behaviour I've mentioned as it's not obvious 🙂 

Another thing I've noticed that in case of failed parsing there is no error on `tc check` and therefore some default logic just applying. It's worth another issue I think

akhill10: @meskill yeah I agree with you, in present case client have to define all the fields. 

Previously I played with [derivative](https://mcarton.github.io/rust-derivative/latest/index.html). Its pretty straightforward. But this project is not using this crate. Maybe we can use this ? not sure.

And regarding tests, a comment from tushar

> I think we can ignore the tests, as this is something that we can rely on for the compiler for. It is not possible to create an invalid client using the builder.

meskill: I'm not sure that derivative could help in this case as it seems like allows to generate implementation for `Default` trait with some customization, but here we need not just default for the whole struct but defaults for every field separately when deserializing it.

About test, just wanted to point out that compiler verification sometimes is not enough to be sure everything works as expected 🙂

tusharmath: I agree with @meskill I think we should add some tests around the serde bits of `Server`. We can add a simple test where we assert that default is empty.

```rust
Server::default().to_json() == "{}"
```

@meskill I think we could remove the option wrapping from `Server` as well.

src/http/client.rs
```diff
@@ -20,14 +20,24 @@ impl Default for HttpClient {
...
-      .timeout(Duration::from_secs(60))
-      .connect_timeout(Duration::from_secs(60))
-      .user_agent("Tailcall/1.0");
-
-    if let Some(ref proxy) = server.proxy {
+    let mut upstream_settings_serilaised: String = serde_json::to_string(&server.upstream).unwrap();
+    if upstream_settings_serilaised == "null" {
+      upstream_settings_serilaised = "{}".to_owned()
+    }
+    let upstream_settings: Upstream = serde_json::from_str(&upstream_settings_serilaised).unwrap();
```
meskill, line 27: instead of doing this consider marking the `server.upstream` with `serde(default)` as well and implement Default trait for it

akhill10: @meskill done.

meskill: you can fully get rid of conversions with serde_json and use just `server.upstream` - the defaults for it will be provided by serde on root parsing

src/config/server.rs
```diff
@@ -47,3 +48,85 @@ impl Server {
...
+}
+
+fn keep_alive_while_idle_default() -> bool {
+  false
+}
+
+fn proxy_default() -> Option<Proxy> {
+  None
+}
+
```
meskill, line 117: we can omit defining default functions in cases when the default is equal to the type's default as it is for Option and bool.

for attribute it will be just `#[serde(default)]`

src/config/server.rs
```diff
@@ -17,8 +17,9 @@ pub struct Server {
   pub enable_response_validation: Option<bool>,
   pub global_response_timeout: Option<i64>,
   pub port: Option<u16>,
-  pub proxy: Option<Proxy>,
   pub vars: Option<BTreeMap<String, String>>,
+  #[serde(default, skip_serializing_if = "Option::is_none")]
+  pub upstream: Option<Upstream>,
```
meskill, line 22: now as we marked the property with serde(default) we can remove wrapping with `Option` and leave the type itself. The serde will put the default value in that case if user isn't define it.

Just make sure that `skip_serializing_if` will equal to `is_default` in that case (see examples in other places in repo)

meskill: server operator is not required so changing test undesirable. To remove wrapping and to be able to leave tests as is try to use `skip_serializing_if` option as I've mentioned. Check [the example](https://github.com/tailcallhq/tailcall/blob/458716df62eacd670e8ccd20f9c10160ff752bc4/src/config/config.rs#L211) of it

tusharmath: @akhill10 I think this is a different problem. `Eq` is not implemented.

akhill10: Thanks, fixed.

src/config/server.rs
```diff
@@ -61,7 +63,85 @@ impl Server {
...
+  pub proxy: Option<Proxy>,
+  #[serde(default = "connect_timeout_default")]
+  pub connect_timeout: u64,
+  #[serde(default = "timeout_default")]
+  pub timeout: u64,
+  #[serde(default = "tcp_keep_alive_default")]
+  pub tcp_keep_alive: u64,
+  #[serde(default = "user_agent_default")]
+  pub user_agent: String,
+}
```
tusharmath, line 94: will need to update `.tailcall.graphql` with these settings.

src/http/client.rs
```diff
@@ -33,14 +33,20 @@ impl Default for DefaultHttpClient {
...
-      .user_agent("Tailcall/1.0");
+      .tcp_keepalive(Some(Duration::from_secs(upstream_settings.tcp_keep_alive)))
+      .timeout(Duration::from_secs(upstream_settings.timeout))
+      .connect_timeout(Duration::from_secs(upstream_settings.connect_timeout))
+      .http2_keep_alive_interval(Some(Duration::from_secs(upstream_settings.keep_alive_interval)))
+      .http2_keep_alive_timeout(Duration::from_secs(upstream_settings.keep_alive_timeout))
+      .http2_keep_alive_while_idle(upstream_settings.keep_alive_while_idle)
+      .pool_idle_timeout(Some(Duration::from_secs(upstream_settings.pool_idle_timeout)))
+      .pool_max_idle_per_host(upstream_settings.pool_max_idle_per_host)
+      .user_agent(upstream_settings.user_agent.clone());
```
tusharmath, line 47: Fantastic clean up ❤️

src/http/client.rs
```diff
@@ -33,14 +33,20 @@ impl Default for DefaultHttpClient {
 
 impl DefaultHttpClient {
   pub fn new(server: Server) -> Self {
+    let upstream_settings = &server.upstream;
```
tusharmath, line 36:
```suggestion
    let upstream = &server.upstream;
```
nitpik 🙈

tusharmath:
@akhill10 This looks good to go baring a few clean ups and a rebase.

akhill10:
I think it is good to go.

@tusharmath @meskill Thanks for the reviews and patience with my lame questions as I was new to rust. Learned so many new things during the process. 😄

examples/.tailcallrc.graphql
```diff
@@ -13,6 +13,7 @@ directive @server(
   proxy: Proxy
```
meskill, line 13: I think proxy is now obsolete

src/config/server.rs
```diff
@@ -20,7 +20,9 @@ pub struct Server {
   pub enable_response_validation: Option<bool>,
   pub global_response_timeout: Option<i64>,
   pub port: Option<u16>,
-  pub proxy: Option<Proxy>,
+  #[serde(default)]
+  #[serde(skip_serializing_if = "is_default")]
```
meskill, line 24:
```suggestion
  #[serde(default, skip_serializing_if = "is_default")]
```

You can just do like this

examples/.tailcallrc.graphql
```diff
@@ -13,6 +13,7 @@ directive @server(
...
   vars: [KeyValue]
   batch: Batch
+  upstream: Upstream
```
meskill, line 16: You have to specify input type for this to make it work.

meskill (changes requested):

tusharmath (approved):

examples/.tailcallrc.graphql
```diff
@@ -41,6 +41,19 @@ input Proxy {
   url: String
 }
 
+input Upstream {
+  pool_idle_timeout: Int
+  pool_max_idle_per_host:  Int
```
tusharmath, line 46:
```suggestion
  pool_max_idle_per_host: Int
```

tusharmath:
Thank you once again @akhill10 🙏 
This makes the writing configurations a lot cleaner !

PR #396, opened 2023-10-01
Update order of steps in GH

### LINKED ISSUE #394: Re-order steps on CI
**Problem**
Currently fmt and linting happens before building and running unit tests.

**Solution**
We should run the build first and make sure the code compiles, then run unit tests and in the end run clippy and fmt.

**Update 1**
Linting and Testing can be parallized and the developer can be informed about it independently.

utsav00:
**Summary:**  
- Changes the order of steps running according:
  - Build -> Test -> Code Coverage -> fmt -> Lint

**Issue Reference(s):**  
Fixes #394 
/claim #394

**Build & Testing:**

**Checklist:**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,13 +47,6 @@ jobs:
           command: llvm-cov
           args: --all-features --lcov --output-path lcov.info
 
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
-          files: lcov.info
-          fail_ci_if_error: true
-
       - name: Format code
         uses: actions-rs/cargo@v1
         with:
@@ -67,6 +60,13 @@ jobs:
           toolchain: nightly
           command: clippy
           args: --all-targets --all-features -- -D warnings
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
+          files: lcov.info
+          fail_ci_if_error: true
 
   deploy:
     name: Deploy app
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,19 +8,14 @@ on:
     branches: [main]
     types: [opened, reopened, synchronize]
 jobs:
-  check:
-    name: Rust project
+  test:
+    name: Run tests
     if: (github.event_name == 'pull_request') || (github.event_name == 'push' && github.ref == 'refs/heads/main')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: taiki-e/install-action@cargo-llvm-cov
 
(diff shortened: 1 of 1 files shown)


utsav00:
@tusharmath, confirming if we want the Code Coverage upload after all the steps are executed. That is how it is structured in this PR, cause it didn't make sense to me to not let the linter run in case of code coverage upload failure. But let me know if you think any differently.

Also, can you approve the workflow so I can verify it here again? Updated workflow tested once here - https://github.com/utsav00/tailcall/actions/runs/6369299383/job/17289423954

tusharmath (approved):

.github/workflows/ci.yml
```diff
@@ -36,6 +36,17 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
+      - name: build
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
```
tusharmath, line 42: I think test will automatically build. we probably don't need this step. What do you think?

tusharmath:
Hey @utsav00 thank you for contributing to tailcall 🙏 

> @tusharmath, confirming if we want the Code Coverage upload after all the steps are executed. That is how it is structured in this PR, cause it didn't make sense to me to not let the linter run in case of code coverage upload failure. But let me know if you think any differently.

That makes perfect sense! Thanks!

.github/workflows/ci.yml
utsav00: Ah right, thanks for noticing this. We can remove the build step since test step will cover it.

However, I added build in case there are tests that do not cover some part of code, in which case it will not compile everything and thus won't throw error for that part (if there are any issues). I see that there is ~82.5% coverage but just wanted to confirm some component is not being tested fully before I go ahead and remove it.

tusharmath:
@utsav00 I posted this on twitter — https://twitter.com/tusharmath/status/1708813317404520529

I think there consensus in doing it in parallel, do you mind making that change? I will update the github issue.

utsav00:
Interesting. Then the idea is to have two jobs, one for testing and one for format and lint checks, and the deploy and release should still be dependent on both of these two jobs, right?

tusharmath:
@utsav00 Yes, that would be the best! Thanks :)

utsav00:
Can you review this again @tusharmath please? Tested this here - https://github.com/utsav00/tailcall/actions/runs/6396355226/job/17361951318

tusharmath (approved):

tusharmath (approved):

tusharmath:
Thanks @utsav00! This will truly improve the dev-ex of our contributors 🙏

tusharmath:
@utsav00 It seems like nothing is building on main. Can you have a look.

Screenshot 2023-10-05 at 12 49 13 PM

utsav00:
Hey @tusharmath, I just got time to look into this, apologies for delay. Observing the actions, it seems like the workflow is running well. Did you see a recurrence of this exact issue again from the screenshot? I can look into this if this is a repeating behavior

tusharmath:
Yes working now! Thanks once again.

Issue #347, opened 2023-09-27
Optimize impl `PathString` for `EvaluationContext`


tusharmath:
**Problem**
Implementation for `PathString` for `EvaluationContext` can be optimized

**Solution**
`path_string` function for `EvaluationContext` does quite a bit of cloning. We should write extensive benchmarks to setup baseline performance and then optimize it's implementation for `EvaluationContext`.

PR #404, opened 2023-10-02
Performance improve

Nirali-Ahir:
Reduced Cloning, Optimized Pattern Matching
Added Benchmarks

fix: #347
/claim #347


benches/path_string_evaluationcontext_bench.rs
```diff
@@ -0,0 +1,120 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use tailcall::http::RequestContext;
+use tailcall::lambda::EvaluationContext;
+use tailcall::path_string::PathString;
+
+fn path_string_for_evolution_context(c: &mut Criterion) {
+  let req_ctx = RequestContext::default();
+  let eval_ctx = EvaluationContext::new(&req_ctx);
```
meskill, line 8: I think testing default bare value of EvaluationContext might be not fully representative for the benchmark purposes as during testing we measure only cases when there is no value under the path and all the internal data is null and that will quick bailout to None ignoring most of the code of the impl itself.

I've tried to implement such cases but it's little tricky because some of the contexts require async_graphql stuff that is hard to instantiate easily in tests.

Nirali-Ahir: i am planning to add more benchmarks 

- Finding vs. not finding a path
- Short vs. long paths
- Various data types
- Different headers/args/vars paths

tusharmath: Awesome! Let us know @Nirali-Ahir if there is anyway we can help.

tusharmath:
Closing in favour of #347

Issue #348, opened 2023-09-27
Optimize `RequestTemplate::to_request`


tusharmath:
**Problem**
The `to_request` implementation is supposed to be called for every http call. We want to make sure that it's as fast as possible.

**Solution**
We need to perform a deep analysis of how it can be implemented in the fastest way possible. We write benchmarks to showcase current performance of the function and then once optimized show % improvements.

**Technical Requirements**
- Write benchmarks and optimize `RequestTemplate::to_request`.
- Showcase initial performance number and numbers after optimizations.

**Additional Notes:**
- Performance can be improved by reducing allocations and using references.
- Reducing overall compute requirements.
- Pre-computing the request from a template in scenarios where it's possible.

mahmoudfarouq:
Hi, can you please elaborate more on what kind of optimizations do you have in mind?

tusharmath:
@mahmoudfarouq Updated the issue. Hope it makes sense, feel free to post questions on the discord channel for a faster response time :)

tusharmath:
fixed in #407

PR #407, opened 2023-10-03
performance: improve performance of `to_request`

digital-phoenix:
I improved to_request by 30% by editing the eval_url function

/claim https://github.com/tailcallhq/tailcall/issues/348

Benchmark results:
Benchmarking old_to_request: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.1s, enable flat sampling, or reduce sample count to 60.
old_to_request          time:   [966.76 µs 973.95 µs 982.18 µs]
                        change: [-9.8271% -8.7950% -7.6926%] (p = 0.00 < 0.05)
                        Performance has improved.

new_to_request          time:   [671.69 µs 676.29 µs 680.92 µs]
                        change: [-10.082% -7.5743% -5.3185%] (p = 0.00 < 0.05)
                        Performance has improved.

**Issue Reference(s):**  
Improvement for issue #348

**Build & Testing:**

- [x] Run benchmark request_old_vs_new.rs

**Checklist:**
I replaced the old to_request with the new one and have run all of the tests and they all passed
Ran lint 
Ran bench to verify improvements


tusharmath:
Fantastic @digital-phoenix This looks really promising! Looking forward to this PR 🙏

digital-phoenix:
@tusharmath I believe this has now been completed. Let me know if you have any questions or concerns with it.

digital-phoenix:
@tusharmath is there something I still need to do to resolve this issue?

src/request_template.rs
```diff
@@ -314,10 +368,14 @@ mod tests {
...
     let ctx = Context::default();
     let req = tmpl.to_request(&ctx).unwrap();
+    println!("testing post");
     assert_eq!(req.method(), reqwest::Method::POST);
+    println!("testing headers");
     assert_eq!(req.headers().get("foo").unwrap(), "bar");
     let body = req.body().unwrap().as_bytes().unwrap().to_owned();
+    println!("testing body");
     assert_eq!(body, "foo".as_bytes());
+    println!("testing url");
```
tusharmath, line 378: drop all `println`

src/request_template.rs
```diff
@@ -82,6 +119,23 @@ impl RequestTemplate {
     Ok(req)
   }
 
+  // /// A high-performance way to reliably create a request
+  pub fn to_request2<C: PathString + HasHeaders>(&self, ctx: &C) -> anyhow::Result<reqwest::Request> {
```
tusharmath, line 123: RequestTemplate should keep a `Option<Request>` so that the request doesn't need to be created again and again in scenarios where the mustache templates are all literals and don't need to be evaluated again and again. It will still need to be updated based on the ctx, but that would still be an optimization.

digital-phoenix: @tusharmath  How common of a use case is it to have all of the mustache objects be literals?  Also, does mustache currently have a function to tell if a mustache object is composed only of literals and constants?

digital-phoenix: @tusharmath  Actually mustache templates shouldn't interfere with caching. Either the request_to function should just resolve all the mustache templates at the beginning and then check if the inputs have already been cached or to_request shouldn't know it's working on mustache templates in the first place. In order to allow to_request to work on mustache templates without knowing about it you would need to create a mustache iterator and a mustache map. If you had a mustache iterator that could just iterate through a list of mustache templates you could use mustache templates on any function that takes an iterator as input.

src/request_template.rs
digital-phoenix: @tusharmath  I've removed the println statements

src/request_template.rs
tusharmath: > How common of a use case is it to have all of the mustache objects be literals? 

Common for resolvers that are provided at the root of the Query.


> Also, does mustache currently have a function to tell if a mustache object is composed only of literals and constants?

Mustache could implement a method called `is_literal()` which could iterate over the template and see if everything is a literal. 

To your second point, I agree, and this is how we can go about it — 
- Let RequestTemplate (RT) have a `Option<Request>` inside it.
- While creating a RT from an Endpoint, we will check if `is_const()` is true for the `Endpoint` We need `is_const` on `Endpoint` that checks if every mustache template is a literal in the Endpoint.
- If `is_literal` is true, we create a Request AOT and save it inside the RT.
- While calling `to_request` we check if AOT `Request` exists, if it does, return a ref otherwise, create a new one and return. Don't update cache.


This way `to_request` doesn't care about `Mustache` and it's internals. It just performs an AOT on the Endpoint if we know that it's not going to change.

Cargo.toml
```diff
@@ -77,6 +77,10 @@ harness = false
 name = "request_template_bench"
 harness = false
 
+[[bench]]
+name = "request_old_vs_new"
```
tusharmath, line 81:
```suggestion
name = "to_request"
```

src/request_template.rs
```diff
@@ -82,6 +119,23 @@ impl RequestTemplate {
     Ok(req)
   }
 
+  // /// A high-performance way to reliably create a request
```
tusharmath, line 122:
```suggestion
  /// A high-performance way to reliably create a request
```

tusharmath:
@digital-phoenix Let me know if you have any more questions. Looking forward to closing this PR soon!

src/request_template.rs
```diff
@@ -20,6 +20,43 @@ pub struct RequestTemplate {
...
+    let q: Vec<(String, String)> = url
+      .query_pairs()
+      .filter_map(|(k, v)| {
+        if v.is_empty() {
+          None
+        } else {
+          Some((k.to_string(), v.to_string()))
+        }
+      })
+      .collect();
```
tusharmath, line 37: I think we can further optimize by dropping the tuples and collect operations.

tusharmath:
@digital-phoenix I am a bit confused about the status of this PR. Is it complete from your end?

digital-phoenix:
@tusharmath 
The code is fully functional but I still want to clean it up a little. What are your policies on using macros?

tusharmath:
Using is all good. Writing only if there is no other way.

digital-phoenix:
@tusharmath Can I create a function that takes 4 arguments or is that against the lint formatting rules?

tusharmath:
@digital-phoenix We have a lint rule for new code to not go beyond 3 params. Why do you think it's required btw?

digital-phoenix:
@tusharmath I think this is complete now

src/request_template.rs
digital-phoenix: @tusharmath I believe this has been resolved in the latest commit

src/request_template.rs
digital-phoenix: @tusharmath  this has now been added

Cargo.toml
digital-phoenix: @tusharmath  This has now been changed

digital-phoenix:
@tusharmath is there something I need to do to get Codecov to work?

tusharmath:
Can you remove the older implementation of to_request

digital-phoenix:
> Can you remove the older implementation of to_request

I made this change and pushed it. Let me know if there is anything else you need me to do.

digital-phoenix:
@tusharmath is there anything still left to do for this or can you merge it into the main branch?

tusharmath:
It seems like `request_template_bench.rs` has become slower by 24%

Screenshot 2023-10-19 at 4 34 20 PM

digital-phoenix:
> It seems like `request_template_bench.rs` has become slower by 24%
> Screenshot 2023-10-19 at 4 34 20 PM

I've fixed this case and push it now. Since the code currently creates a cached value of the static part of the request and then only adds the dynamic part everytime it is called, creating the endpoint is slower then it used to be. Since the  request_template_bench only has a dynamic root_url and and a static header it does not benefit much from the current caching mechanism. In fact it performed the most expensive part of the function twice every iteration. I added a line of code to avoid caching in the case of a dynamic root_url and in my testing request_template_bench is now faster then the original version.

tusharmath (approved):

tusharmath:
Thanks @digital-phoenix

Issue #373, opened 2023-09-29
add a `const` directive with a const value


tusharmath:
**Description**
A `const` directive would basically always resolve with a constant value. The value will be provided within the directive. This is useful to writing tests and sometimes exposing constant values.

**Usage**
```graphql
schema {
  query: Query
}

type Query {
  user: User @const(data: {name: "John", age: 12})
}

type User {
  name: String
  age: Int
}
```

**Expected Behaviour**
- The provided `data` should always match the shape of the type provided. For example 
  
  **correct**
  ```graphql
  type Query {
    user: User @const(data: {name: "John", age: 12})
  }
  
  type User {
    name: String!
    age: Int!
  }
  ```
  
  **wrong**
  ```graphql
  type Query {
    user: User @const(data: {name: "John"})
  }
  
  type User {
    name: String!
    age: Int!
  }
  ```

**NOTE**
- Make sure existing functionality doesn't break
- Add tests in the `tests/` dir.

tusharmath:
@akansh-bansal Updated the expected behaviour — 

> The output schema should mark the type of the field as required.

amitksingh1490:
Fixed by #408

PR #408, opened 2023-10-03
feat: add const directive

heyrutvik:
**Summary:**  
It adds `@const` directive to set const value for test and other purposes. `const` also makes sure that provided `data` conforms with the type `T`.

**Build & Testing:**

**Checklist:**

/claim #373

### DIFF
--- a/src/blueprint/from_config.rs
+++ b/src/blueprint/from_config.rs
@@ -5,6 +5,7 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use async_graphql::parser::types::ConstDirective;
 #[allow(unused_imports)]
 use async_graphql::InputType;
+use async_graphql_value::ConstValue;
 use hyper::header::{HeaderName, HeaderValue};
 use hyper::HeaderMap;
 use regex::Regex;
@@ -16,6 +17,7 @@ use crate::config::{Arg, Config, Field, InlineType};
 use crate::directive::DirectiveCodec;
 use crate::endpoint::Endpoint;
 use crate::json::JsonSchema;
+use crate::lambda::Expression::Literal;
 use crate::lambda::Lambda;
 use crate::request_template::RequestTemplate;
 use crate::valid::{OptionExtension, Valid as ValidDefault, ValidExtensions, ValidationError, VectorExtension};
@@ -187,6 +189,11 @@ fn to_field(
   name: &str,
   field: &Field,
 ) -> Valid<Option<blueprint::FieldDefinition>> {
+  let directives = field.resolvable_directives();
+  if directives.len() > 1 {
+    return Valid::fail(format!("Multiple resolvers detected [{}]", directives.join(", ")));
+  }
+
   let field_type = &field.type_of;
   let args = to_args(field)?;
 
@@ -201,6 +208,7 @@ fn to_field(
 
   let field_definition = update_http(field, field_definition, config).trace("@http")?;
   let field_definition = update_unsafe(field.clone(), field_definition);
+  let field_definition = update_const_field(field, field_definition, config).trace("@const")?;
   let field_definition = update_inline_field(type_of, field, field_definition, config).trace("@inline")?;
   let maybe_field_definition = update_modify(field, field_definition, type_of, config).trace("@modify")?;
   Ok(maybe_field_definition)
@@ -283,8 +291,7 @@ fn update_unsafe(field: config::Field, mut b_field: FieldDefinition) -> FieldDef
   b_field
 }
 
-fn update_http(field: &config::Field, b_field: FieldDefinition, config: &Config) -> Valid<FieldDefinition> {
-  let mut b_field = b_field;
+fn update_http(field: &config::Field, mut b_field: FieldDefinition, config: &Config) -> Valid<FieldDefinition> {
   match field.http.as_ref() {
     Some(http) => match http
       .base_url
@@ -358,8 +365,23 @@ fn update_modify(
     None => Valid::Ok(Some(b_field)),
   }
 }
-fn needs_resolving(field: &config::Field) -> bool {
-  field.unsafe_operation.is_some() || field.http.is_some()
+fn update_const_field(field: &config::Field, mut b_field: FieldDefinition, config: &Config) -> Valid<FieldDefinition> {
+  match field.const_field.as_ref() {
+    Some(const_field) => {
+      let data = const_field.data.to_owned();
+      match ConstValue::from_json(data.to_owned()) {
+        Ok(gql_value) => match to_json_schema_for_field(field, config).validate(&gql_value) {
+          Ok(_) => {
+            b_field.resolver = Some(Literal(data));
+            Valid::Ok(b_field)
+          }
+          Err(err) => err.into(),
+        },
+        Err(e) => Valid::fail(format!("invalid JSON: {}", e)),
+      }
+    }
+    None => Valid::Ok(b_field),
+  }
 }
 fn is_scalar(type_name: &str) -> bool {
   ["String", "Int", "Float", "Boolean", "ID", "JSON"].contains(&type_name)
@@ -424,11 +446,15 @@ fn process_field_within_type(
   invalid_path_handler: &dyn Fn(&str, &[String]) -> Valid<Type>,
 ) -> Valid<Type> {
   if let Some(next_field) = type_info.fields.get(field_name) {
-    if needs_resolving(next_field) {
+    if next_field.has_resolver() {
       return Valid::<Type>::validate_or(
         Valid::fail(format!(
           "Inline can't be done because of {} resolver at [{}.{}]",
-          next_field.http.as_ref().map(|_| "http").unwrap_or_else(|| "unsafe"),
+          {
+            let next_dir_http = next_field.http.as_ref().map(|_| "http");
+            let next_dir_const = next_field.const_field.as_ref().map(|_| "const");
+            next_dir_http.or(next_dir_const).unwrap_or("unsafe")
+          },
           field.type_of,
           field_name
         )),
--- a/src/valid/cause.rs
+++ b/src/valid/cause.rs
@@ -35,6 +35,16 @@ impl<E> Cause<E> {
   }
 }
 
+impl Cause<&str> {
+  pub fn to_owned(&self) -> Cause<String> {
+    Cause {
+      message: self.message.to_owned(),
+      description: self.description.map(|d| d.to_owned()),
+      trace: self.trace.iter().map(|t| t.to_owned()).collect(),
+    }
+  }
+}
+
 #[cfg(test)]
 mod tests {
   use std::collections::VecDeque;
--- a/tests/graphql/errors/test-const.graphql
+++ b/tests/graphql/errors/test-const.graphql
@@ -1,0 +1,17 @@
+#> server-sdl
+schema @server(baseURL: "https://jsonplaceholder.typicode.com") {
+  query: Query
+}
+
+type User {
+  name: String
+  age: Int!
+}
+
+type Query {
+  user: User @const(data: {name: "John"})
+}
+
+#> client-sdl
+type Failure
+@error(message: "expected field to be non-nullable", trace: ["Query", "user", "@const"])
--- a/tests/graphql/passed/test-const-nullable.graphql
+++ b/tests/graphql/passed/test-const-nullable.graphql
@@ -1,0 +1,21 @@
+#> server-sdl
+schema @server(baseURL: "https://jsonplaceholder.typicode.com") {
+  query: Query
+}
+
+type User {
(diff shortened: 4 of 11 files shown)


tests/graphql/errors/test-const.graphql
```diff
@@ -0,0 +1,17 @@
...
+    query: Query
+}
+
+type User {
+    name: String
+    age: Int
+}
+
+type Query {
+    user: User @const(data: {name: "John"})
```
amitksingh1490, line 12: This should only fail if age is a required feild

src/blueprint/from_config.rs
```diff
@@ -202,6 +204,7 @@ fn to_field(
   let field_definition = update_http(field, field_definition, config).trace("@http")?;
   let field_definition = update_unsafe(field.clone(), field_definition);
   let field_definition = update_inline_field(type_of, field, field_definition, config).trace("@inline")?;
+  let field_definition = update_const_field(field, field_definition, config).trace("@const")?;
```
amitksingh1490, line 207: Please ensure comprehensive test coverage when combining @const, @inline, and @modify decorators. No matter the order of their application, the result should remain consistent.

Add tests to confirm that when @const and @unsafe are used together, @unsafe can still leverage the constant value for data manipulation.

We need tests to demonstrate that the combination of @const and @http should fail.

src/blueprint/from_config.rs
```diff
@@ -358,6 +361,24 @@ fn update_modify(
...
+fn update_const_field(field: &config::Field, mut b_field: FieldDefinition, config: &Config) -> Valid<FieldDefinition> {
+  match field.const_field.as_ref() {
+    Some(const_field) => {
+      let data = &const_field.data;
+      if let Ok(json) = ConstValue::from_json(data.to_owned()) {
+        if to_json_schema_for_field(field, config).validate(&json).is_ok() {
+          b_field.resolver = Some(Literal(data.to_owned()));
+          Valid::Ok(b_field)
+        } else {
+          Valid::fail(format!("const data doesn't conform with the type `{}`.", field.type_of))
```
amitksingh1490, line 373: error message should show what part of the data don't confirm to the schema.

amitksingh1490 (dismissed):

src/blueprint/from_config.rs
tusharmath: `validate` actually returns a validation error, which is very detailed. I think we can just return those errors directly.

src/blueprint/from_config.rs
amitksingh1490: refer this
https://github.com/tailcallhq/tailcall/blob/main/tests/graphql/passed/test-inline-with-modify-field.graphql
and this for faliures
https://github.com/tailcallhq/tailcall/blob/main/tests/graphql/errors/test-inline.graphql

tests/graphql/errors/test-const.graphql
heyrutvik: Covered this comment in new commit.

src/blueprint/from_config.rs
heyrutvik: Covered this remarks in new commit.

src/blueprint/from_config.rs
```diff
@@ -20,11 +21,22 @@ use crate::json::JsonSchema;
...
 type Valid<A> = ValidDefault<A, String>;
 
+impl<A> From<ValidationError<&str>> for Valid<A> {
+  fn from(value: ValidationError<&str>) -> Self {
+    let message = value
+      .as_vec()
+      .iter()
+      .map(|c| Cause::new(c.message.to_owned()))
+      .collect();
+    Valid::fail_cause(message)
```
heyrutvik, line 36: @amitksingh1490 this impl seem fishy. Let me know if you have any suggestions to improve it.

heyrutvik: The main issue with it is that creating a cause here ignores description and trace. 

I'm thikning something like following,
```diff
 use crate::lambda::Expression::Literal;
 use crate::lambda::Lambda;
 use crate::request_template::RequestTemplate;
-use crate::valid::{Cause, OptionExtension, Valid as ValidDefault, ValidExtensions, ValidationError, VectorExtension};
+use crate::valid::{OptionExtension, Valid as ValidDefault, ValidExtensions, ValidationError, VectorExtension};
 use crate::{blueprint, config};
 
 type Valid<A> = ValidDefault<A, String>;
 
 impl<A> From<ValidationError<&str>> for Valid<A> {
   fn from(value: ValidationError<&str>) -> Self {
-    let message = value
-      .as_vec()
-      .iter()
-      .map(|c| Cause::new(c.message.to_owned()))
-      .collect();
+    let message = value.as_vec().iter().map(|c| c.clone().into()).collect();
     Valid::fail_cause(message)
   }
 }
diff --git a/src/valid/cause.rs b/src/valid/cause.rs
index bf5b200..fb8a6c3 100644
--- a/src/valid/cause.rs
+++ b/src/valid/cause.rs
@@ -29,6 +29,16 @@ impl<E: Display> Display for Cause<E> {
   }
 }
 
+impl From<Cause<&str>> for Cause<String> {
+  fn from(value: Cause<&str>) -> Self {
+    Cause {
+      message: value.message.to_owned(),
+      description: value.description.map(|d| d.to_owned()),
+      trace: value.trace,
+    }
+  }
+}
+
 impl<E> Cause<E> {
   pub fn new(e: E) -> Self {
     Cause { message: e, description: None, trace: VecDeque::new() }
```

How about it? :thinking:

amitksingh1490: Ya this seems ok, Please go ahead and add a test for this change

heyrutvik: Thanks, @amitksingh1490. Sorry but the one suggested in the comment or the existing code? :slightly_smiling_face: 

Also, test for the change is already added.

heyrutvik: I think you meant the one suggested in the comment.

tests/graphql/passed/test-const-nullable.graphql
```diff
@@ -0,0 +1,21 @@
...
+    name: String
+    age: Int
+}
+
+type Query {
+    user: User @const(data: {name: "John"})
+}
+
+#> client-query
+query @expect(json: {data:{user:{name: "John", age: null}}}) {
```
heyrutvik, line 16: @amitksingh1490 here is the test for nullable field resoved to `null`

tests/graphql/errors/test-const.graphql
```diff
@@ -14,4 +14,4 @@ type Query {
 
 #> client-sdl
 type Failure
-@error(message: "const data doesn't conform with the type `User`.", trace: ["Query", "user", "@const"])
+@error(message: "expected field", trace: ["Query", "user", "@const"])
```
heyrutvik, line 17: And failing case for non-nullable is here.

src/blueprint/from_config.rs
heyrutvik: Convered with tests in recent commit.

src/blueprint/from_config.rs
```diff
@@ -16,13 +18,21 @@ use crate::config::{Arg, Config, Field, InlineType};
...
 use crate::{blueprint, config};
 
 type Valid<A> = ValidDefault<A, String>;
 
+impl<A> From<ValidationError<&str>> for Valid<A> {
+  fn from(value: ValidationError<&str>) -> Self {
+    let message = value.as_vec().iter().map(|c| c.clone().into()).collect();
+    Valid::fail_cause(message)
+  }
+}
```
tusharmath, line 34: Move to `valid.rs`.

src/blueprint/from_config.rs
```diff
@@ -358,8 +373,41 @@ fn update_modify(
...
+        Valid::fail("@const data: invalid graphql value".to_owned())
+      }
+    }
+    None => Valid::Ok(b_field),
+  }
+}
 fn needs_resolving(field: &config::Field) -> bool {
-  field.unsafe_operation.is_some() || field.http.is_some()
+  field.unsafe_operation.is_some() || field.http.is_some() || field.const_field.is_some()
 }
```
tusharmath, line 411: Now that we have `resolvable_directives` I don't think we need `needs_resolving`

src/blueprint/from_config.rs
```diff
@@ -428,7 +476,11 @@ fn process_field_within_type(
       return Valid::<Type>::validate_or(
         Valid::fail(format!(
           "Inline can't be done because of {} resolver at [{}.{}]",
-          next_field.http.as_ref().map(|_| "http").unwrap_or_else(|| "unsafe"),
+          {
+            let next_dir_http = next_field.http.as_ref().map(|_| "http");
+            let next_dir_const = next_field.const_field.as_ref().map(|_| "const");
+            next_dir_http.or(next_dir_const).unwrap_or("unsafe")
+          },
```
tusharmath, line 483: Showing all the resolvers in one error will be a better experience for the user.

heyrutvik: What do you suggest error message might look like?

heyrutvik:
```
type Query {
  post: Post @http(path: "/posts/1") @http(path: "/users/{{value.userId}}") @inline(path: ["user", "name"])
}

type Post {
  userId: Int
}
```

@tusharmath, I could be wrong but from usecase, it looks like only one resolvable directive will be on the path?

src/blueprint/from_config.rs
```diff
@@ -358,8 +373,41 @@ fn update_modify(
...
+            if let Some(field) = typ.fields.get(key) {
+              let json_has_key = map.contains_key(key);
+              let field_required = field.required;
+              if !json_has_key && !field_required {
+                map.insert(key.to_owned(), Value::Null);
+              }
+            }
+          }
+        }
+      }
```
tusharmath, line 393: Two questions 
- Why do we need this? If the value isn't available, sending `null` or not having the key itself is morally equivalent.
- Wouldn't we have to do this recursively?

src/blueprint/from_config.rs
heyrutvik: Makes sense. We willl remove `needs_resolving` and simply call `has_resolver` instace method.

src/blueprint/from_config.rs
```diff
@@ -358,8 +373,41 @@ fn update_modify(
...
+      if let Ok(gql_value) = ConstValue::from_json(data.to_owned()) {
+        match to_json_schema_for_field(field, config).validate(&gql_value) {
+          Ok(_) => {
+            b_field.resolver = Some(Literal(data.to_owned()));
+            Valid::Ok(b_field)
+          }
+          Err(err) => err.into(),
+        }
+      } else {
+        Valid::fail("@const data: invalid graphql value".to_owned())
```
tusharmath, line 403: Not a very useful error message. It should say invalid JSON or something, and should use the error information available in the `Err`.

src/valid/cause.rs
```diff
@@ -29,6 +29,16 @@ impl<E: Display> Display for Cause<E> {
   }
 }
 
+impl From<Cause<&str>> for Cause<String> {
```
tusharmath, line 32:
```suggestion
impl<E: Display> From<Cause<E>> for Cause<E> {
```

This might be a more generic way to implement From.

tusharmath:
This is fantastic! Just a few nitpicks 🙏

src/blueprint/from_config.rs
heyrutvik: The code handles the following remark, https://github.com/tailcallhq/tailcall/pull/408#discussion_r1343552654. 

1. I think `null` is required. Here is the example,
```
type User {
    name: String
    age: Int
}

type Query {
    user: User @const(data: {name: "John"})
}
```
And query is asking for `age`,
```
user {
  age
  name
}
```
There could be better way to do this as I understood the original remark, it make sense. 

2. Now I think about it, I'll need to check. We currently have flat JSON so will add test for object containing object. But will wait for your reponse on the first answer/explanation. 

@tusharmath

tests/graphql/errors/test-const.graphql
```diff
@@ -0,0 +1,17 @@
...
+    age: Int!
+}
+
+type Query {
+    user: User @const(data: {name: "John"})
+}
+
+#> client-sdl
+type Failure
+@error(message: "expected field", trace: ["Query", "user", "@const"])
```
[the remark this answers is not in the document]
heyrutvik: I'll look into it.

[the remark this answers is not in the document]
heyrutvik: @tusharmath I think this will require but refactoring. Would it make sense to keep it out of the scope?

amitksingh1490 (approved):

amitksingh1490:
fixes #373

PR #428, opened 2023-10-06
refactor: from document with better error handling in @http and @server

### LINKED ISSUE #429: Refactor KeyValues struct
KeyValues are currently part of `config.rs`. There are few changes which are required in this

 

- [ ] Move KeyValue and KeyValues to a separate file `key_values.rs` inside config mod
- [ ]  Implement to_map which will return BtreeMap<String, String> in keyValues
- [ ]  Drop `From<KeyValues>`

amitksingh1490:
**Summary:**  
If map is passed to http @query or @headers was getting error producing a ambiguous error
` no resolver is found`. It should fail with type mismatch error
Fixes #429  as well

**Build & Testing:**

**Checklist:**


src/path_string.rs
```diff
@@ -31,7 +32,17 @@ impl PathString for EvaluationContext<'_> {
...
+            .req_ctx
+            .server
+            .vars
+            .clone()
+            .map(|x| {
+              let b: BTreeMap<String, String> = x.into();
+              b
+            })?
+            .get(&tail[0])
+            .cloned()?,
```
tusharmath, line 45: This will slow down our existing implementation. Can we consider optimizing it or atleast match the performance from main.
- [ ] Post wrk benchmarks before and after this change as a PR.

tests/graphql/passed/test-resolve-with-vars.graphql
```diff
@@ -1,5 +1,5 @@
 #> server-sdl
-schema @server(vars: {id: "1"}) {
+schema @server(vars: [{key: "id",value: "1"}]) {
```
meskill, line 2: I wonder why to change the public api in how the vars are defined?

src/config/from_document.rs
```diff
@@ -273,14 +273,14 @@ fn to_inline(directives: &[Positioned<ConstDirective>]) -> Option<config::Inline
...
-fn to_http(directives: &[Positioned<ConstDirective>]) -> Option<config::Http> {
-  directives.iter().find_map(|directive| {
+fn to_http(directives: &[Positioned<ConstDirective>]) -> Valid<Option<config::Http>> {
+  let mut http = Valid::Ok(None);
+  for directive in directives {
     if directive.node.name.node == "http" {
-      config::Http::from_directive(&directive.node).ok()
-    } else {
-      None
+      http = Http::from_directive(&directive.node).map(Some)
```
meskill, line 280: It might be not a big deal, but using for in this way will iterate over all the values despite if http has been found at the start. While previous code with `find_map` will stop on first entry and return the value.

To mitigate this we could just use return here on first match

tests/graphql/passed/test-resolve-with-vars.graphql
tusharmath: We couldn't do this :

```graphql
@server(vars: {this-is-id: "1"})
```

But now we can:

```graphql
@server(vars: [{key: "this-is-id", value: "1"}])
```

meskill: makes sense

amitksingh1490: this is needed to support Intellisense in the GraphQL SDL;

src/config/server.rs
```diff
@@ -1,8 +1,12 @@
-use std::collections::{BTreeMap, HashSet};
+use std::collections::HashSet;
 
 use derive_setters::Setters;
 use serde::{Deserialize, Serialize};
 
+use crate::config::KeyValues;
+fn is_default<T: Default + Eq>(val: &T) -> bool {
+  *val == T::default()
+}
```
tusharmath, line 9: duplicate function.

tusharmath (approved):

PR #451, opened 2023-10-09
feat: group by endpoint

### LINKED ISSUE #448: Support Custom Batching
Sometimes batching support is provided by services directly for example in on **jsonplaceholder**

**GET One**
```curl
GET https://jsonplaceholder.typicode.com/users/1
```

**GET Many**
```curl
GET https://jsonplaceholder.typicode.com/users?id=1&id=2&id=3
```

When enabled this can reduce the load upstream drastically. We could make One Http call instead of many to load data for multiple user-ids.

**Approach**
- Create a DL (Data Loader) corresponding to each endpoint. This needs to be created once at the time of server start.
- All batchable requests (GET, DELETE etc.) should go via their corresponding Data Loader.
- Endpoint will need to maintain batching information which is how is the URL going to be generated if it's a call for many ids.
- DL should use the information in the endpoint about batching and perform additional operations if the endpoint supports batching.

**DSL Changes**
```graphql
type Query {
  post(id: Int): Post @http(path: "/posts/:id")
}

type Post {
  title: String
  userId: User

  user: User 
    @http(path: "/users", query: [{key: "id", value: "{{value.userId}}"}]) 
    @batch(key: "userId", path: ["id"])
}
type User { ... }
```

**Technical Requirements**
  _Option 1_
  - We can generate `endpoint_id: usize` while creating the blueprint, and use endpoint id to perform a lookup in an Array instead of using a hashmap. This would save us the cost of creating a hash for each upstream call.
  - Generated DLs should be saved in ServerContext.
  
  _Option 2_
  - We could also consider saving the DL in `Expression::Endpoint` with the RequestTemplate.
  - The fact that we use the DL or not should remain with the `Expression::eval`  for now.

**Additional Info:**
- Currently we have one data loader shared across all endpoint. This increases thread contention across HTTP requests. With one DL per endpoint, thread contention will also reduce 🤞 resulting in a little better performance when enabled.

**Limitations:**
- Batching is only supported for GET requests with query params only.

amitksingh1490:
Fixes #448
**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/blueprint/from_config.rs
+++ b/src/blueprint/from_config.rs
@@ -230,7 +230,7 @@ fn to_field(
   };
 
   let field_definition = update_http(field, field_definition, config).trace("@http")?;
-  let field_definition = update_batch(field, field_definition, config).trace("@groupBy")?;
+  let field_definition = update_group_by(field, field_definition, config).trace("@groupBy")?;
   let field_definition = update_unsafe(field.clone(), field_definition);
   let field_definition = update_const_field(field, field_definition, config).trace("@const")?;
   let field_definition = update_inline_field(type_of, field, field_definition, config).trace("@inline")?;
@@ -315,7 +315,7 @@ fn update_unsafe(field: config::Field, mut b_field: FieldDefinition) -> FieldDef
   b_field
 }
 
-fn update_batch(field: &config::Field, mut b_field: FieldDefinition, _config: &Config) -> Valid<FieldDefinition> {
+fn update_group_by(field: &config::Field, mut b_field: FieldDefinition, _config: &Config) -> Valid<FieldDefinition> {
   if let Some(batch) = field.group_by.as_ref() {
     if let Some(http) = field.http.as_ref() {
       if http.method != Method::GET {
--- a/src/blueprint/from_config.rs
+++ b/src/blueprint/from_config.rs
@@ -3,7 +3,6 @@
 use std::collections::{BTreeMap, HashMap, HashSet};
 
 use async_graphql::parser::types::ConstDirective;
-use async_graphql::BatchRequest::Batch;
 #[allow(unused_imports)]
 use async_graphql::InputType;
 use async_graphql_value::ConstValue;
@@ -23,7 +22,7 @@ use crate::lambda::Expression::Literal;
 use crate::lambda::Lambda;
 use crate::request_template::RequestTemplate;
 use crate::valid::{OptionExtension, Valid as ValidDefault, ValidExtensions, ValidationError, VectorExtension};
-use crate::{batch, blueprint, config};
+use crate::{blueprint, config};
 
 type Valid<A> = ValidDefault<A, String>;
 
--- a/src/blueprint/from_config.rs
+++ b/src/blueprint/from_config.rs
@@ -19,7 +19,7 @@ use crate::endpoint::Endpoint;
 use crate::http::Method;
 use crate::json::JsonSchema;
 use crate::lambda::Expression::Literal;
-use crate::lambda::Lambda;
+use crate::lambda::{Expression, Lambda, Operation};
 use crate::request_template::RequestTemplate;
 use crate::valid::{OptionExtension, Valid as ValidDefault, ValidExtensions, ValidationError, VectorExtension};
 use crate::{blueprint, config};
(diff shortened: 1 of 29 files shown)


src/config/config.rs
```diff
@@ -165,7 +165,7 @@ pub struct Field {
   pub http: Option<Http>,
   #[serde(rename = "unsafe")]
   pub unsafe_operation: Option<Unsafe>,
-  pub batch: Option<Batch>,
+  pub batch: Option<GroupBy>,
```
tusharmath, line 168:
```suggestion
  pub group_by: Option<GroupBy>,
```

src/request_template.rs
```diff
@@ -17,6 +17,7 @@ pub struct RequestTemplate {
   pub headers: Vec<(String, Mustache)>,
   pub body: Option<Mustache>,
   pub endpoint: Endpoint,
+  pub id: Option<usize>,
```
tusharmath, line 20:
```suggestion
  pub id: usize,
```

src/http/server_context.rs
```diff
@@ -4,23 +4,70 @@ use async_graphql::dataloader::{DataLoader, NoCache};
...
+  for def in blueprint.definitions.iter() {
+    if let Definition::ObjectTypeDefinition(def) = def {
+      for field in &def.fields {
+        if let Some(Expression::Unsafe(Operation::Endpoint(_))) = &field.resolver {
+          let mut data_loader = Arc::new(
+            HttpDataLoader::new(DefaultHttpClient::default(), None)
+              .to_data_loader(server.batch.clone().unwrap_or_default()),
+          );
+          field.directives.iter().for_each(|directive| {
+            if directive.name == "batch" {
```
tusharmath, line 49: Batching info can remain inside of the Operation::Endpoint.

src/http/server_context.rs
```diff
@@ -4,23 +4,70 @@ use async_graphql::dataloader::{DataLoader, NoCache};
...
+  blueprint: &Blueprint,
+  server: Server,
+) -> Vec<Arc<DataLoader<HttpDataLoader<DefaultHttpClient>, NoCache>>> {
+  let mut data_loaders = Vec::new();
+  for def in blueprint.definitions.iter() {
+    if let Definition::ObjectTypeDefinition(def) = def {
+      for field in &def.fields {
+        if let Some(Expression::Unsafe(Operation::Endpoint(_))) = &field.resolver {
+          let mut data_loader = Arc::new(
+            HttpDataLoader::new(DefaultHttpClient::default(), None)
```
tusharmath, line 45: FIX: client should be passed and not created.

tusharmath: Can enrich the endpoint expression with the DL directly creating a new Blueprint.

src/request_template.rs
tusharmath: id is not required, if we save the DL in the expr.

tests/graphql_spec.rs
```diff
@@ -194,19 +194,15 @@ async fn test_execution() -> std::io::Result<()> {
     let mut config = Config::from_sdl(&spec.server_sdl).unwrap();
     config.server.enable_query_validation = Some(false);
 
-    let blueprint = Blueprint::try_from(&config).unwrap();
-    let schema = blueprint.to_schema(&config.server);
+    let mut blueprint = Blueprint::try_from(&config).unwrap();
+    let server_ctx = ServerContext::new(&mut blueprint, config.server.clone());
+    let schema = server_ctx.schema.clone();
```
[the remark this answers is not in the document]
amitksingh1490: will do in separate PR

src/blueprint/from_config.rs
```diff
@@ -313,6 +315,29 @@ fn update_unsafe(field: config::Field, mut b_field: FieldDefinition) -> FieldDef
   b_field
 }
 
+fn update_batch(field: &config::Field, mut b_field: FieldDefinition, _config: &Config) -> Valid<FieldDefinition> {
```
tusharmath, line 318:
```suggestion
fn update_group_by(field: &config::Field, mut b_field: FieldDefinition, _config: &Config) -> Valid<FieldDefinition> {
```

src/lambda/expression.rs
```diff
@@ -73,17 +100,16 @@ impl Expression {
...
               // Attempt to short circuit GET request
               if is_get && ctx.req_ctx.server.batch.is_some() {
                 let headers = ctx.req_ctx.server.batch.clone().map(|s| s.headers).unwrap_or_default();
                 let endpoint_key = crate::http::DataLoaderRequest::new(req, headers);
-                let resp = ctx
-                  .req_ctx
-                  .data_loader
+                let resp = dl
                   .as_ref()
+                  .ok_or(EvaluationError::IOException("DataLoader not found".to_string()))?
```
tusharmath, line 112: We can safely unwrap over here. This is an unexpected error.

tests/graphql_spec.rs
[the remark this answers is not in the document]
tusharmath: please create a github issue for this.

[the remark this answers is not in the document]
amitksingh1490: Here: https://github.com/tailcallhq/tailcall/issues/464

tusharmath (approved):

PR #403, opened 2023-10-02
feature: add `tc init` command

neo773:
This PR adds init command to bootstrap a project

https://github.com/tailcallhq/tailcall/assets/62795688/3fc0d30b-e3d7-41a0-9990-e6508b428214

**Issue Reference(s):**  
Fixes #402

**Build & Testing:**

- [x] I ran `cargo test` successfully.
- [x] I have run `./lint.sh` to address and fix linting issues.

/claim #402

PR #458, opened 2023-10-10
feat: add tests for the CLI

### LINKED ISSUE #450: Add unit tests for CLI
CLI piece has no unit tests. We need to re-structure the code to be able to write unit tests in the following manner:

**Technical Requirements:**
- Implementation should include sending a CLI Command and asserting what's printed on the Console.
- Avoid using mocks for testing, depend on pure functions as much as possible.
- Achieve 100% coverage for the CLI modules.

## Proposed Architecture

We will take inspiration from the [ELM architecture] for writing CLI tests.

[ELM architecture]: https://guide.elm-lang.org/architecture/

The main idea for Elm is 

We will port it to Rust in the following way

```rust

/// Incoming messages
enum Message {
  ServerStart {file_paths: Vec<String>},
  Init
}

/// Represents side-effects that need to be performed by the program
/// Simple dumb commands that are part of the 
enum Command {
  ReadFile {path: String},
  WriteFile {path: String},
  StartServer {port: Int}
}

/// State of the application
struct State {
  // ... fields relevant to our use case
}


/// Our main application contains two components - update and command
struct App<M, S, C> {
  /// Takes in a message and a state and optionally returns a new state
  pub update: Fn(M, S) -> Option<S>,

  /// Takes in a message and a state and optionally returns a command for the 
  /// driver code to execute
  pub command: Fn(M, S) -> Option<C>
}


/// Effect running code
async fn driver(app: App<Message, Model, Command>, s1: State, m1: Message) -> Result<(), String> {
   let s2 = app.update(m1, s1);
   let cmd = app.command(m1, s1);
   let messages = execute_effects(cmd).await?
   for m in messages {
     driver(app, s2, m).await?;
   }
}
```

## Elm Architecture
The idea of ELM architecture is to write, composable applications that are purely functional and perform no side effects. They are essentially functions that take in a message (known as action) and a state and return another message (known as a command) and the updated state. The application is never async, can not panic, doesn't read a globally mutable value, etc. Given the same input, it returns the same output (referentially transparent). All the side-effects are performed outside the application, using a custom runtime that you have to write. 

**Action:** An input message to the application. Representing typically a user-action.
**Command:** A message produced by the application to do something, without actually doing it. It's interpreted by the external runtime.
**State:** An owned state that allows maintaining contexts between each run of the application. For eg: If we build a counter application, the count would be stored within the state.

Because the application can only be interacted with via messages, the application too becomes stateless and very easy to test. Just send in data, and assert on the returned data. You don't need to mock anything. The runtime is responsible for running the application over and over, each time passing it a new message and the updated state. This process continues unless there is a failure or explicit exit command given by the application. 

The runtime connects with the underlying **drivers** that perform the actual side effects. Drivers are generally simple, for eg: making an http request or writing something to disk, etc. They represent the implementation of the commands.

image

ayewo:
**Summary:**  
_This PR adds integration tests for the CLI._

**Issue Reference(s):**  
Fixes #450 so I can /claim #450

**Build & Testing:**

**Checklist:**


ayewo:
@tusharmath would like to ask if my PR is going in the right direction?

tusharmath:
@ayewo Our goal is to have thourough

ayewo:
> @ayewo Our goal is to have thorough tests that cover all the commands and it's implementation. This is definitely going in that direction :)
> 
> PS: Clicked close by mistake 🙈

@tusharmath Haha, no worries. Second time this is happening in a 7-day span. GitHub needs to fix the placement of the commenting buttons.

Thanks for confirming. Will try to hit 100% coverage.

ayewo:
@tusharmath sorry I’ve not been able to complete it. I’m hoping to find time soon to update my PR.

ayewo:
@tusharmath 

The tests in the `main` branch is failing preventing the `codecov` report from being run/generated.

```sh
test valid::valid::tests::test_validate_or_second_fail ... ok

test result: ok. 111 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.08s

     Running unittests src/main.rs (target/debug/deps/tailcall-25a42f10150f7108)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/custom_resp_header_spec.rs (target/debug/deps/custom_resp_header_spec-70b4b4ab1a14cb6a)

running 1 test
test integration_tests::verify_response_headers ... FAILED

failures:

---- integration_tests::verify_response_headers stdout ----
thread 'integration_tests::verify_response_headers' panicked at 'Server failed to start: Error: Server Failed
       ❯ The port is already in use
Caused by:
  • error creating server listener: Address already in use (os error 48)', tests/custom_resp_header_spec.rs:11:8
thread 'integration_tests::verify_response_headers' panicked at 'called `Option::unwrap()` on a `None` value', tests/custom_resp_header_spec.rs:37:45


failures:
    integration_tests::verify_response_headers

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.31s

error: test failed, to rerun pass `--test custom_resp_header_spec`
```

ayewo:
@tusharmath It seems to happen due to some race condition in the tests. Some tests spawn a thread that listens on port 8000, effectively blocking other tests that do the same.

It seems that the spawned `tailcall` process doesn't always get killed after `cargo test` terminates, causing the issue to affect the next run of the tests.

ayewo:
@tusharmath I've tried as best as I can to get close to 100% coverage.
image

The reason why it's not 100% is because of this [PR ](https://github.com/tailcallhq/tailcall/pull/403) that was recently merged that added new code to the CLI module (which wasn't present when I started).

Please take a look when you have a moment.

tusharmath:
@ayewo I feel some of the tests are kind of testing the feature provide by the clap library. I don't think we need to be worried about those tests.

ayewo:
> @ayewo I feel some of the tests are kind of testing the feature provide by the clap library. I don't think we need to be worried about those tests.

@tusharmath What do you suggest? Combine those into a single test?

tests/cli.rs
```diff
@@ -0,0 +1,198 @@
...
+  cmd.arg("check").arg("examples/jsonplaceholder.graphql").arg("-s");
+  cmd
+    .assert()
+    .success()
+    .stdout(predicate::str::contains("No errors found"));
+
+  Ok(())
+}
+
+// Start command tests
```
meskill, line 109: we can wrap several related tests in `mod` and it will be easier to see the test run results and filter test run based on the command name

meskill:
> @tusharmath It seems to happen due to some race condition in the tests. Some tests spawn a thread that listens on port 8000, effectively blocking other tests that do the same.
> 
> It seems that the spawned `tailcall` process doesn't always get killed after `cargo test` terminates, causing the issue to affect the next run of the tests.

could it be due the fact that tests are run in parallel and we get the race condition in that case? 
or your sure that something isn't cleaned up after execution? maybe in that case you can see the living tailcall process in your system after the execution?

ayewo:
> could it be due the fact that tests are run in parallel and we get the race condition in that case?
or your sure that something isn't cleaned up after execution? maybe in that case you can see the living tailcall process in your system after the execution?

Yes, whenever it happened, I had to manually terminated the orphaned process with `killall tailcall`.

I did try with `--test-threads=1` and the issue remained. To fix the issue, I had to remove the test that I added that was spawning a thread that was leading to the race condition.

ayewo:
> > @ayewo I feel some of the tests are kind of testing the feature provide by the clap library. I don't think we need to be worried about those tests.
> 
> @tusharmath What do you suggest? Combine those into a single test?

@tusharmath would appreciate your reply here so I can wrap this up.

ayewo:
I'm confused. In the original [ticket](https://github.com/tailcallhq/tailcall/issues/450) you scoped the tests to just the CLI module:
> CLI piece has no unit tests. We need to re-structure the code to be able to write unit tests in the following manner:

> Technical Requirements:
>    Implementation should including sending a CLI Command and asserting on what's printed on the Console.
>   Avoid using mocks for testing, depend on pure functions as much as possible.
>   Achieve 100% coverage for the **CLI modules**.

But in your last reply you brought up `src/http/server.rs` which is in a different module with its own set of tests:
> I think the code-coverage is increasing but not enough. The expectation is close to 100% for CLI modules.
>  
> file 	missed lines 	head % 	change %
> **src/http/server.rs** 	**0** 	**68.85%** 	**1.64%**
> src/cli/command.rs 	0 	57.14% 	57.14%
> src/cli/tc.rs 	0 	41.10% 	41.10%
> 
> The core pieces where we often find bugs aren't getting covered and are also the tricky bits to test.

tusharmath:
Hey @ayewo I imagined server to be part of the CLI because it's interacted directly via the CLI. Do you think we can try to achieve ~100% for command.rs and cli.rs at least?

ayewo:
> Do you think we can try to achieve ~100% for command.rs and cli.rs at least?

@tusharmath I'll try.

tusharmath:
@ayewo Are you still working on this? I think the code coverage hasn't increased since your last commit message — 



src/cli/command.rs | 0 | 57.14% | 57.14%
-- | -- | -- | --
src/cli/tc.rs | 0 | 41.10% | 41.10%

[src/cli/command.rs](https://app.codecov.io/gh/tailcallhq/tailcall/pull/458/blob/src/cli/command.rs)
0
57.14%
57.14%
[src/cli/tc.rs](https://app.codecov.io/gh/tailcallhq/tailcall/pull/458/blob/src/cli/tc.rs)
0
41.10%
41.10%

ayewo:
@tusharmath Can I refactor `src/cli/tc.rs` to make it easier to test?

tusharmath:
Yes that's the best.

tests/cli.rs
```diff
@@ -0,0 +1,278 @@
...
+
+  #[test]
+  fn check_command_file_exists_and_valid() -> Result<(), Box<dyn std::error::Error>> {
+    let mut cmd = Command::cargo_bin("tailcall")?;
+
+    cmd
+      .arg("check")
+      .arg("examples/jsonplaceholder.graphql")
+      .arg("--n-plus-one-queries")
+      .arg("--schema");
```
meskill, line 79: should we also check that the cli outputs schema in that case? at least part of it

tests/cli.rs
```diff
@@ -0,0 +1,278 @@
...
+      .arg("start")
+      .arg("tests/graphql/errors/test-const-with-inline.graphql");
+    cmd
+      .assert()
+      .failure()
+      .stderr(predicate::str::contains("Error: Invalid Configuration"));
+
+    Ok(())
+  }
+}
```
meskill, line 210: can we add test for successful start command as well to see that command will start successfully?

tests/cli.rs
ayewo: I'll see if this is possible.

tests/cli.rs
ayewo: Okay.

ayewo:
@tusharmath please take a look when you have a moment.

1. I had to replace `inquire::Confirm` with `dialoguer::Confirm` because it didn't play nice with automated tests.
2. Then I added the `rexpect` crate to make it easier to test the CLI output. 
3. Some tests were failing under `cargo test`'s multi-threading so I made them run serially using the `serial_test` crate.

tusharmath:
@ayewo Closing this PR for now. I don't think this is the best way to write tests for a CLI application.

ayewo:
@tusharmath 🙏.

Still interested in what you would have liked to see.
(Would appreciate it if you could tag me if and when you revisit this in a fresh issue.)

tusharmath:
@ayewo I think I will create some facade with a few testcases like we did with http-spec. And then tag you to continue on it. Hope that works.

ayewo:
That works.

PR #474, opened 2023-10-12
fix: upstream setting keys should be optional

amitksingh1490:
**Summary:**  
upstream setting keys should be optional in order to have the capability to merge right else if right don't have upstream values it would pick default always

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -36,15 +36,15 @@ impl DefaultHttpClient {
     let upstream = &server.upstream;
 
     let mut builder = Client::builder()
-      .tcp_keepalive(Some(Duration::from_secs(upstream.tcp_keep_alive)))
-      .timeout(Duration::from_secs(upstream.timeout))
-      .connect_timeout(Duration::from_secs(upstream.connect_timeout))
-      .http2_keep_alive_interval(Some(Duration::from_secs(upstream.keep_alive_interval)))
-      .http2_keep_alive_timeout(Duration::from_secs(upstream.keep_alive_timeout))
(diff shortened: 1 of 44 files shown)


src/http/client.rs
```diff
@@ -36,15 +36,15 @@ impl DefaultHttpClient {
...
-      .user_agent(upstream.user_agent.clone());
+      .tcp_keepalive(Some(Duration::from_secs(upstream.tcp_keep_alive.unwrap_or(5))))
+      .timeout(Duration::from_secs(upstream.timeout.unwrap_or(60)))
+      .connect_timeout(Duration::from_secs(upstream.connect_timeout.unwrap_or(60)))
+      .http2_keep_alive_interval(Some(Duration::from_secs(upstream.keep_alive_interval.unwrap_or(60))))
+      .http2_keep_alive_timeout(Duration::from_secs(upstream.keep_alive_timeout.unwrap_or(60)))
+      .http2_keep_alive_while_idle(upstream.keep_alive_while_idle.unwrap_or(false))
+      .pool_idle_timeout(Some(Duration::from_secs(upstream.pool_idle_timeout.unwrap_or(60))))
+      .pool_max_idle_per_host(upstream.pool_max_idle_per_host.unwrap_or(200))
+      .user_agent(upstream.user_agent.clone().unwrap_or("Tailcall/1.0".to_string()));
```
tusharmath, line 47: Don't put hard coded values here in the client.
Create `methods` for each field type to get the default values from the Server.

tusharmath (approved):

tusharmath (approved):

Issue #470, opened 2023-10-12
Publish tailcall on npm


tusharmath:
Nothing much to describe here except that we need to publish tailcall on `tailcallhq/tailcall` :) 

**Technical Requirements:**
- CI should be integrated and for every release on github, the package should be published on npm.
- Use `NPM_TOKEN` secret to publish on npm.

**Additional Information:**
~~- Use this as reference: https://github.com/orhun/git-cliff/tree/main/npm~~
- Use this as reference - https://github.com/tusharmath/rust-npm-demo/edit/main/README.md

**IMPORTANT**
PLEASE DON'T ATTEMPT TO CALL A RUST PROGRAM VIA NODE.JS.

tusharmath:
fixed in #562

PR #483, opened 2023-10-13
add npm publish

neo773:
This PR adds a npm package folder along with necessary files and GitHub workflow to auto publishh.

/claim #470


.github/workflows/ci.yml
```diff
@@ -233,3 +233,16 @@ jobs:
...
+        with:
+         node-version: "20.x"
+         registry-url: "https://registry.npmjs.org"
+      - run: cd npm && mkdir target
+      - run: cp -r target/${{ matrix.target }}/release/tailcall-${{ matrix.target }} npm/target
+      - name: Update NPM version
+        run: cd npm && node publish.js
+        env:
+          APP_VERSION: ${{ needs.draft_release.outputs.create_release_name }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
tusharmath, line 248: Please run `./lint.sh`

neo773:
@tusharmath 
This implementation also resolves all platform binaries, we'll be shipping over 50MB of binaries
https://github.com/orhun/git-cliff/blob/bcfcd1fd59fa2c1ef3222d588f3a563c3e10027e/npm/git-cliff/package.json#L64C1-L71C5

I was thinking about downloading the binaries on the fly for the initial run and caching for subsequent runs, we'll pull them via GitHub releases API.
The version would still be locked to package.json so it won't break semver, this would simplify the whole pipeline.

```
node_moodules/tailcall/
├── package.json
├── src
│   └── index.js
└── target
    └── ${os}-${arch} -- binary would be pulled in this directory
```

What do you think about approach?

tusharmath:
@neo773 I think spawning a node.js thread to run a Rust application doesn't seem like a good approach. It's going to slow down the startup time.

neo773:
@tusharmath We could run a `postinstall` script to preload the binary, this should eliminate the initial startup time.

neo773:
@tusharmath 

Added my changes. binary is auto installed via `postinstall` script so it doesn't affect startup time.
Overall it's very compact

https://github.com/tailcallhq/tailcall/assets/62795688/bfddfee5-ec1f-4113-8fa9-100610ca93d5

neo773:
@tusharmath 
Hi there, were you able to look at my last comment?
https://github.com/tailcallhq/tailcall/pull/483#issuecomment-1762059532

npm/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "name": "@tailcallhq/tailcall",
+  "version": "0.13.0",
+  "type": "module",
+  "files": [
+    "src/*",
+    "target/*"
+  ],
+  "description": "A high-performance GraphQL proxy, optimized for cloud-native environments, covering edge, middle, and service layers.",
+  "main": "./src/index.js",
+  "bin": "./src/index.js",
```
tusharmath, line 11: Are these required?

neo773: Yeah, won't work without it, npm packages need an entry point.

tusharmath:
Can you collaborate on #488. I think the approach there seems more inline with https://blog.orhun.dev/packaging-rust-for-npm/

neo773:
So I initially read that post, when I started on the bounty but it added unnecessary complexity/abstraction in my opinion
Then I remembered about [Prisma's](https://github.com/prisma/prisma) approach which is also written in Rust and it seemed more cleaner.

You can look the final diff, it only adds a single npm.yml file and a single npm package.

tusharmath:
@neo773 I think the following minimal `package.json` should work.

```json
{
  "name": "@tailcallhq/tailcall",
  "version": "0.1.0",
  "description": "A high-performance GraphQL proxy, optimized for cloud-native environments, covering edge, middle, and service layers.",
  "bin": "./tailcall"
}
```

We now just need to make sure the correct version of binary is installed on the platform.

You approach, starts a node process first and then run the Rust application. That's going to be a non-optimal user experience.

npm/package.json
meskill: it's possible to specify shell script or shell command here instead of js file that won't spin up the nodejs runtime to run the binary

neo773:
@tusharmath 
Added the changes, removed the Node.js overhead, check the demo below
Tested installing as global.

https://github.com/tailcallhq/tailcall/assets/62795688/5c34d74c-88d1-4fff-bef5-94ecf9e3efa1

npm/src/index.js
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env node
+// @ts-check
+
+import { spawnSync } from "child_process"
+import { fileURLToPath } from "url"
+import { dirname } from "path"
+import EasyDl from "easydl"
```
tusharmath, line 7: Instead of downloading the dependencies, we should use the binaries that are produced locally.

neo773: So we'll compile the binary on user's machine?

neo773: If you're talking about node dependencies all of them are included in node except `easydl` which is a zero dependency library.

tusharmath: Yeah, we should compile them on the user's machine. In fact we have a job that does this already and uploads to github releases. We can use those builds directly. That way `easydl` might not be required.

tusharmath:
Please add `js` to the prettier command in `lint.sh`

npm/src/index.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+async function run() {
+  await preload()
+  const args = process.argv.slice(2)
+  const processResult = spawnSync(fullPath, args, {stdio: "inherit"})
+  process.exit(processResult.status ?? 0)
+}
+
+process.argv.includes("--preload") ? preload() : run()
```
tusharmath, line 67: I think this code can be simplified. We don't need a preload or run. We just need to perform some file copy operations and generation of package.json and readme. 

**NOTE:** these operations should happen in parallel.

npm/src/index.js
neo773: Sorry a little confused here 
> We can use those builds directly

Isn't that what the EasyDL code does pull from GitHub releases API

meskill:
Thanks for your effort.

But while the idea to use such approach might look simpler it has significant drawbacks:
- postinstall scripts have serious security concerns. See [yarn docs](https://yarnpkg.com/advanced/lifecycle-scripts#postinstall). In short postinstall scripts could be disabled, they could be restriction on using deps with them, they hiding activity from users, they affect performance
- putting files in user directories raises questions about versioning the files and deleting files that are not used. Basically, we are losing any advantages of using npm as distribution system for packages with project specific scope. And it could be easily to just provide install script separately for doing this.

So, the https://github.com/tailcallhq/tailcall/pull/488 has better chances to get merged as it doesn't have those drawbacks and just little more complex

PR #488, opened 2023-10-13
feat: add support for npm package

### LINKED ISSUE #416: Pass `operations` to `check` command
**Problem**
As a user I wish to validate the generated graphql schema with a set of graphQL queries.

**Solution 1**
Since #438 we now support loading multiple graphQL files. Which means that we can pass the queries as a part of the configurations more easily. We should just validate the queries that are in the graphQL files without and special check parameter.

The command would look a lot simpler now:
```bash
tc check ./jsonplaceholder.graphql query1.graphql query2.graphql
```
All the above options should be allowed to test the graphQL schema against the provided queries and mutations. 

**Solution 2**
We will extend the check command on the CLI to support checking of the schema against provided queries and mutations:

```bash
# using operation parameter
tc check ./jsonplaceholder.graphql --operation query1.graphql --operation mutation1.graphql

# using comma as a delimiter
tc check ./jsonplaceholder.graphql --operation file1.graphql,file2.graphql

# using space as a delimeter
tc check ./jsonplaceholder.graphql --operation file1.graphql file2.graphql
```
All the above options should be allowed to test the graphQL schema against the provided queries and mutations.

**Technical Requirements**
- Use `ValidationError` to capture and display all the errors.
- Errors should be composed with errors `from_config`.

### LINKED ISSUE #470: Publish tailcall on npm

b4s36t4:
**Summary:**  
This PR adds support for tailcall npm package to be publishable via CI pipelines.

**Issue Reference(s):**  
Fixes #470  

/claim #470 

**Build & Testing:**

**Checklist:**

<details><summary>Demo</summary>

https://github.com/tailcallhq/tailcall/assets/59088937/d7c888d6-43dd-4c4b-bc05-2c7c2a254abc

</details>


b4s36t4:
Heavily inspired from swc and biomejs implementations.

This version only installs the binary depends on the arch of user's system @tusharmath

tusharmath:
Saw the video ❤️ 
Will we be able to use it by installing globally?
```
npm i -g @tailcallhq/tailcall
```

tusharmath:
Can you run `./lint.sh`

npm/tailcall/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "@tailcallhq/tailcall",
+  "version": "0.13.0",
+  "type": "module",
+  "homepage": "https://tailcall.run",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/tailcallhq/tailcall.git"
+  },
+  "author": "TailcallHQ",
+  "license": "MIT",
```
tusharmath, line 11: Apache 2.0

npm/tailcall/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "bugs": "https://github.com/tailcallhq/tailcall/issues",
+  "description": "A high-performance GraphQL proxy, optimized for cloud-native environments, covering edge, middle, and service layers.",
+  "files": [
+    "scripts/postinstall.js"
```
tusharmath, line 15: do we need this?

.github/workflows/ci.yml
```diff
@@ -227,9 +227,32 @@ jobs:
...
+
+      - name: Generate npm packages
+        run: node npm/tailcall/scripts/generate-packages.js
+        env: APP_VERSION=${{ needs.draft_release.outputs.create_release_name }}
+
+      - name: Publish npm packages as latest
+        # CD to package and publish and move back to top-level folder
+        run: for package in npm/*; do cd $package; npm publish --tag latest --access public --provenance; cd ../..; done
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
tusharmath, line 258: - Can we not publish everything together? What happens if one of the publish command fails?
- Can we log after every publish?

.github/workflows/ci.yml
```diff
@@ -227,9 +227,32 @@ jobs:
...
           overwrite: true
+
+      - name: Install Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          registry-url: "https://registry.npmjs.org"
+
+      - name: Generate npm packages
+        run: node npm/tailcall/scripts/generate-packages.js
```
tusharmath, line 251: Let's use `npm run gen` command instead over here.

tusharmath:
We can remove the generated files from being committed to git?

b4s36t4:
Yes, we can able to run it globally too. 

I will update the PR with the comments you provided. Thanks.

.github/workflows/ci.yml
b4s36t4: If one of them fails other one continues. 

Logging in the sense? Just printing that one of the package is published right? 

Also by default npm would console everything related to push and  the reason for push fails etc.

b4s36t4:
I can't able to lint.sh can you provide a doc on how to run it? Some errors it seems.

.github/workflows/ci.yml
b4s36t4: Can you check now?

.github/workflows/ci.yml
tusharmath: Good point. npm logs should be enough.

tusharmath:
There are a few other comments @b4s36t4 
For the lint, you can just run prettier.  The build is passing, so it might not be required.

b4s36t4:
I think due to the difference in remote config and other issues. 
Also I did a force push which caused some of issues stale. 

If you could can you please review it again? 

> except removing postinstall.js file exporting in package.json.

Rest looks good to me based on previous review.

tusharmath:
Yikes!
Commenting below so that it doesn't get lost — 

- [ ] Remove generated files from Git.

npm/tailcall/scripts/generate-packages.js
```diff
@@ -0,0 +1,66 @@
...
+    rootManifest["optionalDependencies"] = optionalPackages
+
+    fs.writeFileSync(MANIFEST_PATH, JSON.stringify(rootManifest, null, 2))
+
+    console.log("[Publish] Updated root package.json")
+  } catch (error) {
+    console.log("[Publish] [Error] with reading Target files")
+    process.exit(1)
+  }
+}
```
tusharmath, line 64: Can perform these operations in parallel using the async API. Small things affect build performance :)

.github/workflows/ci.yml
meskill: but ignoring error in that case could mask the problem as the ci will pass successfully and to know about problem we need to thoroughly read through logs or inspect npm packages

npm/tailcall/scripts/generate-packages.js
```diff
@@ -0,0 +1,66 @@
...
+
+function generatePackages() {
+  const __filename = fileURLToPath(import.meta.url)
+  const __dirname = dirname(__filename)
+
+  let target
+
+  try {
+    target = resolve(__dirname, "../../../dist")
+  } catch {
```
meskill, line 19: if I'm not mistaken the `path.resolve` doesn't check the built path for existence and it just concatenates string in right way. So I think catch error won't happen or at least will have misleading error message

npm/tailcall/scripts/generate-packages.js
```diff
@@ -0,0 +1,66 @@
+import * as fs from "node:fs"
+import {dirname, resolve} from "node:path"
+import {fileURLToPath} from "node:url"
+
+const APP_VERSION = process.env.APP_VERSION
+
+const MANIFEST_PATH = resolve(fileURLToPath(import.meta.url), "../..", "package.json")
```
meskill, line 7:
```suggestion
const MANIFEST_PATH = resolve(fileURLToPath(import.meta.url), "../package.json")
```

shouldn't it be like this? as package.json is jus one level above

npm/tailcall/package.json
meskill: has it been resolved? I still see that entry but no such file in the pr

.github/workflows/ci.yml
```diff
@@ -227,9 +227,32 @@ jobs:
       - name: Rename Binary with Target Name
         run: mv target/${{ matrix.target }}/release/tailcall target/${{ matrix.target }}/release/tailcall-${{ matrix.target }}
 
+      # Copy the CLI binary and rename it to include the name of the target platform
+      - name: Copy CLI binary
+        id: copy-cli
+        run: |
+          mkdir dist
+          cp target/${{ matrix.target }}/release/tailcall ./dist/tailcall-${{ matrix.target }}
```
meskill, line 235: wouldn't it be easier to just copy to the destination folder inside `npm/*` and do not do copying thing in the `generate-packages` script again?

npm/tailcall/scripts/generate-packages.js
```diff
@@ -0,0 +1,66 @@
...
+      console.log(`[Publish] Copy ${_file} to ${packageRoot} `)
+      fs.copyFileSync(`${target}/${_file}`, `${packageRoot}/tailcall`)
+
+      const packageFileName = `${packageRoot}/package.json`
+
+      const packagefile = fs.readFileSync(packageFileName, {encoding: "utf-8"})
+
+      const packageJSON = JSON.parse(packagefile)
+
+      packageJSON["version"] = APP_VERSION || "latest"
```
meskill, line 43: `APP_VERSION || "latest"` repeats several time and I think could be replaced by settings initially when defining the variable.

Or might be better to check that `env.APP_VERSION` is defined and not to publish with any defaults as it could be marker of some issues

npm/tailcall/scripts/generate-packages.js
```diff
@@ -0,0 +1,66 @@
...
+
+    rootManifest["version"] = APP_VERSION || "latest"
+
+    rootManifest["optionalDependencies"] = optionalPackages
+
+    fs.writeFileSync(MANIFEST_PATH, JSON.stringify(rootManifest, null, 2))
+
+    console.log("[Publish] Updated root package.json")
+  } catch (error) {
+    console.log("[Publish] [Error] with reading Target files")
```
meskill, line 61: basically it could be not only errors about reading but anything else. And in this case it would be better to log the `error` as well as it'll help with investigating of issues

.github/workflows/ci.yml
```diff
@@ -227,9 +227,32 @@ jobs:
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          registry-url: "https://registry.npmjs.org"
+
+      - name: Generate npm packages
+        run: cd npm/tailcall && npm run gen
+        env: APP_VERSION=${{ needs.draft_release.outputs.create_release_name }}
+
+      - name: Publish npm packages as latest
```
meskill, line 254: the updated files won't be committed to the repo and the repo content will be different from the published one on the npm.

I don't know if this is much a problem

.github/workflows/ci.yml
```diff
@@ -161,31 +161,31 @@ jobs:
...
             rust: stable
-            target: aarch64-unknown-linux-musl
+            target: aarch64-linux-musl
             cross: true
 
           - build: linux-i686-gnu
             os: ubuntu-latest
             rust: stable
-            target: i686-unknown-linux-gnu
+            target: i686-linux-gnu
```
meskill, line 188: windows is missing am I right?

b4s36t4: I think we don't have builds for windows in our CI. Do you want me to add that as well?

If so I would request for a new PR for that so that it won't collide with the feature and so.

npm/tailcall/scripts/generate-packages.js
b4s36t4: Yes, we can remove the `try/catch`.

npm/tailcall/scripts/generate-packages.js
b4s36t4: Yes, fixed.

.github/workflows/ci.yml
b4s36t4: No it won't be a problem.

b4s36t4:
@meskill Thanks for the review, it does make a lot of sense to make your suggested changes. It did came up nice and easy.

Really appreciate it :)

b4s36t4:
@tusharmath Please review the PR now. Let me know if there's anything missing thanks :).

tusharmath:
@b4s36t4 https://github.com/tailcallhq/tailcall/pull/488#issuecomment-1762814132

b4s36t4:
I don't think there are any generated file. All the package.json files are intended.

https://github.com/swc-project/swc/tree/main/scripts/npm/ see this for the reference.

tusharmath:
@b4s36t4 Though SWC is committing the files to git, I don't think it's a good idea for our current setup. These files can be generated, and contain the release version in them. We will end up creating a commit everytime we make a release, which is unnecessary.

b4s36t4:
Will update.

.github/workflows/ci.yml
meskill: we can also add an dry-run npm release workflow in order to able to test that everything works as expected especially before merging this

b4s36t4:
@tusharmath Sorry for asking questions such lately.

But As you mentioned we need to wipe/no-commit package.json files committed to github, I want to bring this attention if we delete the package.json and do this file generation with the script wouldn't that be complex?

If you see we're building the tailcall for different platforms for different arch types handling these conditions inside the script would make the script a bit complex, agree?

As per the comment to commit everytime the package.json we don't have to. We can keep them as a read-only files (need to remove `version` key from `package.json`) except I don't see any problem with committing these files.

Let me know your thoughts.

tusharmath:
> But As you mentioned we need to wipe/no-commit package.json files committed to github, I want to bring this attention if we delete the package.json and do this file generation with the script wouldn't that be complex?

It would be easier to manage the platforms this way. We could just configure the script to generate the package.json file for that particular platform. This is quite standard in node.js projects actually.

>If you see we're building the tailcall for different platforms for different arch types handling these conditions inside the script would make the script a bit complex, agree?

I am not sure why should the build script change. What I am saying is - we should write a node.js script, that generates the dirs, readme, package.json, and copy appropriate release builds for the platforms we wish to support. Those package.json should be created from a template.package.json.

> As per the comment to commit everytime the package.json we don't have to. We can keep them as a read-only files (need to remove version key from package.json) except I don't see any problem with committing these files.

You are right we don't "need" to commit. However as discussed, generating the files using a template would make it easier for us to change the meta data about the package easily.

.github/workflows/ci.yml
tusharmath: We should use `npm publish --workspaces` command instead of looping using bash.

https://docs.npmjs.com/cli/v7/commands/npm-publish#workspaces

npm/tailcall/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "description": "A high-performance GraphQL proxy, optimized for cloud-native environments, covering edge, middle, and service layers.",
+  "keywords": [
+    "GraphQL",
+    "Rust",
+    "microservices"
+  ],
+  "engines": {
+    "node": ">=14.*"
+  }
+}
```
tusharmath, line 25: This should be a private package. 
set `private:true`.

Add workspaces for each supported platform.

b4s36t4:
Working on few other things, probably would be completing things by tomorrow.

PR #489, opened 2023-10-13
feat/416/pass operations to check command

ologbonowiwi:
- feat: add operation arg
- feat: enable space delimiter on operation
- feat: enable comma delimiter on operation

**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

- [ ] I have run `./lint.sh` to address and fix linting issues.

**Checklist:**

### DIFF
--- a/src/cli/tc.rs
+++ b/src/cli/tc.rs
@@ -28,7 +28,14 @@ pub async fn run() -> Result<()> {
       start_server(config).await?;
       Ok(())
     }
-    Command::Check { file_path, n_plus_one_queries, schema } => {
+    Command::Check { file_path, n_plus_one_queries, schema, operation } => {
+      let operation = operation
+        .iter()
+        .flat_map(|operation| operation.split(',').map(|operation| operation.to_string()))
+        .collect::<Vec<String>>();
+
+      println!("Operation: {:?}", operation);
+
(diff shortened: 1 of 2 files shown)


src/cli/tc.rs
```diff
@@ -28,7 +28,14 @@ pub async fn run() -> Result<()> {
...
       Ok(())
     }
-    Command::Check { file_path, n_plus_one_queries, schema } => {
+    Command::Check { file_path, n_plus_one_queries, schema, operation } => {
+      let operation = operation
+        .iter()
+        .flat_map(|operation| operation.split(',').map(|operation| operation.to_string()))
+        .collect::<Vec<String>>();
+
+      println!("Operation: {:?}", operation);
```
ologbonowiwi, line 37: @tusharmath, how can I use the queries now?

also, can you provide some queries that I can run against jsonplaceholder? Thanks!

tusharmath: - You will need to parse the graphQL file to identify the queries.
- Sure: Something like should be a valid graphQL query `query {posts {user {name}}}`

tusharmath:
@ologbonowiwi I updated the issue. I think we should check for all queries that are part of the document even if there is no CLI flag provided.

PR #499, opened 2023-10-15
feat: Load Config via HTTP

### LINKED ISSUE #441: Load Config via HTTP
Currently Tailcall server can load on configurations that are on the file system. We would like to load these configurations over HTTP. This will allow the server to update automatically without performing restarts on production.

**Technical Requirements**
- `tc start` should also take http endpoints as inputs.
- Update cli documentation to depict that it supports both file and http paths.
- Incase of load failure error should be shown on the console
  **Refresh Mechanism:** Implement a refresh mechanism to check for updates to the configuration file at a defined interval.
    - The refresh interval should be configurable, with a reasonable default value (e.g., 5 minutes).
    - An additional feature could be to implement a smart refresh mechanism which would use HTTP headers like ETag or Last-Modified to determine if the configuration has changed and only download the new configuration if there's a change. This can be done via http_client as it supports http caching as per RFC.

**Additional Note**
- Http url hosting the configuration could expose data in any of the formats - YML, JSON, GraphQL, and we should support all. refer to #439 for more details.

PS: Make sure you update documentation on https://tailcall.run/docs 

**Design**
- Write config reading logic to a `config::read::ConfigReader` module.
- Module should be responsible for reading the file from http and disk.
- Module should have support to reload config via poll
- Should expose an API that gives access to current configuration without any async
- Add unit tests.

neo773:
This PR adds ability to load config over HTTP with polling.

https://github.com/tailcallhq/tailcall/assets/62795688/ae30250b-9958-4588-b314-479d9d044d8c

**Issue Reference(s):**  
Fixes #441

**Build & Testing:**

/claim #441

### DIFF
--- a/src/cli/command.rs
+++ b/src/cli/command.rs
@@ -27,6 +27,8 @@ pub enum Command {
     file_path: Vec<String>,
     #[arg(long)]
     log_level: Option<log::Level>,
+    #[arg(long)]
+    refresh_interval: Option<u64>,
   },
 
   /// Validate a composition spec
--- a/src/cli/command.rs
+++ b/src/cli/command.rs
@@ -28,7 +28,7 @@ pub enum Command {
     #[arg(long)]
     log_level: Option<log::Level>,
     #[arg(long)]
-    refresh_interval: Option<u64>,
+    poll_interval: Option<u64>,
   },
 
   /// Validate a composition spec
--- a/src/config/config_poll.rs
+++ b/src/config/config_poll.rs
@@ -1,9 +1,7 @@
-use std::collections::HashMap;
-use std::sync::{Arc, RwLock};
+use std::sync::atomic::AtomicPtr;
+use std::sync::Arc;
 use std::time::Duration;
 
-use hyper::header::HeaderValue;
-use reqwest::header::IF_NONE_MATCH;
 use reqwest::Client;
 use tokio::time;
 
@@ -15,11 +13,11 @@ use crate::http::ServerContext;
 pub struct ConfigLoader {
   file_path: String,
   refresh_interval: u64,
-  state: Arc<RwLock<ServerContext>>,
+  state: Arc<AtomicPtr<ServerContext>>,
 }
 
 impl ConfigLoader {
-  pub fn new(file_path: String, refresh_interval: u64, state: Arc<RwLock<ServerContext>>) -> Result<Self, CLIError> {
+  pub fn new(file_path: String, refresh_interval: u64, state: Arc<AtomicPtr<ServerContext>>) -> Result<Self, CLIError> {
     Ok(Self { file_path, refresh_interval, state })
   }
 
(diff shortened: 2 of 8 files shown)


src/cli/command.rs
```diff
@@ -27,6 +27,8 @@ pub enum Command {
     file_path: Vec<String>,
     #[arg(long)]
     log_level: Option<log::Level>,
+    #[arg(long)]
+    refresh_interval: Option<u64>,
```
tusharmath, line 31: update `.tailcallrc.graphql`

tusharmath, line 31:
```suggestion
    poll_interval: Option<u64>,
```

src/config/config_poll.rs
```diff
@@ -0,0 +1,98 @@
...
+                continue;
+              }
+            }
+          }
+          Err(_) => continue,
+        };
+      }
+    });
+  }
+}
```
tusharmath, line 98: Add unit tests.

src/config/config_poll.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use crate::blueprint::Blueprint;
+use crate::cli::CLIError;
+use crate::config::Config;
+use crate::http::ServerContext;
+
+pub struct ConfigLoader {
+  file_path: String,
+  refresh_interval: u64,
+  state: Arc<RwLock<ServerContext>>,
```
tusharmath, line 18: I think we should use a non-blocking implementation instead of RWLock. RWLock can drastically affect the server's performance, even when this feature is disabled.

tusharmath: Moreover we should have two separate implementations, one that just load's once and doesn't ever update. This doesn't require any locking while reads. The other should be implemented using an `AtomicPtr` for better performance.

src/config/config_poll.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+impl ConfigLoader {
+  pub fn new(file_path: String, refresh_interval: u64, state: Arc<RwLock<ServerContext>>) -> Result<Self, CLIError> {
+    Ok(Self { file_path, refresh_interval, state })
+  }
+
+  pub async fn start_polling(&self) {
+    let client = Client::new();
```
tusharmath, line 27: We should use the existing HttpClient. It has all the cache RFC implemented.

tusharmath:
I think design wise we need a few changes:

```rust

pub struct ServerContext {
  pub schema: SchemaLoader, // changed
  pub http_client: DefaultHttpClient,
  pub blueprint: Blueprint,
}


// Add schema_loader.rs
enum SchemaLoader {
  Static(dynamic::Schema),
  Dynamic(ConfigLoader)
}

impl SchemaLoader {
  fn schema(self) -> dynamic::Schema {
    // pattern match and implement
  }
}
```

We should add unit tests only for `ConfigLoader` separately.

neo773:
@tusharmath 
I looked at it for a while and I'm sort of lost here to how it fits with `config_poll.rs`
Sorry if this is stupid, I'm still quite new to Rust coming from a TypeScript background.

> I think design wise we need a few changes:
> 
> ```rust
> pub struct ServerContext {
>   pub schema: SchemaLoader, // changed
>   pub http_client: DefaultHttpClient,
>   pub blueprint: Blueprint,
> }
> 
> 
> // Add schema_loader.rs
> enum SchemaLoader {
>   Static(dynamic::Schema),
>   Dynamic(ConfigLoader)
> }
> 
> impl SchemaLoader {
>   fn schema(self) -> dynamic::Schema {
>     // pattern match and implement
>   }
> }
> ```
> 
> We should add unit tests only for `ConfigLoader` separately.

ssddOnTop:
What happened to the PR? 
I am curious to know more on the progress and if I can contribute in completion

PR #508, opened 2023-10-17
fix: install.sh

neo773:
This PR fixes installing binary when version isn't specified
Current implementation was broken.


tusharmath:
thanks @neo773 !

Issue #519, opened 2023-10-20
Enable GraphQL BatchRequest support in Tailcall


sujeetsr:
Async graphql already supports graphql batch requests:
https://docs.rs/async-graphql/latest/async_graphql/enum.BatchRequest.html
https://www.apollographql.com/blog/apollo-client/performance/batching-client-graphql-queries/

When tailcall receives a graphql batch request with data as follows:
```json
[{"query":"query { post(id: 1) { title } }"}, {"query":"query { post(id: 1) { title } }"}]
```
it should respond with an array of responses for each query as shown below:
```json
[{"data":{"post":{"title":"sunt aut facere repellat provident occaecati excepturi optio reprehenderit"}}},{"data":{"post":{"title":"qui est esse"}}}]
```

PR #520, opened 2023-10-20
Support request batching on GraphQL Server

sujeetsr:
**Summary:**  
Enable GraphQL BatchRequest support in tailcall as described in [519](https://github.com/tailcallhq/tailcall/issues/519)
https://docs.rs/async-graphql/latest/async_graphql/enum.BatchRequest.html
https://www.apollographql.com/blog/apollo-client/performance/batching-client-graphql-queries/

**Issue Reference(s):**  
Fixes https://github.com/tailcallhq/tailcall/issues/519

**Build & Testing:**

**Checklist:**


tests/batched_request_spec.rs
```diff
@@ -0,0 +1,55 @@
...
+
+  #[tokio::test]
+  async fn test_batched_request() {
+    let schema_path = "tests/graphql_mock/test-batched-request.graphql";
+    tokio::spawn(initiate_test_server(schema_path.into()));
+    tokio::time::sleep(tokio::time::Duration::from_millis(300)).await;
+
+    let http_client = reqwest::Client::new();
+    let query_1_data = "{\"query\":\"query { post(id: 1) { title } }\"}";
+    let query_2_data = "{\"query\":\"query { post(id: 2) { title } }\"}";
```
tusharmath, line 37: nitpic!
```suggestion
    let query_1_data = json!({query: "query {post(id: 1) {title}}"});
    let query_1_data = json!({query: "query {post(id: 2) {title}}"});
```

tests/graphql_mock/test-batched-request.graphql
```diff
@@ -0,0 +1,14 @@
...
+}
+
+type Query {
+  post(id: Int): Post @http(path: "/posts/{{args.id}}")
+}
+
+type Post {
+  id: Int!
+  title: String!
+}
```
tusharmath, line 14: Question: Can we reuse some existing graphQL config?

src/http/server.rs
```diff
@@ -24,9 +24,18 @@ async fn graphql_request(req: Request<Body>, server_ctx: &ServerContext) -> Resu
...
+  let mut response = match server_ctx.blueprint.server.enable_batch_requests {
+    true => {
+      let request: async_graphql_hyper::GraphQLBatchRequest = serde_json::from_slice(&bytes)?;
+      request.data(req_ctx.clone()).execute(&server_ctx.schema).await
+    }
+    false => {
+      let request: async_graphql_hyper::GraphQLRequest = serde_json::from_slice(&bytes)?;
+      request.data(req_ctx.clone()).execute(&server_ctx.schema).await
+    }
+  };
```
tusharmath, line 38: This check doesn't need to happen for each request. We can create two special handlers — One for a GraphQLRequest and other to handle both GraphQLRequest and GraphQLBatchRequest. The first one will be the default one to use, focussed on performance and the second one will be slightly slower but handle request batching.

src/async_graphql_hyper.rs
```diff
@@ -16,6 +16,14 @@ impl GraphQLBatchRequest {
   {
     GraphQLResponse(executor.execute_batch(self.0).await)
   }
+
+  #[must_use]
```
tusharmath, line 20:
```suggestion
  #[must_use]
```
Add a comment to explain why `must_use` is required.

tusharmath:
Publish benchmark numbers. Compare with `main`

sujeetsr:
Benchmark with `ci-benchmark/benchmark.graphql` and `wrk -d 30 -t 4 -c 100 -s ci-benchmark/wrk.lua http://localhost:8000/graphql`

From main:
Screenshot 2023-11-08 at 10 09 58 AM

Form this PR branch:
Screenshot 2023-11-08 at 10 41 58 AM

The PR branch is about 500 req/sec slower

src/http/server.rs
```diff
@@ -21,15 +21,18 @@ fn graphiql() -> Result<Response<Body>> {
   )))
 }
 
-async fn graphql_request(req: Request<Body>, server_ctx: &ServerContext) -> Result<Response<Body>> {
+async fn graphql_request(
+  req: Request<Body>,
+  server_ctx: &ServerContext,
+  executor: Arc<dyn RequestExecutor + Send + Sync>,
```
tusharmath, line 27:
```suggestion
  executor: Arc<impl RequestExecutor + Send + Sync>,
```

This should improve performance.

src/async_graphql_hyper.rs
```diff
@@ -17,6 +17,13 @@ impl GraphQLBatchRequest {
   {
     GraphQLResponse(executor.execute_batch(self.0).await)
   }
+
+  pub fn data<D: Any + Clone + Send + Sync>(mut self, data: D) -> Self {
```
tusharmath, line 21: Are we using this function anywhere?

src/http/server.rs
```diff
@@ -81,3 +96,42 @@ pub async fn start_server(config: Config) -> Result<()> {
...
+  async fn execute(
+    &self,
+    bytes: &hyper::body::Bytes,
+    req_ctx: Arc<RequestContext>,
+    server_ctx: &ServerContext,
+  ) -> Result<GraphQLResponse> {
+    let request: async_graphql_hyper::GraphQLBatchRequest = serde_json::from_slice(bytes)?;
+    Ok(request.data(req_ctx.clone()).execute(&server_ctx.schema).await)
+  }
+}
```
tusharmath, line 137: nitpick
Consider moving request_handle to it's own file. The logic has become quite complicated now.

src/async_graphql_hyper.rs
sujeetsr: Yes, in `graphql_batch_request` in  `server.rs`

src/http/server.rs
```diff
@@ -63,20 +102,47 @@ fn create_allowed_headers(headers: &HeaderMap, allowed: &BTreeSet<String>) -> He
...
+    }
+    false => {
+      let server = hyper::Server::try_bind(&addr)
+        .map_err(CLIError::from)?
+        .serve(make_svc_single_req);
+      log_startup_messages(&addr, blueprint.server.enable_graphiql);
+
+      Ok(server.await.map_err(CLIError::from)?)
+    }
+  }
```
meskill, line 147:
```suggestion

      let server = hyper::Server::try_bind(&addr)
        .map_err(CLIError::from)?
        .serve(if blueprint.server.enable_batch_requests { make_svc_batch_req } else { make_svc_single_req });
      log_startup_messages(&addr, blueprint.server.enable_graphiql);

      Ok(server.await.map_err(CLIError::from)?)
```

wdyt?

src/http/server.rs
```diff
@@ -21,38 +22,76 @@ fn graphiql() -> Result<Response<Body>> {
...
+pub async fn graphql_batch_request(req: Request<Body>, server_ctx: &ServerContext) -> Result<Response<Body>> {
+  let req_ctx = Arc::new(create_request_context(&req, server_ctx));
+  let bytes = hyper::body::to_bytes(req.into_body()).await?;
+  let request: async_graphql_hyper::GraphQLBatchRequest = serde_json::from_slice(&bytes)?;
+  let mut response = request.data(req_ctx.clone()).execute(&server_ctx.schema).await;
+  response = update_cache_control_header(response, server_ctx, req_ctx);
+  let mut resp = response.to_response()?;
+  update_response_headers(&mut resp, server_ctx);
+  Ok(resp)
 }
```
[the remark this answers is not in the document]
sujeetsr: There are two considerations:
1. the decision of which function to use should be made after the config file is read, and not per request
2. We're trying to avoid dynamic dispatch (in my earlier code I had a `dyn` `Fn` which was doing what you suggested)
The difficulty I faced in making it one function is that if the appropriate struct or function needs to be passed in (based on the enable_batch_requests flag, then Rust insists that it should be a `dyn`

src/http/server.rs
sujeetsr: This is the error I get 
```
`if` and `else` have incompatible types
expected struct `hyper::service::make::MakeServiceFn<[closure@src/http/server.rs:125:44: 125:56]>`
   found struct `hyper::service::make::MakeServiceFn<[closure@src/http/server.rs:120:45: 120:57]>`
no two closures, even if identical, have the same type
consider boxing your closure and/or using it as a trait object
```
I think if I try to Box them or make it a trait obj, it will insist on it being a `dyn`

src/http/server.rs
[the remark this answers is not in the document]
meskill: I see, yeah sometimes it could be quite tricky.

I'm thinking about this way:
```rust
pub async fn graphql_request<T: DeserializeOwned + Data>(req: Request<Body>, server_ctx: &ServerContext) -> Result<Response<Body>> {
  let req_ctx = Arc::new(create_request_context(&req, server_ctx));
  let bytes = hyper::body::to_bytes(req.into_body()).await?;
  let request: T = serde_json::from_slice(&bytes)?;
  let mut response = request.data(req_ctx.clone()).execute(&server_ctx.schema).await;
  response = update_cache_control_header(response, server_ctx, req_ctx);
  let mut resp = response.to_response()?;
  update_response_headers(&mut resp, server_ctx);
  Ok(resp)
}

pub async fn graphql_single_request(req: Request<Body>, server_ctx: &ServerContext) -> Result<Response<Body>> {
  graphql_request::<async_graphql_hyper::GraphQLRequest>(req, server_ctx).await
}

pub async fn graphql_batch_request(req: Request<Body>, server_ctx: &ServerContext) -> Result<Response<Body>> {
  graphql_request::<async_graphql_hyper::GraphQLBatchRequest>(req, server_ctx).await
}
```

So, we still have two different functions without dyn and they should compile to different code due to monomorphization

src/http/server.rs
meskill: ofc...

at least we can strip it to some degree

```rust
  let builder = hyper::Server::try_bind(&addr).map_err(CLIError::from)?;
  log_startup_messages(&addr, blueprint.server.enable_graphiql);

  Ok(if blueprint.server.enable_batch_requests {
    builder.serve(make_svc_batch_req).await
  } else {
    builder.serve(make_svc_single_req).await
  }.map_err(CLIError::from)?)
  ```

**Comment by sujeetsr** [comment]
I have separated the request processing into two functions, graphql_single_request and graphql_batch_request and avoided any dynamic calls. I have tried multiple ways of combining them into one function that can be called by http_spec , but, without passing an additional parameter, or making them dynamic, I haven't found a way to do it.

**Comment by meskill** [comment]
> I have separated the request processing into two functions, graphql_single_request and graphql_batch_request and avoided any dynamic calls. I have tried multiple ways of combining them into one function that can be called by http_spec , but, without passing an additional parameter, or making them dynamic, I haven't found a way to do it.

at first we want to figure out why would we want to do anything with http_spec. For now I see that we may want to run the same set of tests for batch and non-batch server to tests everything in different setup. To solve this we can just change the http_spec to run tests for single and batched versions of tests in some way. 

Do you have any concerns about this except the mentioned one?

File: tests/batched_request_spec.rs [hunk 10]
```diff
@@ -0,0 +1,70 @@
...
+
+    let unbatched_query = format!("{}", query1);
+    let response = make_request(http_client, unbatched_query).await;
+    let json = response.json::<Resp>().await.unwrap();
+    assert_eq!(
+      json.data.post.title,
+      "sunt aut facere repellat provident occaecati excepturi optio reprehenderit"
+    );
+  }
+}
```
tusharmath, line 70: Use the http_spec

ci-benchmark/benchmark.graphql
```diff
@@ -1,5 +1,6 @@
 schema
-  @server(port: 8000, baseURL: "http://jsonplaceholder.typicode.com", enableHttpCache: true, enableGraphiql: true) {
+  @server(port: 8000, enableGraphiql: true, enableQueryValidation: false, hostname: "0.0.0.0")
```
tusharmath, line 2:
```suggestion
  @server(port: 8000, enableGraphiql: true, enableQueryValidation: false)
```

PR #521, opened 2023-10-20
feat: compile time checks for nullable/undefined fields

### LINKED ISSUE #205: Compile time check for preventing use of optional fields in Http
**Description:**
The current compiler does not catch potential issues when a field that can be undefined or optional is used in the route parameter of Http. To enhance robustness and provide clearer developer feedback, the compiler should be improved in the following areas:

1. **Potentially Undefined Route Parameter:**
   When a field that can be undefined or optional is used in the route parameter of Http, the code should not compile to prevent potential runtime issues.

   **Examples:**
   ```graphql
   type Query {
   }
   ```

   In this first example, since `id` is an optional parameter, using it directly in the route should trigger a compiler error.

   ```graphql
   type Post {
      id: ID
      userId: ID
   }
   ```

   In the second example, the `userId` field is not guaranteed to always be present in a `Post`. Using it in the `path` should also raise a compiler error, as it can lead to unexpected behavior or failures.

2. **Undefined References:**
   The compiler should be enhanced to detect any values, fields, arguments, etc. that are used but not defined, and raise appropriate errors.

git-bruh:
**Summary:**  
This PR adds compile time checks for nullable/undefined fields & arguments as requested in #205

**Issue Reference(s):**  
Fixes #205

/claim #205

**Build & Testing:**

**Checklist:**

- [x] I have performed a self-review of my own code


git-bruh:
Example Run

```
...
type Post {
  id: Int!
  userId: Int
  title: String!
  body: String!
}
```

```sh
Error: Invalid Configuration
Caused by:
  • value 'userId' is a nullable type [at Post.user]
```

Another Example

```
...
type Query {
}
...
```

```sh
Error: Invalid Configuration
Caused by:
  • value 'userId' is a nullable type [at Post.user]
  • argument 'id' is a nullable type [at Query.user]
```

src/blueprint/from_config.rs
```diff
@@ -207,7 +208,101 @@ fn to_fields(type_of: &config::Type, config: &Config) -> Valid<Vec<blueprint::Fi
...
+    // So we must duplicate some of that logic here :(
+    if let Some(Expression::Unsafe(Operation::Endpoint(req_template, _, _))) = &field.resolver {
+      match &req_template.root_url {
+        Mustache(segments) => {
+          for segment in segments {
+            if let Segment::Expression(parts) = segment {
+              // TODO is this invariant enforced anywhere so that we can omit
+              // this check?
+              if parts.len() < 2 {
+                continue;
```
git-bruh, line 293:
```
  user: User! @http(path: "/users/{{value.userId}}")
  test: User @http(path: "/users/{{value.user.phone}}")
```

@tusharmath This is valid, right? Looks like will have to refactor a bit here to account for nested fields

src/mustache.rs
```diff
@@ -50,6 +50,22 @@ impl Mustache {
         .collect(),
     }
   }
+
+  pub fn all_parts(&self) -> Vec<&Vec<String>> {
```
tusharmath, line 54:
```suggestion
  pub fn segments(&self) -> Option<&Vec<String>> {
```

I think would be a better function signature.

src/blueprint/from_config.rs
```diff
@@ -207,7 +207,100 @@ fn to_fields(type_of: &config::Type, config: &Config) -> Valid<Vec<blueprint::Fi
...
+            .trace(&field.name)?;
+        }
+      }
+    }
+  }
+
+  Valid::Ok(())
 }
 
 fn to_field(
```
tusharmath, line 306: I think we can clean this up 
- Add a function that gets a type at a particular path 
  ```rust
  fn infer_type(&self, path: Vec<String>) -> Option<(&str, &Type)> {  

  }
  ```
  This function will just go thru the config and try to infer the type at a particular path.

- Then use the `segments` function from mustache to validate if that path exists.

tusharmath: - Also note there are scenarios where values are optional, so it shouldn't fail if the path doesn't exist.
- We need to make sure that the value at that particular path is a "string" or some "Scalar" type or else validation should fail.

src/mustache.rs
git-bruh: Hm, making the return type an `Option` of a `Vec` won't work here since we're collecting all the segments (`Vec`) into another `Vec`, not just returning one of the inner segments

src/blueprint/from_config.rs
git-bruh: I think there's already [`process_path`](https://github.com/tailcallhq/tailcall/blob/0061a2419211dbd4226a373bfda41ef57fe1f5a1/src/blueprint/from_config.rs#L534) that can infer types at a given path, given the `Config` struct

Now that is fine, but the actual traversal logic where we go through Mustache directives like `args`, `value`, etc. will have to remain in `validate_fields` only

So the logic would look something like so

```rs
fn validate_fields(...) {
    type_at_current_traversal = ...; // say `Post`
    match head {
        "value" => process_path([type_at_current_traversal, tail]), // say `[Post, userId]`
        "args" => process_path([??, tail]) // not sure how to get the "path" to an argument
}
```

Does this sound good?

git-bruh: Actually wait, I forgot that we already know the types of the fields here, so we can just check the `of_type` field for scalar values, just like how the nullable check is being performed right now..

src/mustache.rs
tusharmath: How would this work for an expression - `{{a.b}}c{{d.e}}`?

tests/graphql/test-http-tmpl.graphql
```diff
@@ -6,7 +6,7 @@ schema @server @upstream(baseURL: "http://jsonplacheholder.typicode.com") {
 type Post {
   id: Int
   user: User @http(path: "/users", query: [{key: "id", value: "{{value.userId}}"}])
-  userId: Int
+  userId: Int!
```
tusharmath, line 9: Query params are optional. In this case there shouldn't be any error.

src/mustache.rs
git-bruh: yes, that is intentional since i just indented for this to be a function to get the segments we care about for validation, so that i dont have to spam this pattern in multiple places where you `match` `self` with `Mustache` and then get the inner `segments`

git-bruh: renamed the function to be more explicit regarding this

git-bruh:
Pushed these changes

- made mustache helper name more descriptive https://github.com/tailcallhq/tailcall/pull/521/commits/323ff50fa1c29161cec62873ad55bf4cabe30da4

- allowed queries to use nullable values https://github.com/tailcallhq/tailcall/pull/521/commits/600112fda9fbd73eee1ec379c2cd8397c85ecca5

- disallowed usage of list types in templates https://github.com/tailcallhq/tailcall/pull/521/commits/e34c7a44880d80e7da7a6815d5193d131e306d18

- disallowed usage of non-scalar types in the `value` directive https://github.com/tailcallhq/tailcall/pull/521/commits/e06636a46e99d3f6709805928c900731cf932753

TODO

- [ ] streamline / deduplicate tests by returning multiple errors at once rather than the current fail-fast approach

- [ ] validate fields used from `vars` config

- [ ] finalize behaviour of `headers` directive (whether we can use it in templates or not)

src/blueprint/from_config.rs
```diff
@@ -207,7 +207,114 @@ fn to_fields(type_of: &config::Type, config: &Config) -> Valid<Vec<blueprint::Fi
       .trace(name)
   })?;
 
-  Ok(fields.into_iter().flatten().collect())
+  let fields = fields.into_iter().flatten().collect::<Vec<_>>();
+
+  validate_fields(&fields)?;
```
tusharmath, line 212: move the call to `validate_fields` above inside of the `validate_all` closure. Using `?` will short circuit and exit.

src/mustache.rs
```diff
@@ -50,6 +50,22 @@ impl Mustache {
...
+        for s in segments {
+          if let Segment::Expression(parts) = s {
+            out.push(parts);
+          }
+        }
+      }
+    }
+
+    out
+  }
```
tusharmath, line 68: nitpik: Can use declarative style

```rust
self.0.iter().filter_map(|a| match a {
      Segment::Expression(parts) => Some(parts),
      _ => None,
    })
    .collect()

```

src/blueprint/from_config.rs
git-bruh: i don't think we can do that in-place because a given field can rely on the construction of N number of other fields, so the validation must be done after all fields are known: https://github.com/tailcallhq/tailcall/pull/521/files#diff-0934da3dca97565bf71c4fd107919b495a6a971c292009709dfc4756760ac27eR287

tests/graphql/passed/test-nullable-values-query.graphql
```diff
@@ -0,0 +1,19 @@
+#> server-sdl
+schema {
+  query: Query
+}
+
+type Query {
+  user(id: ID!): User @http(path: "/user/{{args.id}}", baseURL: "http://localhost:8080")
```
tusharmath, line 7:
```suggestion
  user(id: ID!): User @http(path: "/user/{{args.id}}", baseURL: "http://localhost:8080")
  users(id: ID): [User] @http(path: "/user", query: [{key: "id", value: "{{args.id}}"}], baseURL: "http://localhost:8080")
```

tests/graphql/passed/test-nullable-values-query.graphql
```diff
@@ -0,0 +1,19 @@
...
+type User {
+  id: ID!
+  name: String!
+}
+
+type Post {
+  id: Int!
+  userId: Int
+  user: User @http(path: "/users", query: [{key: "id", value: "{{value.userId}}"}], baseURL: "http://localhost:8080")
+}
```
tusharmath, line 19: We should add a `#> client-sdl` to make sure we are able to generate the correct schema.

tests/graphql/errors/test-directives-undef-null-fields.graphql
```diff
@@ -0,0 +1,47 @@
...
+  user_null_value: User @http(path: "/users/{{value.userId}}", baseURL: "http://localhost:8080")
+  # nullable values are allowed in queries
+  user_null_value_query: User
+    @http(path: "/users", query: [{key: "id", value: "{{value.id}}"}], baseURL: "http://localhost:8080")
+  user_undefined_value: User @http(path: "/users/{{value.userid}}", baseURL: "http://localhost:8080")
+  # but not undefined values
+  user_undefined_value_query: User
+    @http(path: "/users", query: [{key: "id", value: "{{value.userid}}"}], baseURL: "http://localhost:8080")
+  user_vars: User @http(path: "/users/{{vars.a}}", baseURL: "http://localhost:8080")
+}
```
tusharmath, line 34: Use camelCase for all graphQL fields.

src/blueprint/from_config.rs
```diff
@@ -238,6 +339,8 @@ fn to_field(
   let field_definition = update_unsafe(field.clone(), field_definition);
   let field_definition = update_const_field(field, field_definition, config).trace("@const")?;
   let field_definition = update_inline_field(type_of, field, field_definition, config).trace("@inline")?;
+  validate_field(type_of, config, &field_definition)?;
```
tusharmath, line 342: Can we move this call into `update_http`?

tusharmath (approved):

PR #537, opened 2023-10-23
Merge @groupBy and @http directives

### LINKED ISSUE #463: Merge `@groupBy` and `@http`
Currently `groupBy` operator is used for batching and only supports Http. Even the implementation currently is very http specific. We can merge them.

**Before:**
```graphql
type Post {
  id: Int!
  userId: Int!
  title: String!
  body: String!
  user: User @http(path: "/users", query: [{key: "id", value: "{{value.userId}}"}]) @groupBy(key: "id", path: ["id"])
}
```

**After:**
```graphql
type Post {
  id: Int!
  userId: Int!
  title: String!
  body: String!
  user: User 
    @http(
      path: "/users", query: [{key: "id", value: "{{value.userId}}"}],
      batch: {
        groupByPath: ["id"], 
        selectKey: "id"
      }
    )
}
```

**IMPORTANT:**
Update the documentation on https://github.com/tailcallhq/tailcallhq.github.io

tobihans:
**Summary:**  
The @http directive now include the groupBy option.

- Old version
```graphql
type Post {
  id: Int!
  userId: Int!
  title: String!
  body: String!
  user: User @http(path: "/users", query: [{key: "id", value: "{{value.userId}}"}]) @groupBy(path: ["id"])
}
```

- New version
```graphql
type Post {
  id: Int!
  userId: Int!
  title: String!
  body: String!
  user: User @http(path: "/users", query: [{key: "id", value: "{{value.userId}}"}], groupBy: ["id"])
}
```

**Issue Reference(s):**  
Fixes #463 

**Build & Testing:**

**Checklist:**


src/config/config.rs
```diff
@@ -221,8 +220,8 @@ pub struct Field {
   pub http: Option<Http>,
   #[serde(rename = "unsafe")]
   pub unsafe_operation: Option<Unsafe>,
-  #[serde(rename = "groupBy")]
-  pub group_by: Option<GroupBy>,
+  // #[serde(rename = "groupBy")]
+  // pub group_by: Option<GroupBy>,
```
tusharmath, line 224: delete commented code.

src/config/config.rs
```diff
@@ -244,7 +243,11 @@ impl Field {
     directives
   }
   pub fn has_batched_resolver(&self) -> bool {
-    self.group_by.is_some()
+    if let Some(http) = &self.http {
+      !http.group_by.is_empty()
+    } else {
+      false
+    }
```
tusharmath, line 250:
```suggestion
    self.http.iter().all(|http| !http.group_by().is_empty())
```
May be this is a bit more concise. 
Can we add a quick unit test in the same file `config.rs` under `mod tests`. I think it will make sure we don't break this functionality later.

tobihans: Done. I finally used this instead:
```rust
self.http.as_ref().is_some_and(|http| !http.group_by.is_empty())
```

tobihans:
I've update the code according to the latest reviews, @tusharmath.
Thanks

tusharmath:
@tobihans can you do a quick rebase.

tobihans:
Hi @tusharmath , sorry for the late response. I've merged locally. Thanks.

src/blueprint/from_config.rs
```diff
@@ -456,6 +433,11 @@ fn update_http(
         let output_schema = to_json_schema_for_field(field, config);
         let input_schema = to_json_schema_for_args(&field.args, config);
 
+        // Short-circuiting validation for group_by
+        if !http.group_by.is_empty() && http.method != Method::GET {
```
tusharmath, line 437:
```suggestion
        if is_batch_request {
```
Create a common variable that check if the request once.

src/blueprint/from_config.rs
```diff
@@ -456,6 +433,11 @@ fn update_http(
...
+          return Valid::fail("GroupBy is only supported for GET requests".to_string());
```
tusharmath, line 438: We shouldn't return from here. 

```rust
Valid::fail("GroupBy is only supported for GET requests".to_string())
  .when(|| !http.group_by.is_empty() && http.method != Method::GET)
  .and(/* other valid instance */)
```

src/blueprint/from_config.rs
tobihans: The later suggestion invalidates this, I think.

tusharmath (approved):

Issue #574, opened 2023-11-04
GroupBy does not work when the type of the grouped field is a list


sujeetsr:
### Prerequisites

- [x] I have checked the existing issues to ensure this is not a duplicate.
- [x] I am running the latest version of Tailcall.

### Describe the bug
If the field with the batched api is a list, with a groupBy, an error is returned while querying

```graphql
type User {
  id: Int!
  name: String!
  username: String!
  email: String!
  phone: String
  website: String
  posts: [Post] @http(path: "/posts?userId={{value.id}}" groupBy: ["userId'])
}
```

tusharmath:
fixed in #571

PR #595, opened 2023-11-07
test: add request handle tests

### LINKED ISSUE #576: Add Integration Tests for Request Handler
We take inspiration from `GraphQLSpec` which makes it quite easy to write tests for Schema.

**Approach**
Write the an api-specification as a JSON/YML/JSON5 or some other format that one is comfortable with. Deserialize it as `APISpecification`:

```rust
struct APIRequest {
  method: Method,
  uri: Uri,
  headers: HeaderMap,
  body: Option<serde_json::Value>
}

struct APIResponse {
  status: u8,
  headers: HeaderMap,
  body: Option<serde_json::Value>
}

struct UpstreamRequest(APIRequest);
struct UpstreamResponse(APIResponse);
struct DownstreamRequest(APIRequest);
struct DownstreamResponse(APIResponse);

struct DownstreamAssertion {
  request: DownstreamRequest,
  response: DownstreamResponse
}

struct APISpecification {
  config: Config,
  name: String,
  description: Option<String>,
  upstream_mocks: Vec<UpstreamResponse>,
  expected_upstream_requests: Vec<UpstreamRequest>, 
  downstream_assertions: Vec<DownstreamAssertion>,
}
```

To test the specification we will need to extract a request_handler in a way that it will have two implementations — 
- A real implementation (current one).
- A test implementation that allows mocking and passing test values as inputs.

**Technical Requirements**
- Test only the request handler
- The test shouldn't need to start a real web-server

amitksingh1490:
fixes #576


src/http/method.rs
```diff
@@ -13,6 +13,22 @@ pub enum Method {
...
+      Method::PUT => "PUT",
+      Method::PATCH => "PATCH",
+      Method::DELETE => "DELETE",
+      Method::HEAD => "HEAD",
+      Method::OPTIONS => "OPTIONS",
+      Method::CONNECT => "CONNECT",
+      Method::TRACE => "TRACE",
+    }
+  }
+}
```
tusharmath, line 30: Serde should help.

tests/data/sample.yaml
```diff
@@ -0,0 +1,54 @@
+---
+config: tests/graphql_mock/test-custom-headers.graphql
+name: Example API
+description: This is a sample API specification.
+upstream_mocks:
+  - - method: GET
```
tusharmath, line 6: default method should be GET, and shouldn't require any specification.

tests/data/sample.json
```diff
@@ -0,0 +1,81 @@
...
+          "Content-Type": "application/json"
+        },
+        "body": {
+          "status": "ok",
+          "message": "Data processed correctly"
+        }
+      }
+    }
+  ]
+}
```
tusharmath, line 81: Write a test for the bug fixed in #574

tests/api_spec.rs
```diff
@@ -0,0 +1,145 @@
...
+
+  #[derive(Default, Deserialize)]
+  pub struct APISpecification {
+    pub config: String,
+    pub name: String,
+    pub description: Option<String>,
+    pub upstream_mocks: Vec<(UpstreamRequest, UpstreamResponse)>,
+    pub expected_upstream_requests: Vec<UpstreamRequest>,
+    pub downstream_assertions: Vec<DownstreamAssertion>,
+  }
```
tusharmath, line 53: Take inspiration from https://github.com/tailcallhq/tailcall/issues/603
To allow developers to control which spec to run/ignore or expect to fail.

tests/http_spec.rs
```diff
@@ -0,0 +1,215 @@
...
+      .url
+      .clone()
+      .unwrap_or(Url::parse("http://localhost:8080/graphql").unwrap());
+    let state = spec.setup().await;
+    let req = Request::builder()
+      .method(method)
+      .uri(url.as_str())
+      .body(Body::from(query_string))
+      .unwrap();
+    Ok(graphql_request(req, state.as_ref()).await?)
```
meskill, line 213: sometimes instead of putting `unwrap` in many places it could be simper to use `?` and pass the error to the caller and on caller side we could have single unwrap or handle the error the way we want

tests/http_spec.rs
```diff
@@ -0,0 +1,215 @@
...
+
+  #[tokio::test]
+  async fn test_body_yaml() {
+    let spec = HttpSpec::read("tests/data/sample.yaml").unwrap();
+    assert_downstream(spec).await;
+  }
+
+  async fn assert_downstream(spec: HttpSpec) {
+    for downstream_assertion in spec.clone().downstream_assertions.iter() {
+      if let Some(annotation) = downstream_assertion.annotation.clone() {
```
meskill, line 145: adding additional default variant to annotation could simplify using it here

tests/http_spec.rs
```diff
@@ -0,0 +1,215 @@
...
+                .request
+                .0
+                .body
+                .clone()
+                .unwrap_or(serde_json::Value::Null)
+            );
+            println!("Skipping test in: {}\nRequest Details: {}", spec.name, request_details);
+            continue;
+          }
+          Annotation::Only => {
```
meskill, line 167: how does only should work here? 
I think in current implementation all of the tests before test with `only` will still be executed anyway

tests/http_spec.rs
```diff
@@ -0,0 +1,215 @@
...
+          Annotation::Skip => {
+            let request_details = format!(
+              "Method: {:?}, Path: {}, Body: {}",
+              downstream_assertion.request.0.method.clone().unwrap_or_default(),
+              downstream_assertion
+                .request
+                .0
+                .url
+                .clone()
+                .unwrap_or(Url::parse("http://localhost:8080/graphql").unwrap()),
```
meskill, line 156: I see the similar code below so it could be deduplicated, for example we can put this into method of downstream_assertion

tests/http_spec.rs
```diff
@@ -0,0 +1,215 @@
...
+    async fn execute(&self, req: reqwest::Request) -> anyhow::Result<Response> {
+      let upstream_mocks = self.upstream_mocks.clone();
+      let upstream_mock = upstream_mocks.iter().find(|(upstream_request, _res)| {
+        let method_match = req.method().as_str()
+          == serde_json::to_string(&upstream_request.0.method.clone().unwrap_or_default())
+            .unwrap()
+            .as_str()
+            .trim_matches('"');
+        let url_match = req.url().as_str() == req.uri().to_string().as_str();
+        method_match && url_match
```
meskill, line 113: splitting the big `execute` function into smaller functions could significantly improve readability of the code

amitksingh1490: Http_client trait ony has execute fn defined defining the function outside of trait is not a good idea, added comments for readability

tests/http_spec.rs
amitksingh1490: ideally it should fail if URL is not given added a expect

tests/http_spec.rs
amitksingh1490: fixed

tests/http_spec.rs
amitksingh1490: simplified

tests/http_spec.rs
```diff
@@ -0,0 +1,215 @@
+#[cfg(test)]
+mod test {
```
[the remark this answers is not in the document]
amitksingh1490: done

tests/http_spec.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub annotation: Option<Annotation>,
+  }
+
+  #[derive(Default, Deserialize, Clone)]
+  pub struct HttpSpec {
+    pub config: String,
+    pub name: String,
+    pub description: Option<String>,
+    pub upstream_mocks: Vec<(UpstreamRequest, UpstreamResponse)>,
+    pub expected_upstream_requests: Vec<UpstreamRequest>,
```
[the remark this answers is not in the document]
amitksingh1490: used to assert upstream in execute

tests/http_spec.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub downstream_assertions: Vec<DownstreamAssertion>,
+  }
+
+  impl HttpSpec {
+    fn read(spec: &str) -> Option<Self> {
```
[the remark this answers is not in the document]
amitksingh1490: done

tests/http_spec.rs
amitksingh1490: done

tests/http_spec.rs
amitksingh1490: default is not needed we not annotation is needed case only can run

tests/http_spec.rs
amitksingh1490: fixed

tests/http_spec.rs
amitksingh1490: removed the option as it was not needed

tests/data/sample.json
```diff
@@ -0,0 +1,63 @@
+{
+  "config": "tests/graphql_mock/test-custom-headers.graphql",
+  "name": "Example API",
```
tusharmath, line 3: add proper names to files depicting what they are doing.

tests/http_spec.rs
```diff
@@ -0,0 +1,216 @@
...
+    "Method: {:?}, Path: {}, Body: {}",
+    request.method.clone(),
+    request.url.clone(),
+    request.body.clone()
+  )
+}
+
+#[tokio::test]
+async fn test_body_json() {
+  let spec = HttpSpec::read("tests/data/sample.json").unwrap();
```
tusharmath, line 202:
```suggestion
  let spec = HttpSpec::read("tests/http/sample.json").unwrap();
```

Like `/graphql` folder, let's keep `/http` folder.

tests/http_spec.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    // Set the body of the response.
+    response.body = ConstValue::try_from(serde_json::from_value::<Value>(mock_response.0.body)?)?;
+
+    Ok(response)
+  }
+}
+
+#[tokio::test]
+async fn test_body_yaml() {
```
tusharmath, line 153: We should detect the format and then load the files. Shouldn't write special handlers for file extensions.

tests/http_spec.rs
tusharmath: Secondly, we need to load all the files at once. We shouldn't need to update the spec code every time we add a new file.

tests/http_spec.rs
```diff
@@ -0,0 +1,216 @@
...
+pub struct UpstreamResponse(APIResponse);
+#[derive(Deserialize, Clone)]
+pub struct DownstreamRequest(pub APIRequest);
+#[derive(Deserialize, Clone)]
+pub struct DownstreamResponse(pub APIResponse);
+#[derive(Deserialize, Clone)]
+pub struct DownstreamAssertion {
+  pub request: DownstreamRequest,
+  pub response: DownstreamResponse,
+  pub annotation: Option<Annotation>,
```
tusharmath, line 54: I think the annotation should be at Spec level. The spec succeeds or fails. It's possible that we apply a `skip` annotation for an assertion, but because the mock working correctly because of a batching bug, the spec throws up an error.

tests/http_spec.rs
```diff
@@ -0,0 +1,216 @@
...
+  let has_only_annotation = spec
+    .downstream_assertions
+    .iter()
+    .any(|assertion| matches!(assertion.annotation, Some(Annotation::Only)));
+
+  for downstream_assertion in spec.downstream_assertions.iter() {
+    match &downstream_assertion.annotation {
+      Some(Annotation::Skip) if !has_only_annotation => {
+        let request_details = format_request_details(&downstream_assertion.request.0);
+        println!("Skipping test in: {}\nRequest Details: {}", spec.name, request_details);
```
tusharmath, line 168: Use log.warn instead of println.

tests/http_spec.rs
```diff
@@ -0,0 +1,216 @@
...
+async fn run(spec: HttpSpec, downstream_assertion: &&DownstreamAssertion) -> anyhow::Result<hyper::Response<Body>> {
+  let query_string = serde_json::to_string(&downstream_assertion.request.0.body).expect("body is required");
+  let method = downstream_assertion.request.0.method.clone();
+  let url = downstream_assertion.request.0.url.clone();
+  let state = spec.setup().await;
+  let req = Request::builder()
+    .method(method)
+    .uri(url.as_str())
+    .body(Body::from(query_string));
+  graphql_request(req?, state.as_ref()).await
```
tusharmath, line 215: - [ ] Log `ok` after each file is executed. Like we do it in GraphQLSpec.
- [ ] Make sure tests for each file run in parallel.

tests/http_spec.rs
```diff
@@ -0,0 +1,216 @@
...
+      .extension()
+      .ok_or(anyhow!("not a valid extension"))?
+      .to_str()
+      .ok_or(anyhow!("not a valid Unicode"))?;
+
+    let spec = match extension.to_lowercase().as_str() {
+      "json" => anyhow::Ok(serde_json::from_str(&contents)?),
+      "yaml" => anyhow::Ok(serde_yaml::from_str(&contents)?),
+      _ => Err(anyhow!("only json and yaml are supported")),
+    };
```
tusharmath, line 81: Can leverage `source.rs` instead of duplicating code.

tests/http_spec.rs
```diff
@@ -0,0 +1,216 @@
...
+
+  for downstream_assertion in spec.downstream_assertions.iter() {
+    match &downstream_assertion.annotation {
+      Some(Annotation::Skip) if !has_only_annotation => {
+        let request_details = format_request_details(&downstream_assertion.request.0);
+        println!("Skipping test in: {}\nRequest Details: {}", spec.name, request_details);
+        continue;
+      }
+      Some(Annotation::Only) | None
+        if !has_only_annotation || matches!(downstream_assertion.annotation, Some(Annotation::Only)) =>
```
tusharmath, line 172: The filtering of `only` `skip` `fail` should happen across all files, not just the downstream assertion.

amitksingh1490: Done

tests/http_spec.rs
amitksingh1490: done

tests/http_spec.rs
amitksingh1490: Done

tests/http_spec.rs
amitksingh1490: done

tests/http_spec.rs
amitksingh1490: done

tests/http_spec.rs
amitksingh1490: done

tests/http_spec.rs
amitksingh1490: done

tests/data/sample.json
amitksingh1490: done

tests/http_spec.rs
```diff
@@ -0,0 +1,248 @@
...
+        serde_json::to_string(&downstream_assertion.response.0.body).unwrap()
+      )
+    }
+  }
+  log::info!("{} {} ... ok", spec.name, spec.path.display());
+}
+#[tokio::test]
+async fn test_body() -> std::io::Result<()> {
+  let spec = HttpSpec::cargo_read("tests/http").unwrap();
+  let spec = HttpSpec::filter_specs(spec);
```
meskill, line 227: so, we actually building a test runner here 😄
I wonder if something like this https://crates.io/crates/datatest could help us with that

tusharmath: They all use macro APIs, and I think we will end up using a lot of  statics also to share things between tests. I agree the current way isn't the best but I think we need a nicer library that can do more for us.

PR #602, opened 2023-11-08
refactor: use source structs instead of sdl properties on GraphQLSpec struct

### LINKED ISSUE #480: Refactor GraphQLSpec
To support more SDLs type with tags we should do the following:

**From:**
```rust
struct GraphQLSpec {
  path: PathBuf,
  client_sdl: String,
  server_sdl: Vec<String>,
  merged_server_sdl: String,
  sdl_errors: Vec<SDLError>,
  test_queries: Vec<GraphQLQuerySpec>,
}
```

**To:**

```rust
struct Source {
  tag: String, // could be #> client-sdl or #>server-sdl or #>foo-bar
  sdl: String
}

struct GraphQLSpec {
  path: PathBuf,
  sdl_sources: Vec<Source>
  sdl_errors: Vec<SDLError>,
  test_queries: Vec<GraphQLQuerySpec>,
}

impl GraphQLSpec {
  fn find_source(tag: String) -> Vec<Source> {
     todo!()
  }
}
```

ologbonowiwi:
**Summary:**  
Use vector of source structs instead of plain sdl properties on `GraphQLSpec`

**Issue Reference(s):**  
Fixes #480

**Build & Testing:**

**Checklist:**

### DIFF
--- a/tests/graphql_spec.rs
+++ b/tests/graphql_spec.rs
@@ -23,14 +23,67 @@ mod graphql_mock;
 
 static INIT: Once = Once::new();
 
+#[derive(Debug, Clone)]
+struct Source {
+  sdl: String,
+  tag: String,
+}
+
 #[derive(Debug, Default, Setters)]
 struct GraphQLSpec {
   path: PathBuf,
-  client_sdl: String,
-  server_sdl: Vec<String>,
-  merged_server_sdl: String,
+  sources: Vec<Source>,
   sdl_errors: Vec<SDLError>,
   test_queries: Vec<GraphQLQuerySpec>,
+}
+
+impl GraphQLSpec {
+  fn client_sdl(mut self, sdl: String) -> Self {
+    self.sources.push(Source {
+      sdl,
+      tag: CLIENT_SDL.to_string(),
+    });
+    self
+  }
+
+  fn server_sdl(mut self, sdl: Vec<String>) -> Self {
+    for s in sdl {
+      self.sources.push(Source {
+        sdl: s,
+        tag: SERVER_SDL.to_string(),
+      });
+    }
+    self
+  }
+
+  fn merged_server_sdl(mut self, sdl: String) -> Self {
+    self.sources.push(Source {
+      sdl,
+      tag: MERGED_SDL.to_string(),
+    });
+
+    self
+  }
+
+  fn find_source(&self, tag: &str) -> String {
+    self
+      .sources
+      .clone()
+      .into_iter()
+      .find(|s| s.tag == tag)
+      .unwrap()
+      .sdl
+  }
+
+  fn get_sources(&self, tag: &str) -> Vec<String> {
+    self
+      .sources
+      .clone()
+      .into_iter()
+      .filter(|s| s.tag == tag)
+      .map(|s| s.sdl)
+      .collect()
+  }
 }
 
 #[derive(Debug, Default, Deserialize, Serialize, PartialEq)]
@@ -175,7 +228,8 @@ fn test_config_identity() -> std::io::Result<()> {
   let specs = GraphQLSpec::cargo_read("tests/graphql");
 
   for spec in specs? {
-    let content = spec.server_sdl[0].as_str();
+    let content = spec.find_source(SERVER_SDL);
+    let content = content.as_str();
     let expected = content;
 
     let config = Config::from_sdl(content).to_result().unwrap();
@@ -193,10 +247,14 @@ fn test_server_to_client_sdl() -> std::io::Result<()> {
   let specs = GraphQLSpec::cargo_read("tests/graphql");
 
   for spec in specs? {
-    let expected = spec.client_sdl;
-    let content = spec.server_sdl[0].as_str();
+    let expected = spec.find_source(SERVER_SDL);
+    let expected = expected.as_str();
+    let content = spec.find_source(CLIENT_SDL);
+    let content = content.as_str();
     let config = Config::from_sdl(content).to_result().unwrap();
+    // error is on the line below
     let actual = print_schema::print_schema((Blueprint::try_from(&config).unwrap()).to_schema());
+    log::info!("ClientSDL: {} ... not ok yet", spec.path.display());
     assert_eq!(actual, expected, "ClientSDL: {}", spec.path.display());
     log::info!("ClientSDL: {} ... ok", spec.path.display());
   }
@@ -216,7 +274,7 @@ async fn test_execution() -> std::io::Result<()> {
     .into_iter()
     .map(|spec| {
       tokio::spawn(async move {
-        let mut config = Config::from_sdl(&spec.server_sdl[0]).to_result().unwrap();
+        let mut config = Config::from_sdl(spec.find_source(SERVER_SDL).as_str()).to_result().unwrap();
         config.server.enable_query_validation = Some(false);
 
         let blueprint = Valid::from(Blueprint::try_from(&config))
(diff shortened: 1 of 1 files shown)


tests/graphql_spec.rs
```diff
@@ -193,10 +232,14 @@ fn test_server_to_client_sdl() -> std::io::Result<()> {
...
   for spec in specs? {
-    let expected = spec.client_sdl;
-    let content = spec.server_sdl[0].as_str();
+    let expected = spec.find_source(SERVER_SDL);
+    let expected = expected.as_str();
+    let content = spec.find_source(CLIENT_SDL);
+    let content = content.as_str();
     let config = Config::from_sdl(content).to_result().unwrap();
+    // error is on the line below
     let actual = print_schema::print_schema((Blueprint::try_from(&config).unwrap()).to_schema());
```
ologbonowiwi, line 241: I'm trying to understand why this line is throwing an error; can I get some help here? :eyes:

tests/graphql_spec.rs
```diff
@@ -23,16 +23,54 @@ mod graphql_mock;
...
-  client_sdl: String,
-  server_sdl: Vec<String>,
-  merged_server_sdl: String,
+  sources: Vec<Source>,
   sdl_errors: Vec<SDLError>,
   test_queries: Vec<GraphQLQuerySpec>,
 }
 
+impl GraphQLSpec {
+  fn client_sdl(mut self, sdl: String) -> Self {
```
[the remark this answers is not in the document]
tusharmath: can also use `get_sources` instead.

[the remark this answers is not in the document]
ologbonowiwi: sure, I'll use `get_sources`; the tag is not `"client-sdl"`?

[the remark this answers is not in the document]
tusharmath: this is a getter and not a setter.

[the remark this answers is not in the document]
ologbonowiwi: I'm using `get_sources` on the getter as suggested; I'm resolving this.

if you have any further comments, let me know

tests/graphql_spec.rs
```diff
@@ -193,10 +232,14 @@ fn test_server_to_client_sdl() -> std::io::Result<()> {
...
 
   for spec in specs? {
-    let expected = spec.client_sdl;
-    let content = spec.server_sdl[0].as_str();
+    let expected = spec.find_source(SERVER_SDL);
+    let expected = expected.as_str();
+    let content = spec.find_source(CLIENT_SDL);
+    let content = content.as_str();
     let config = Config::from_sdl(content).to_result().unwrap();
+    // error is on the line below
```
ologbonowiwi, line 240: I figured out the `.graphql` file throwing this error is `/tests/graphql/test-upstream.graphql`; does this make any sense?

tests/graphql_spec.rs
```diff
@@ -23,16 +23,54 @@ mod graphql_mock;
 
 static INIT: Once = Once::new();
 
+#[derive(Debug, Clone)]
+struct Source {
+  sdl: String,
+  tag: String,
```
meskill, line 29: wouldn't it be better if use enum with explicit variants instead of arbitrary string?

tusharmath: Agree with @meskill

tests/graphql_spec.rs
tusharmath: What's the error?

tests/graphql_spec.rs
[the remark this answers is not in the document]
tusharmath: Why do we need a setter?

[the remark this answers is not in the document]
ologbonowiwi: we used a setter before but via directives; I'm keeping the same approach as before (I chose this way to avoid significant changes), but it is not required

tests/graphql_spec.rs
ologbonowiwi:
```
thread 'test_server_to_client_sdl' panicked at 'called `Result::unwrap()` on an `Err` value: ValidationError([Cause { message: "No resolver has been found in the schema", description: None, trace: ["Query", "hello"] }])', tests/graphql_spec.rs:242:75
```

ologbonowiwi: I've mentioned on [this comment](https://github.com/tailcallhq/tailcall/pull/602#discussion_r1386887309), but this error is appearing when executing `/tests/graphql/test-upstream.graphql`

ologbonowiwi: I've fixed it; thanks @tusharmath

tests/graphql_spec.rs
```diff
@@ -23,16 +23,61 @@ mod graphql_mock;
...
+  }
+
+  fn get_sources(&self, tag: Tag) -> Vec<String> {
+    self
+      .sources
+      .clone()
+      .into_iter()
+      .filter(|s| s.tag == tag)
+      .map(|s| s.sdl)
+      .collect()
```
meskill, line 77: it is not a big deal in tests but could be a good pattern to follow especially when writing app's code:

instead of heavily use `clone` and `collect` we can use iterators as is and do any actions only in the end. So here we can return `impl Iterator` and consume it later

ologbonowiwi: Regarding `collect`, it makes a lot of sense (I've implemented the change already), but I could not eliminate the clone.

I've tried implementing the Copy trait to avoid cloning it, but I have had no success here. Thanks for the feedback :smile:

tests/graphql_spec.rs
```diff
@@ -23,16 +23,65 @@ mod graphql_mock;
...
+  }
+
+  fn get_sources(&self, tag: Tag) -> impl Iterator<Item = String> {
+    self
+      .sources
+      .clone()
+      .into_iter()
+      .filter(move |s| s.tag == tag)
+      .map(|s| s.sdl)
+  }
```
meskill, line 82: yeah, good start, and you can improve it even more!
- I think we can get do just `self.source.iter()` to not clone anything from the beginning - we'll have to work with references there but this is even better
- for doing `.first()` we don't even need to collect anything as it can be replaced by `.next()` on iterator itself
- returning String in some cases could be overcopying things and if you can use  &str it's better use it

ologbonowiwi: > for doing .first() we don't even need to collect anything as it can be replaced by .next() on iterator itself

done. nice catch!

> I think we can get do just self.source.iter() to not clone anything from the beginning - we'll have to work with references there but this is even better

it does not work because `Source` does not implement the `Copy` trait (I've tried to implement it myself but gave up after a lot of work).

> returning String in some cases could be overcopying things and if you can use &str it's better use it

while this makes a lot of sense, I'm struggling to implement it; in the end, I'd have way more `.clone()` to have it working

ologbonowiwi:
🚀

tests/graphql_spec.rs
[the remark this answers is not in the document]
tusharmath: @ologbonowiwi I think we can remove the setters.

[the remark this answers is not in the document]
ologbonowiwi: done @tusharmath

tusharmath (approved):

tusharmath:
@ologbonowiwi Can you resolve the conflicts please.

ologbonowiwi:
checks passing, we can merge it :rocket:

tests/graphql_spec.rs
meskill: this code could look like this to minimize cloning

```rust
  fn find_source(&self, tag: Tag) -> Option<&str> {
    self.get_sources(tag).next()
  }

  fn get_sources(&self, tag: Tag) -> impl Iterator<Item = &str> {
    self
      .sources
      .iter()
      .filter(move |s| s.tag == tag)
      .map(|s| s.sdl.as_str())
  }
  ```

ologbonowiwi: done, @meskill, it worked pretty well, thank you :smile:

ologbonowiwi: I did not succeed on changing `find_source` signature, ended up having issues with it but for `get_sources` it worked and we're not cloning anymore.

meskill: yeah it requires changes in usage as well in that case, anyway it is minor thing no need to spend time on this

meskill (approved):

tusharmath (approved):

Issue #606, opened 2023-11-09
Fail with an error if no platform build is found


tusharmath:
**Steps**
```bash
npm i -g @tailcallhq/tailcall
```

Because all platform dependencies are optional, sometimes nothing gets installed. In that case, there should be an error issued stating that the platform is not yet supported, and the user should create an issue on our github repository.

Min2who:
can you assign this to me @tusharmath

Min2who:
@tusharmath I can use an external library right

Min2who:
@tusharmath is this approach correct
![Screenshot (83)](https://github.com/tailcallhq/tailcall/assets/127925465/79a931c4-e20c-4e43-893f-3bcf342c4544)

jkcs:
@tusharmath Regarding this issue, I have a few questions. Are you suggesting the following: 
1. When users globally install `@tailcallhq/tailcal`l using `npm i -g @tailcallhq/tailcall`, perform a platform compatibility check?
2. When users install `@tailcallhq/tailcall` within a project using `npm i @tailcallhq/tailcall`, also perform the compatibility check?

Current: Only when installing the package within a project (not globally), the corresponding platform-specific implementation package is directly installed.

If you want the code to execute during global installation, placing it in the postinstall script will not work. If you want to achieve execution during global installation, you can assign this issue to me.

sc231997:
I am also facing a similar issue my os:linux and arch:x64 and none of the packages are getting installed.

PR #608, opened 2023-11-09
Added the error #606

Min2who:
/claim #606


npm/post-install.cjs
ologbonowiwi, line 1: can you keep `.js` as an extension?

since we're not declaring module type on `package.json`, it'll be CJS by default

ologbonowiwi: Oh I noticed now that we're using modules; why not keep using modules and use CJS instead?

npm/post-install.cjs
```diff
@@ -0,0 +1,20 @@
+var os = require('os');
+var fs = require('fs');
+
+// Define the valid platform builds
+var validPlatformBuilds = ['win32', 'darwin', 'linux'];
```
ologbonowiwi, line 5: can you change the variables to `const` instead of using `var`?

PR #609, opened 2023-11-09
feature: load configuration via HTTP

### LINKED ISSUE #441: Load Config via HTTP

ssddOnTop:
**Summary:**  
This PR adds ability to load config over HTTP with polling (extension of #499). With addition of some features like checking type of file received via http.

**Issue Reference(s):**  
Fixes #441 
/claim #441 
/split @neo773 30%

**Build & Testing:**

**Checklist:**


ssddOnTop:
Documentation is pending, I want someone to review the code and suggest changes before I update documentation.

amitksingh1490:
please resolve conflicts

ssddOnTop:
Sorry I don't know how I closed the PR but this was accidental, can you please repoen the pr @amitksingh1490 @tusharmath

amitksingh1490:
> Sorry I don't know how I closed the PR but this was accidental, can you please repoen the

There are no changes on the PR thats why couldn't reopen
image

ssddOnTop:
@amitksingh1490 can you please review

ssddOnTop:
@amitksingh1490 I removed the unnecessary derives  (i.e. clone in config and schema loader) but is there anything else I need to change?

amitksingh1490:
@ssddOnTop Great start on this 👏🏽 

Few functionalities need to be implemented
1. Ability to provide Combination of http and local files
1. Multiple http paths (Like we allow multiple files)
1. cli documentation update file-path should change to path and doc should tell it accepts http and file system paths
image

4. Polling interval doc should specify its in secs

Some non functional requirements
1. Reuse the httpclient which is already created in server context.
1. In log level info mode when schema is updated log the diff ++ and --
1. logs need to specify whether schema has changed or not change, if no change is detected do not update the context.
1. Add tests for config loader

src/config/config.rs
```diff
@@ -350,11 +350,23 @@ impl Config {
...
     }
   }
-
   pub fn n_plus_one(&self) -> Vec<Vec<(String, String)>> {
     super::n_plus_one::n_plus_one(self)
   }
 
+  pub async fn from_file_or_url(file_paths: std::slice::Iter<'_, String>) -> Result<(Config, Option<String>)> {
+    let first_path = file_paths.clone().next().unwrap();
+    if first_path.starts_with("http://") || first_path.starts_with("https://") {
```
amitksingh1490, line 359: should work with multiple paths of different types

src/config/config_poll.rs
```diff
@@ -0,0 +1,78 @@
...
+  refresh_interval: u64,
+  state: Arc<AtomicPtr<ServerContext>>,
+}
+
+impl ConfigLoader {
+  pub fn new(file_path: String, refresh_interval: u64, state: Arc<AtomicPtr<ServerContext>>) -> Result<Self, CLIError> {
+    Ok(Self { file_path, refresh_interval, state })
+  }
+
+  pub async fn start_polling(&self) {
```
amitksingh1490, line 24: should have exponential backoff in case of failures

src/config/config_poll.rs
```diff
@@ -0,0 +1,78 @@
...
+        let resp = match resp {
+          Ok(resp) => resp,
+          Err(e) => {
+            log::error!("Failed to refresh configuration: {}", e);
+            continue;
+          }
+        };
+
+        if !resp.status().is_success() {
+          log::info!("Unknown error.");
```
amitksingh1490, line 49: errors should be logged as is and log.error should be used for that

amitksingh1490 (changes requested):

ssddOnTop:
On it

src/config/config_poll.rs
ssddOnTop: Servers might be facing temporary downtime or might have got a temporary rate limit, in those case throwing exceptions won't be ideal, it checks everything while starting the server, and I think that should be enough

amitksingh1490: We ought to avoid throwing exceptions; instead, they should be recorded as errors. Moreover, continuous polling of a server that has surpassed its rate limit is counterproductive. Implementing an exponential backoff strategy in case of errors is essential to effectively manage and eventually overcome the rate limit issue.

ssddOnTop:
hi @amitksingh1490 its been 4 days, can you please post an update

amitksingh1490:
I am getting segmentation fault when I stress test this using `wrk` for a simple query
image
Steps to reproduce
1. Start the server using 
`cargo run --release -- start https://raw.githubusercontent.com/tailcallhq/tailcall/test/config-from-http/examples/jsonplaceholder.graphql --poll-interval 10 --log-level error`
2. run wrk using
`wrk -t4 -c100 -d10s -s wrk.lua http://localhost:8000/graphql`
wrk.lua for your reference
```lua
wrk.method = "POST"
wrk.body = '{"operationName":null,"variables":{},"query":"{ posts { title } }"}'
wrk.headers["Connection"] = "keep-alive"
wrk.headers["User-Agent"] = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36"
wrk.headers["Content-Type"] = "application/json"
```

What we can do is lets split this into two PR's 
1. Ability to load configs via Http.
2. Polling capability

Ability to load configs via Http seems ok apart from some improvement comments you have yourself left in your code.
Polling capability needs to be tested thoroughly and get rid of segmentation failures and other errors if any.
Let me know if this is ok with you

amitksingh1490 (changes requested):

src/config/source.rs
```diff
@@ -34,4 +35,44 @@ impl Source {
...
+    // needs improvement
+    match serde_json::from_str::<serde_json::Value>(val) {
+      Ok(_) => Ok(Self::Json),
+      Err(_) => match serde_yaml::from_str::<serde_yaml::Value>(val) {
+        Ok(_) => Ok(Self::Yml),
+        Err(_) => match async_graphql::parser::parse_schema(val) {
+          Ok(_) => Ok(Self::GraphQL),
+          Err(e) => Err(anyhow!(e)),
+        },
+      },
```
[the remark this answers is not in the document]
amitksingh1490: also avoid excessive pattern matching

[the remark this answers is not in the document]
ssddOnTop: Shouldn't it remain a part of source.rs as it's just detecting the type of file?

src/config/source.rs
```diff
@@ -29,9 +29,14 @@ impl Source {
   }
 
   pub fn detect(name: &str) -> Result<Source, UnsupportedFileFormat> {
-    ALL
-      .into_iter()
-      .find(|format| format.ends_with(name))
-      .ok_or_else(|| UnsupportedFileFormat(name.to_string()))
+    let re = regex::Regex::new(r"\.(graphql|yml|yaml|json)$").unwrap();
```
meskill, line 32: why introduce regex here? looks like we still doing the same iteration later

tests/custom_resp_header_spec.rs
```diff
@@ -37,4 +32,17 @@ mod integration_tests {
...
+  async fn test_verify_response_headers() {
+    let schema_path = "tests/graphql_mock/test-custom-headers.graphql";
+    verify_response_headers(schema_path).await;
+  }
+
+  #[tokio::test]
+  async fn test_verify_response_headers_over_http() {
+    let schema_path =
+      "https://raw.githubusercontent.com/tailcallhq/tailcall/main/tests/graphql_mock/test-custom-headers.graphql";
+    verify_response_headers(schema_path).await;
```
meskill, line 46: so the test could fail in case of disabled internet?
that could be unexpected, lets at least add some comment about this in code

src/http/server.rs
```diff
@@ -66,8 +69,9 @@ fn create_allowed_headers(headers: &HeaderMap, allowed: &BTreeSet<String>) -> He
 
   new_headers
 }
-pub async fn start_server(config: Config) -> Result<()> {
-  let blueprint = Blueprint::try_from(&config).map_err(CLIError::from)?;
+
+pub async fn start_server_with_url(config: &Config) -> Result<()> {
```
meskill, line 73: does it worth renaming? it doesn't bring more sense for me

src/config/config.rs
```diff
@@ -350,33 +347,23 @@ impl Config {
...
+    for file_path in file_paths {
+      config_reader.serialize_config(file_path).await?;
     }
+    Ok(config_reader)
+  }
 
-    Ok(config)
+  pub async fn from_file_path(file_path: &String) -> Result<Config> {
+    let source = Source::detect(file_path)?;
+    let server_sdl = ConfigReader::read_file(file_path).await;
```
meskill, line 365: I have a little confusion about relationship between Config and ConfigReader as they both use each other. Could we consider this to make only single direction relationship? Maybe ConfigReader that creates Config with their methods and Config will contain only config entries without logic about how to parse it?

src/config/read.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct ConfigReader {
+  config: Config,
+}
+
+impl ConfigReader {
+  pub fn init() -> Self {
+    Self { config: Config::default() }
+  }
+  pub async fn serialize_config(&mut self, path: &String) -> anyhow::Result<()> {
+    let conf = if path.starts_with("http://") || path.starts_with("https://") {
```
meskill, line 15: we could just try parse this to [Url](https://docs.rs/reqwest/0.11.22/reqwest/struct.Url.html) and not check the schema manually and actually be able to use any other protocol that is supported by the reqwest

src/config/source.rs
ssddOnTop: @tusharmath suggested me to do the same

src/config/config.rs
ssddOnTop: Yes but when you need instance of config then Config::from_* makes sense however it calls ConfigReader for I/O but ConfigReader also needs instance of Config to merge them

this didn't make sense but I definitely wanted review on this 

as you suggested, I will move source detection to config reader itself

src/config/read.rs
ssddOnTop: The goal of this pr was to add fetch over HTTP

the code would be very different for fetching with… say websocket, can we discuss this on Discord?

tests/custom_resp_header_spec.rs
ssddOnTop: without internet connection, several cases in graphql_spec.rs will fail as well. We can not really test this offline

src/config/read.rs
meskill: the main goal that I wanted to achieve with my comment is to simplify the code and make it easier to extend in the future if we need to. For now we a targeting only http as you rightly pointed out.

tag me on discord if you want to discuss anything further

src/config/read.rs
```diff
@@ -0,0 +1,49 @@
...
+
+pub struct ConfigReader {
+  config: Config,
+}
+
+impl ConfigReader {
+  pub fn init() -> Self {
+    Self { config: Config::default() }
+  }
+  pub async fn serialize_config(&mut self, path: &str) -> anyhow::Result<()> {
```
tusharmath, line 14:
```suggestion
  pub async fn read(&mut self, path: &str) -> anyhow::Result<()> {
```

src/config/config.rs
```diff
@@ -350,33 +347,22 @@ impl Config {
...
-        Ok(conf) => config = config.clone().merge_right(&conf),
-        Err(e) => return Err(e), // handle error
-      }
+  pub async fn from_file_or_url(file_paths: std::slice::Iter<'_, String>) -> Result<ConfigReader> {
+    // maintaining config reader will help to implement poll intervals in future
+    let mut config_reader = ConfigReader::init();
+    for file_path in file_paths {
+      config_reader.serialize_config(file_path).await?;
     }
+    Ok(config_reader)
```
tusharmath, line 360:
```suggestion
    Ok(config_reader.get_config())
```

src/config/config.rs
```diff
@@ -350,33 +347,22 @@ impl Config {
...
-        Config::from_source(source, &server_sdl)
-      })
-      .collect();
-
-    for res in join_all(futures).await {
-      match res {
-        Ok(conf) => config = config.clone().merge_right(&conf),
-        Err(e) => return Err(e), // handle error
-      }
+  pub async fn from_file_or_url(file_paths: std::slice::Iter<'_, String>) -> Result<ConfigReader> {
```
tusharmath, line 354:
```suggestion
  pub async fn from_file_or_url(file_paths: std::slice::Iter<'_, String>) -> Result<Config> {
```

Let's return a config for now. Until we know exactly how polling etc. should be implemented.

src/config/config.rs
```diff
@@ -350,33 +347,22 @@ impl Config {
...
+    // maintaining config reader will help to implement poll intervals in future
+    let mut config_reader = ConfigReader::init();
```
tusharmath, line 356:
```suggestion
    let mut config_reader = ConfigReader::init(file_paths);
```

src/config/config.rs
```diff
@@ -350,33 +347,22 @@ impl Config {
...
+    for file_path in file_paths {
+      config_reader.serialize_config(file_path).await?;
     }
+    Ok(config_reader)
+  }
 
-    Ok(config)
+  pub async fn from_file_path(file_path: &str) -> Result<Config> {
```
tusharmath, line 363: Move this function to `ConfigReader`

src/config/mod.rs
```diff
@@ -4,6 +4,7 @@ pub mod group_by;
 mod into_document;
 mod key_values;
 mod n_plus_one;
+mod read;
```
tusharmath, line 7:
```suggestion
mod reader;
```

src/config/read.rs
ssddOnTop: This will just create unnecessary clones and reduce performance

src/config/reader.rs
```diff
@@ -0,0 +1,55 @@
+use std::slice::Iter;
+
+use tokio::fs::File;
+use tokio::io::AsyncReadExt;
+
+use crate::config::{Config, Source};
+
+pub struct ConfigReader {
+  config: Config,
```
tusharmath, line 9:
```suggestion
```

src/config/reader.rs
```diff
@@ -0,0 +1,55 @@
...
+        Config::from_source(source, &st)?
+      } else {
+        let path = path.trim_end_matches('/');
+        Self::from_file_path(path).await?
+      };
+      self.config = self.config.clone().merge_right(&conf);
+    }
+    Ok(self.config.clone())
+  }
+  pub async fn from_file_path(file_path: &str) -> anyhow::Result<Config> {
```
tusharmath, line 30:
```suggestion
  async fn from_file_path(file_path: &str) -> anyhow::Result<Config> {
```

src/config/reader.rs
```diff
@@ -0,0 +1,55 @@
...
+    let (server_sdl, source) = ConfigReader::read_file(file_path).await?;
+    Config::from_source(source, &server_sdl)
+  }
+  pub async fn read_file(file_path: &str) -> anyhow::Result<(String, Source)> {
```
tusharmath, line 34:
```suggestion
  async fn read_file(file_path: &str) -> anyhow::Result<(String, Source)> {
```

tests/custom_resp_header_spec.rs
```diff
@@ -37,4 +32,17 @@ mod integration_tests {
...
+    let schema_path = "tests/graphql_mock/test-custom-headers.graphql";
+    verify_response_headers(schema_path).await;
+  }
+
+  #[tokio::test]
+  async fn test_verify_response_headers_over_http() {
+    let schema_path =
+      "https://raw.githubusercontent.com/tailcallhq/tailcall/main/tests/graphql_mock/test-custom-headers.graphql";
+    verify_response_headers(schema_path).await; //might fail while testing without internet
+  }
```
tusharmath, line 47: I think we should add tests for the reader insider of `reader.rs`.

ssddOnTop: @tusharmath it should be placed in tests/reader.rs as we'll need files to test IO and hardcoding example/* will be a problem.

sujeetsr: @ssddOnTop Why is this spec modified? Its only testing that custom headers set in the config file are set in the response. Its not related to how the config file is loaded.

ssddOnTop: @sujeetsr I don't think it was necessary as it was already being tested by local file path. Running the tests with github url just rate limits the developer

src/config/reader.rs
```diff
@@ -0,0 +1,95 @@
...
+    server
+      .mock("GET", "/")
+      .with_status(200)
+      .with_header("content-type", "application/graphql")
+      .with_body(TEST_GQL_BODY)
+      .create();
+    let files: Vec<String> = [
+      "examples/jsonplaceholder.yml", // from file
+      "http://localhost:3080/",       // with content-type header
+      "https://raw.githubusercontent.com/tailcallhq/tailcall/main/examples/jsonplaceholder.json", // with url extension
```
sujeetsr, line 86: Why a github link, can't you use the mock server for this case too?

src/config/reader.rs
```diff
@@ -0,0 +1,227 @@
...
+          "website": {
+            "type_of": "String"
+          }
+        }
+      }
+    },
+    "unions": {}
+  }
+}
+  "#;
```
tusharmath, line 168: Instead of writing a config as string u could create one using Config:: default () and then convert it to sdl and other formats. There is an API to convert config to yaml json and graphql format.

tusharmath (commented):
Overall looks good. Just one comment

tusharmath (approved):

amitksingh1490 (approved):

PR #614, opened 2023-11-09
Adds `only`,`fail` and `skip` annotations to `GraphQLSpec`

### LINKED ISSUE #603: Add support for `only` `skip` and `failing` annotations in `GraphQLSpec`
Currently there is no way for a developer to control which spec to run, ignore or expect to fail.

```diff
 struct GraphQLSpec {
   path: PathBuf,
   client_sdl: String,
   server_sdl: Vec<String>,
   merged_server_sdl: String,
   sdl_errors: Vec<SDLError>,
   test_queries: Vec<GraphQLQuerySpec>,
+  annotation: Option<Annotation>
 }

+ enum Annotation {Skip, Only, Fail}
```

The annotations will contain `skip` or `only` or `failing`

**Technical Specifications**
- `only`: If any spec contains `only` only that spec should be executed. Eg:
  ```graphql
  #> spec-only
  
  #> server-sdl
  schema { query: Query }
  
  #> client-sdl
  schema {query: Query}
  ```

- `skip`: if a spec contains `skip` then that should be skipped, and a warn log should be printed. Eg:

  ```graphql
  #> spec-skip
  
  #> server-sdl
  schema { query: Query }
  
  #> client-sdl
  schema {query: Query}
  ```

- `fail`: Expect the spec to fail. Panic incase it doesn't fail. Eg:
  ```graphql
  #> spec-fail
  
  #> server-sdl
  schema { query: Query }
  
  #> client-sdl
  schema {query: Query}
  ```

tobihans:
**Summary:**  
This PR add annotations that will be useful to mark specs to should be ran, skipped or should fail.

**Issue Reference(s):**  
Fixes #603 

**Build & Testing:**

**Checklist:**

/claim #603


tests/graphql_spec.rs
```diff
@@ -151,7 +168,15 @@ impl GraphQLSpec {
       if path.is_file() && path.extension().unwrap_or_default() == "graphql" {
         let contents = fs::read_to_string(path.clone())?;
         let path_buf = path.clone();
-        files.push(GraphQLSpec::new(path_buf, contents.as_str()));
+        let spec = GraphQLSpec::new(path_buf, contents.as_str());
+
+        match spec.annotation {
+          Some(Annotation::Only) | None => files.push(spec),
+          Some(Annotation::Skip) => {
+            log::warn!("Skipping spec: {}", spec.path.display());
```
amitksingh1490, line 176:
```suggestion
            log::warn!("{} ... skipped", spec.path.display());
```

tests/graphql_spec.rs
```diff
@@ -151,7 +168,15 @@ impl GraphQLSpec {
...
+          }
+          Some(Annotation::Fail) => {}
```
amitksingh1490, line 178: The annotation `Fail` should be designed so that it passes when the assertion fails, and conversely, it should fail if the assertion succeeds.

amitksingh1490 (changes requested):

tests/graphql_spec.rs
```diff
@@ -151,7 +168,15 @@ impl GraphQLSpec {
       if path.is_file() && path.extension().unwrap_or_default() == "graphql" {
         let contents = fs::read_to_string(path.clone())?;
         let path_buf = path.clone();
-        files.push(GraphQLSpec::new(path_buf, contents.as_str()));
+        let spec = GraphQLSpec::new(path_buf, contents.as_str());
+
+        match spec.annotation {
```
amitksingh1490, line 173: Also if no annotation is given all tests should run as it is. Ideally you should create two vectors one for only one for all, if only is empty run all if only is non empty run only

amitksingh1490 (approved):

PR #624, opened 2023-11-12
Integrate CI for running benchmarks

### LINKED ISSUE #436: Integrate CI for Running Benchmarks
### Overview

We need to enhance our CI pipeline to not only run our existing Criterion benchmarks but also ensure that the build time stays within acceptable limits. Moreover, a clear and concise report should be generated and published on the PR for quick and easy understanding.

### Requirements

1. **Run Criterion Benchmarks:** Our CI should execute the existing Criterion benchmarks as part of the pipeline.
2. **Fail on Degradation:** The CI should fail if there's a performance degradation when compared to the `main` branch.
3. ~**Build Time Check:** The CI should compare the build time of the current PR with the average build time of the last few successful builds on the `main` branch. If the build time increases by more than 10%, the CI should fail.~
4. **Benchmark Report:** A tabular comparison of benchmarks should be published on the PR, similar to the code-cov report. It should display the current benchmark results alongside the results from the `main` branch for easy comparison.
5. **Conditional Execution:** Benchmarks and build time checks should only be executed when the `benchmark` label is added to a PR or commit.
6. **Support for Forks:** The CI process should be designed to run benchmarks on forks as well.

### Rationale

Performance is paramount for our project. It's essential to ensure that any changes do not adversely affect either the runtime performance (tracked by benchmarks) or the build time. By integrating these checks into our CI, we can maintain a high standard of performance and ensure that contributors receive immediate feedback on any potential issues.

alankritdabral:
**Summary:**  
I have used Criterion benchmarks to check performance degradation and iai-callgrind to check increment in build time

 https://github.com/marketplace/actions/criterion-compare-prs **Criterion benchmarks comparision**
 https://lib.rs/crates/iai-callgrind **iai-callgrind benchmarks**

**Check this pr for implementation of the changes https://github.com/alankritdabral/tailcall/pull/16#issue-1989326231**

**Issue Reference(s):**  
Fixes #436

**Build & Testing:**

**Checklist:**

/claim #436


.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
           files: lcov.info
-          fail_ci_if_error: true
+          fail_ci_if_error: false
```
ologbonowiwi, line 63: Is this change necessary for the scope of this PR?

```suggestion
          fail_ci_if_error: true
```

alankritdabral: The ci was failing, then decided to keep it false

.github/scripts/build_time_compare.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+file1_url="https://raw.githubusercontent.com/alankritdabral/tailcall/main/benches/iai-callgrind/benchmarks.txt"
```
meskill, line 3: could we specify it as a parameter so we could change it when we will want compare with another branch?

benches/iai-callgrind/impl_path_string_for_evaluation_context_iai-callgrind.rs
meskill, line 1: do you have ideas on how we can avoid duplication? 

can something like features or shared modules help with that?

.github/workflows/bot.yml
```diff
@@ -148,3 +148,100 @@ jobs:
...
+    if: github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'benchmark')
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Run json_like_bench
+        if: contains(github.event.pull_request.labels.*.name, 'benchmark')
+        uses: boa-dev/criterion-compare-action@v3.2.4
+        with:
+          benchName: json_like_bench
```
meskill, line 164: do we need to specify different steps for every file with benchName? or we can wrap it in single step for the whole set?
criterion-compare states that benchName is optional

AlooDon: If we wrap it in single step it shows (Unrecognized option: 'save-baseline') which on further solving(https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options) the issue i couldnt solve it but i deepdive into criterion-compare which uses (https://github.com/BurntSushi/critcmp/blob/master/README.md) to compare benchmarks and this critcmp is known to work with Criterion 0.3.3. and we are using 0.5.3 version 
These are my views feel free to correct me

.github/scripts/build_time_compare.sh
```diff
@@ -0,0 +1,79 @@
...
+        fi
+
+        x=$((x + total_read_write))
+        y=$((y + estimated_cycles))
+    done
+
+    echo "----------------------------------"
+done
+
+exit $fail_ci
```
tusharmath, line 79: Maybe shell script isn't the best choice for so much of logical processing. Thoughts on using `cargo script` @meskill @alankritdabral ?

.github/scripts/build_time_compare.sh
```diff
@@ -0,0 +1,79 @@
...
+# Fetching file1 from the specified URL
+curl -s "$file1_url" > file1.txt || { echo "Failed to download file from $file1_url"; exit 1; }
+
+benchmarks=(
+    "json_like_bench_iai_callgrind::batched_body::benchmark_batched_body"
+    "data_loader_bench_iai_callgrind::data_loader::benchmark_data_loader"
+    "impl_path_string_for_evaluation_context_iai_callgrind::bench::bench_main"
+    "request_template_bench_iai_callgrind::bench_to_request::benchmark_to_request"
+    # Add more benchmarks here as needed
+)
```
tusharmath, line 15: We shouldn't have to update this file everytime we add a new benchmark.

.github/workflows/bot.yml
tusharmath: I agree with @meskill, I think we need a way to not have to update the bot.yml every time we add a new benchmark.

.github/workflows/bot.yml
```diff
@@ -148,3 +148,100 @@ jobs:
...
+        uses: boa-dev/criterion-compare-action@v3.2.4
+        with:
+          benchName: data_loader_bench
+          defaultFeatures: false
+          branchName: ${{ github.base_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run request_template_bench
+        if: contains(github.event.pull_request.labels.*.name, 'benchmark')
+        uses: boa-dev/criterion-compare-action@v3.2.4
```
tusharmath, line 189: The compare-action should comment on the PR, but it's not?

AlooDon:
I will be closing this Pr i have a different and simple way @tusharmath 
but when comparing criterion benchmarks what should i keep the  criteria for degradation (keeping in mind all thee vaalues we have aree in nano seconds unit )

for build time it is 10%

and should i only print criterion benchmarks on Pr like a codecov report

PR #633, opened 2023-11-14
fix: npm publish issue with glib

### LINKED ISSUE #628: Replace postinstall script for npm packages with libc field in package.json
### Prerequisites

We have a scripts inside `./npm` that generates npm packages published on npm that allow for easier installation of tailcall cli on different systems.

### Describe the bug

1. `npm/post-install.js` doesn't work as expected due to usage of the result of async call inside sync code. So, the final name of the package is wrong and no package is installed actually. The script came from https://github.com/tailcallhq/tailcall/pull/607
2. on linux despite system's libc both gnu and musl versions are installed (i.e. twice of required space is occupied)  and I can't run "tc" command probably due to the naming clash. The script from above in theory should fix the problem on linux but it's actually does nothing. At least on linux, on other os I expect optionalDependencies should work 

### Expected behavior

The easiest way to fix this would be using `libc` field as it is [should be supported](https://github.com/npm/rfcs/issues/438) by npm and other package managers. At least swc uses [this approach](https://github.com/swc-project/swc/pull/4046)

After that the postinstall script could be removed. 

The main concern would be only compatibility with legacy versions of npm and other package managers

### Testing published packages

We want to make sure that everything will work as expected and we won't face new problems with that. To do that we may want to implement tests for published packages in ci. 

The implementation could look like this:
1. for main target systems (windows, macos, unix-glibc, unix-musl) run test job
2. use local registry as [verdaccio](https://github.[com/verdaccio/github-actions) to run local registry in ci
3. publish npm packages in local registry (may require changing publish script to specify registry)
4. create temporary dir
5. install `@tailcallhq/tailcall` from running local registry
6. make sure that commands `tc --version` works
7. also show the output of `npm ls -a` to verify that only required packages are installed

ologbonowiwi:
- build: add libc to generated package.json
- refactor: delete post-install script

**Summary:**  
Replacing post-install script by libc field on `package.json` (which is handled automatically by Node.js)

**Issue Reference(s):**  
Fixes #628

**Build & Testing:**

**Checklist:**


neo773:
This might get problematic as support for `libc` field landed quite recently in [npm@9.6.5](https://github.com/npm/rfcs/issues/438#issuecomment-1648118990)

We'll also have to account for third party package managers like yarn and pnpm

ologbonowiwi:
I know that yarn, pnpm and bun already supports, but I didn't knew about the version restriction on NPM. 

maybe we should close this @meskill?

meskill:
> This might get problematic as support for `libc` field landed quite recently in [npm@9.6.5](https://github.com/npm/rfcs/issues/438#issuecomment-1648118990)
> 
> We'll also have to account for third party package managers like yarn and pnpm

meskill:
> This might get problematic as support for `libc` field landed quite recently in [npm@9.6.5](https://github.com/npm/rfcs/issues/438#issuecomment-1648118990)
> 
> We'll also have to account for third party package managers like yarn and pnpm

there is already npm@10 and the v9 is shipped with the node18 that soon will enter maintence stage. So, we could actually just support those active versions for npm and probably for other package managers as well. 

I suggest also to take a look into swc repo and issues and search for any issues they have with libc field as I saw they are using this approach

npm/post-install.js
```diff
@@ -1,17 +1,14 @@
 // @ts-check
-import { family, GLIBC, MUSL } from "detect-libc"
+// @ts-ignore
```
tusharmath, line 2:
```suggestion
```

tusharmath (approved):

jkcs:
@tusharmath Seems like forgetting to remove unused dependencies：https://github.com/tailcallhq/tailcall/blob/0fb53ce0ba54aa4a3ace3caf133d969764e23938/npm/gen-root.ts#L50-L52

PR #634, opened 2023-11-14
rename binary to tailcall & fix libc detection

### LINKED ISSUE #628: Replace postinstall script for npm packages with libc field in package.json

neo773:
- Changed `libc` detection to use synchronous call instead of async
- Renamed binary to `tailcall` from `tc` as it clashes with built in utility with the same name on linux https://man7.org/linux/man-pages/man8/tc.8.html

/closes #628

### DIFF
--- a/npm/post-install.js
+++ b/npm/post-install.js
@@ -1,5 +1,6 @@
 // @ts-check
-import { family, GLIBC, MUSL } from "detect-libc"
+// @ts-ignore
+import { familySync, GLIBC, MUSL } from "detect-libc"
 import { exec } from 'child_process'
 import util from 'util'
 
@@ -7,11 +8,7 @@ const execa = util.promisify(exec)
 const platform = process.platform
 const arch = process.arch
 
-let libcFamily
-family().then((fam) => {
-  libcFamily = fam
-})
-
+const libcFamily = familySync()
 let libc
 if (platform === "win32") {
   libc = "-msvc"
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ The below file is a standard `.graphQL` file, with a few additions such as `@ser
 Now, run the following command to start the server with the full path to the jsonplaceholder.graphql file that you created above.
 
 ```bash
-tc start ./jsonplaceholder.graphql
+tailcall start ./jsonplaceholder.graphql
 ```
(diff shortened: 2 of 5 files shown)


npm/post-install.js
```diff
@@ -1,17 +1,14 @@
...
 const execa = util.promisify(exec)
 const platform = process.platform
 const arch = process.arch
 
-let libcFamily
-family().then((fam) => {
-  libcFamily = fam
-})
-
+const libcFamily = familySync()
```
ologbonowiwi, line 11: we can't use top-level await?

amitksingh1490:
@neo773 pulled your changes in #633 thanks for this 🍻

PR #654, opened 2023-11-18
Support for updating schema periodically

ssddOnTop:
**Summary:**  
Added ability to update the schema on given time interval.

**Issue Reference(s):**  
Fixes https://github.com/tailcallhq/tailcall/issues/441
/claim https://github.com/tailcallhq/tailcall/issues/441

**Build & Testing:**

**Checklist:**


tusharmath:
You might want to rebase the last two commits onto main

ssddOnTop:
yup did that

src/cli/command.rs
```diff
@@ -27,6 +27,9 @@ pub enum Command {
     file_path: Vec<String>,
     #[arg(long)]
     log_level: Option<log::Level>,
+    /// The interval after which the schema will be re read and updated automatically
+    #[arg(long)]
+    poll_interval: Option<u64>,
```
tusharmath, line 32:
```suggestion
    poll: Option<u64>,
```

src/cli/tc.rs
```diff
@@ -18,17 +19,27 @@ pub async fn run() -> Result<()> {
...
+          log::info!("N + 1: {}", config.n_plus_one().len().to_string());
+          start_server(config, None).await?;
+        }
+        Some(poll_interval) => {
+          let config = Config::from_file_or_url(file_path.clone()).await?;
+          log::info!("N + 1: {}", config.n_plus_one().len().to_string());
+          let config_loader = ConfigLoader::init(file_path, poll_interval);
+          start_server(config, Some(config_loader)).await?;
+        }
+      }
```
tusharmath, line 38: Can we move the polling logic into a separate module/file inside of CLI.

src/config/reader.rs
```diff
@@ -28,7 +26,7 @@ impl ConfigReader {
     }
     Ok(config)
   }
-  async fn from_file_path(file_path: &str) -> anyhow::Result<Config> {
+  pub(crate) async fn from_file_path(file_path: &str) -> anyhow::Result<Config> {
```
tusharmath, line 29:
```suggestion
  pub async fn from_file_path(file_path: &str) -> anyhow::Result<Config> {
```

Probably `pub(crate)` makes more sense, but to be consistent we can keep it `pub`

examples/foo.graphql
```diff
@@ -0,0 +1,7 @@
+schema @server(port: 8000) {
+  query: Query
+}
+
+type Query {
+  hello: String! @const(data: "Hi")
+}
```
tusharmath, line 7: Do we need this file?

wrk.lua
```diff
@@ -0,0 +1,5 @@
+wrk.method = "POST"
+wrk.body = '{"operationName":null,"variables":{},"query":"{ hello } "}'
+wrk.headers["Connection"] = "keep-alive"
+wrk.headers["User-Agent"] = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36"
+wrk.headers["Content-Type"] = "application/json"
```
tusharmath, line 5: Do we need this file?

tusharmath:
Left one is `main` and the right one is with `polling`
Screenshot 2023-11-28 at 12 22 35 AM

mywrk.lua
```diff
@@ -0,0 +1,86 @@
...
+
+use url::Url;
+
+use crate::blueprint::Blueprint;
+use crate::config::reader::ConfigReader;
+use crate::config::Config;
+use crate::http::ServerContext;
+
+pub struct ConfigLoader {
+  file_paths: HashMap<String, u64>,
```
tusharmath, line 14: do we need a hashmap?

src/config/config_poll.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use url::Url;
+
+use crate::blueprint::Blueprint;
+use crate::config::reader::ConfigReader;
+use crate::config::Config;
+use crate::http::ServerContext;
+
+pub struct ConfigLoader {
+  file_paths: HashMap<String, u64>,
```
tusharmath, line 13: Why is filepaths a hashmap?

src/http/request_handler.rs
```diff
@@ -60,7 +60,10 @@ pub async fn graphql_request<T: DeserializeOwned + GraphQLRequestLike>(
   let request = serde_json::from_slice::<T>(&bytes);
   match request {
     Ok(request) => {
-      let mut response = request.data(req_ctx.clone()).execute(&server_ctx.schema).await;
+      let mut response = request
+        .data(req_ctx.clone())
+        .execute(&*server_ctx.schema.read().await)
```
tusharmath, line 65: Reading the schema asynchronously isn't a great idea. `read` should be sync. It should be updated asynchronously.

ssddOnTop: > Reading the schema asynchronously isn't a great idea. `read` should be sync. It should be updated asynchronously.

@tusharmath `tokio::sync::RwLock` is better in our case as `std::sync::RwLock` do not have any order guaranteed, the decision to provide read or write lock depends upon the OS, but in case of `tokio::sync::RwLock` if there is a request for write lock on at the head of the queue, it will not let out read locks 

referance: [std::sync::RwLock](https://doc.rust-lang.org/std/sync/struct.RwLock.html) and [tokio::sync::RwLock](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html)

ssddOnTop: @tusharmath but `std::sync::RwLockReadGuard` do not have `Send` trait, hence can't be used with threads anyways

mywrk.lua
meskill, line 1: is this some test file that accidently was committed?

ssddOnTop: yes I used it to benchmark my code, I accidentally committed it here

src/http/request_handler.rs
meskill: do we have clear solution here?

we anyway need to some kind of locking to update the config from another task, right? So it will introduce some overhead anyway

ssddOnTop: Yes, I tried with AtomicPtr, bit heap memory is slow and I can't find a way to get it perfectly synced. I think RwLock is the only solution I know as of now.

mywrk.lua
ssddOnTop: Yes in case of any failure, we are increasing timeout for each path/url, so this map stores the path and it's timeout

tusharmath: why should it be different?

src/cli/command.rs
```diff
@@ -27,6 +27,9 @@ pub enum Command {
     file_path: Vec<String>,
     #[arg(long)]
     log_level: Option<log::Level>,
+    /// It is the interval after which the schema will be re read and updated automatically
```
tusharmath, line 30:
```suggestion
    /// The interval after which the configuration files that are served over HTTP will be reloaded
```

src/config/config_poll.rs
tusharmath: I don't think we need to store poll at an individual file level.

src/config/config_poll.rs
```diff
@@ -0,0 +1,84 @@
...
+impl ConfigLoader {
+  pub fn init(file_paths: Vec<String>, poll_interval: u64) -> ConfigLoader {
+    let mut path_hm = HashMap::new();
+    for file_path in file_paths {
+      path_hm.insert(file_path, poll_interval);
+    }
+    Self { file_paths: path_hm, poll_interval }
+  }
+  pub async fn start_polling(&mut self, batch_single: Arc<ServerContext>) {
+    loop {
```
tusharmath, line 26: Instead of `loop` it would be better to only query every `n` units of time. Internally keep a last fetch time and request it only if the time has expired. Essentially behaving as TTL and not Polling. This way we don't need to worry about leaks and using `drop`.

tusharmath: The other advantage with using a TTL is the ability to set that in the config instead of passing it via the CLI.

src/http/server_context.rs
```diff
@@ -9,7 +10,7 @@ use crate::http::HttpDataLoader;
 use crate::lambda::{Expression, Unsafe};
 
 pub struct ServerContext {
-  pub schema: dynamic::Schema,
+  pub schema: RwLock<dynamic::Schema>,
```
tusharmath, line 13: Try to write code that you can reason about by looking at locally. For eg: Adding internal mutability of a particular type, and then not using it anywhere within the server context can be confusing. One way to fix this is by moving the polling/reloading logic into server context, and making all fields private. I think this is also a problem with http_client.

src/config/config_poll.rs
ssddOnTop: > Instead of `loop` it would be better to only query every `n` units of time. Internally keep a last fetch time and request it only if the time has expired. Essentially behaving as TTL and not Polling. This way we don't need to worry about leaks and using `drop`.

So you mean that TTL can be checked when user queries something and if it's expired then we try to fetch new schema?

tusharmath:
We need further thinking of how reloads should be implemented. The current implementation is fairly thorough but may this isn't within the scope of the tailcall server. Configuration management should be done via external tools ideally.

PR #657, opened 2023-11-18
fix: cache-control-tests #652

### LINKED ISSUE #652: Add tests for `cache-control` header using http_spec
Write tests under `/http` to assert if the cache control headers are being forwarded correctly when set in any upstream responses. Also test if minimum is computed correctly in case multiple requests sent upstream have different cache control headers.

ezhil56x:
**Summary:**  
Added tests for `cache-control` under `/http`

**Issue Reference(s):**  
Fixes #652
/claim #652

**Build & Testing:**

**Checklist:**

### DIFF
--- a/tests/http/cache-control.yml
+++ b/tests/http/cache-control.yml
@@ -1,0 +1,44 @@
+config: !file tests/http/config/simple.graphql
+name: Cache Control Header Test
+mock:
+  - request:
+      url: http://jsonplaceholder.typicode.com/users/1
+    response:
+      body:
+        id: 1
+        name: foo
+      headers:
+        Cache-Control: "max-age=3600, public"
+  - request:
+      url: http://jsonplaceholder.typicode.com/users/2
+    response:
+      body:
+        id: 2
+        name: bar
+      headers:
+        Cache-Control: "max-age=7200, public"
+assert:
+  - request:
+      method: POST
+      url: http://localhost:8080/graphql
+      body:
+        query: query { user { name } }
+    response:
+      body:
+        data:
+          user:
+            name: foo
+      headers:
+        Cache-Control: "max-age=3600, public"
+  - request:
+      method: POST
+      url: http://localhost:8080/graphql
+      body:
+        query: query { user { name } }
+    response:
+      body:
+        data:
+          user:
+            name: foo
+      headers:
+        Cache-Control: "max-age=7200, public"


A-N-uraag (changes requested):
Hi @ezhil56x, just adding the file under `tests/http` would not be testing the `cache-control` headers. Relevant assertions need to be added in the `http_spec_e2e` test. For example, try running the tests by just changing the header value in assertion to `max-age=36, public`. All tests would still pass.

ologbonowiwi (approved):
LGTM

tests/http/cache-control.yml
```diff
@@ -0,0 +1,44 @@
...
+        data:
+          user:
+            name: foo
+      headers:
+        Cache-Control: "max-age=3600, public"
+  - request:
+      method: POST
+      url: http://localhost:8080/graphql
+      body:
+        query: query { user { name } }
```
amitksingh1490, line 37: The query should data from two different API calls and minimum of the `cache-control` headers should received at downstream.

amitksingh1490: The goal is to test if the headers are calculated properly

tests/http/cache-control.yml
```diff
@@ -0,0 +1,44 @@
+config: !file tests/http/config/simple.graphql
```
tusharmath, line 1: [optional] You can checkout `!inline` and write the configuration inside the spec. That makes it easier to understand what's happening in the spec.

A-N-uraag:
> Hi @ezhil56x, just adding the file under `tests/http` would not be testing the `cache-control` headers. Relevant assertions need to be added in the `http_spec_e2e` test. For example, try running the tests by just changing the header value in assertion to `max-age=36, public`. All tests would still pass.

Hi @ezhil56x(and others if you were wondering), I was confused why the update of `http_spec_e2e` was not a requirement. After a bit, I realized my local branch was behind and there was a change pushed few hours before I checked this. Sorry for the confusion, @ezhil56x. My review is not relevant anymore.

PS: It's cool to see things moving at "a few hours back" fast 🚀

tests/http/cache-control.yml
ezhil56x: @amitksingh1490
mapping values are not allowed. I don't know how to use two difference API calls here, it would be really helpful if you could assist me!
`query: query { user(id: "1") { name } }`
`query: query { user(id: "2") { name } }`

tests/http/cache-control.yml
ezhil56x: I will checkout in next commit 👍🏻

tests/http/cache-control.yml
amitksingh1490:
```
query: query { 
 u1: user(id: "1") { name }
 u2: user(id: "1") { name }
}
```

ezhil56x: > ```graphql
> query: query { 
>  u1: user(id: "1") { name }
>  u2: user(id: "1") { name }
> }
> ```

Still the test is getting failed when I use this query on both the requests in assert

I have also tried

> ```graphql
> query: query { 
>  u1: user(id: "1") { name }
> }
> ```

> ```graphql
> query: query { user(id: "1") { name } }
> ```

Caused by: mapping values are not allowed in this context

PR #661, opened 2023-11-19
GraphQL DataSource implementation

meskill:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/config/config.rs
```diff
@@ -319,6 +331,28 @@ pub struct Http {
   pub group_by: Vec<String>,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, Default)]
+pub struct GraphQLSource {
```
tusharmath, line 335:
```suggestion
pub struct GraphQL {
```

src/http/client.rs
```diff
@@ -65,7 +65,7 @@ impl DefaultHttpClient {
 
   pub async fn execute(&self, request: reqwest::Request) -> reqwest_middleware::Result<Response> {
     log::info!("{} {} ", request.method(), request.url());
-    let response = self.client.execute(request).await?;
+    let response = self.client.execute(request).await?.error_for_status()?;
```
tusharmath, line 68: Need a `http spec`  for this change.

meskill: I've added the http_spec for this but the problem is that http_spec has mocked http client that implements that functionality by its logic and therefore I have to add error handler in http mock as well.

we may require full e2e testing to add proper tests

meskill: added more sane test in server_spec

benches/impl_path_string_for_evaluation_context.rs
```diff
@@ -87,6 +88,12 @@ impl<'a> ResolverContextLike<'a> for MockGraphqlContext {
   fn args(&'a self) -> Option<&'a IndexMap<Name, Value>> {
     Some(&TEST_ARGS)
   }
+
+  fn field(&'a self) -> Option<SelectionField> {
```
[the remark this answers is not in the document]
meskill: yeah if we want to mock that method since we can't construct SelectionField on our side

src/path_string.rs
```diff
@@ -20,6 +24,14 @@ impl PathString for serde_json::Value {
   }
 }
 
+impl PathGraphql for serde_json::Value {
+  fn path_graphql<T: AsRef<str>>(&self, path: &[T]) -> Option<String> {
+    self
+      .get_path(path)
+      .and_then(|v| async_graphql::Value::from_json(v.clone()).ok().map(|v| v.to_string()))
+  }
+}
```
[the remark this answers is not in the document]
meskill: I've replied in the discord

src/http/data_loader_request.rs
```diff
@@ -19,6 +23,9 @@ impl DataLoaderRequest {
 impl Hash for DataLoaderRequest {
   fn hash<H: Hasher>(&self, state: &mut H) {
     self.0.url().hash(state);
+    if let Some(body) = self.0.body() {
+      body.as_bytes().hash(state);
+    }
```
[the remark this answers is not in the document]
meskill: @sujeetsr could you help with that?

I'm not quite sure why it's changed, but probably due to interest to dedupe and batch graphql requests that are represented as POST requests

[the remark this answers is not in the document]
sujeetsr: If body is not included in the hash, wouldn't two different graphql requests to the same url have the same hash?

src/path.rs
```diff
@@ -0,0 +1,260 @@
...
+      _ => None,
+    })
+  }
+}
+
+impl PathGraphql for serde_json::Value {
+  fn path_graphql<T: AsRef<str>>(&self, path: &[T]) -> Option<String> {
+    self
+      .get_path(path)
+      .and_then(|v| async_graphql::Value::from_json(v.clone()).ok().map(|v| v.to_string()))
```
tusharmath, line 44: As a separate note: This is going to be quite slow.

src/graphql_request_template.rs
```diff
@@ -0,0 +1,217 @@
...
+        args
+          .iter()
+          .map(|(k, v)| format!(r#"{}: {}"#, k, v.render_graphql(ctx).escape_default()))
+          .collect::<Vec<_>>()
+          .join(", ")
+      })
+      .map(|args| format!("{}({})", self.operation_name, args))
+      .unwrap_or(self.operation_name.clone());
+
+    let graphql_query = format!(r#"{{ "query": "{operation_type} {{ {operation} {selection_set} }}" }}"#);
```
tusharmath, line 82: Instead of hand-crafting a GraphQL query, we should use:

`async_graphql_parser::types::executable::ExecutableDocument`

src/http/data_loader_request.rs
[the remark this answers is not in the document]
tusharmath: Not saying that it wouldn't. Hashes will collide, but the bigger question is do we need the DL of POST requests?

tests/graphql_spec.rs
```diff
@@ -369,8 +362,8 @@ fn test_failures_in_client_sdl() -> std::io::Result<()> {
   Ok(())
 }
 
-#[test]
-fn test_merge_sdl() -> std::io::Result<()> {
+#[tokio::test]
+async fn test_merge_sdl() -> std::io::Result<()> {
```
tusharmath, line 366: What is the purpose of changes in this file?

src/blueprint/from_config/operators/graphql.rs
```diff
@@ -0,0 +1,39 @@
...
+  operation_type: &'a GraphQLOperationType,
+) -> TryFold<'a, (&'a Config, &'a Field, &'a config::Type, &'a str), FieldDefinition, String> {
+  TryFold::<(&Config, &Field, &config::Type, &'a str), FieldDefinition, String>::new(
+    |(config, field, _, _), b_field| {
+      let Some(graphql) = &field.graphql else {
+        return Valid::succeed(b_field);
+      };
+
+      let Some(base_url) = graphql.base_url.as_ref().or(config.upstream.base_url.as_ref()) else {
+        return Valid::fail("No base URL defined".to_string());
```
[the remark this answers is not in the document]
meskill: fixed here and for the http operator

tests/graphql_spec.rs
meskill: was needed in some intermediate changes, but now is obsolete, I've removed it

src/path.rs
meskill: I've removed this particular code as I mentioned in discord it doesn't bring much value

src/graphql_request_template.rs
meskill: There are some rough edges in using this:
- the main purpose for the struct is to parse query to the server it's little harder to convert it to string to pass further. yeah, there is a method to [stringify](https://docs.rs/async-graphql/latest/async_graphql/extensions/struct.ExtensionContext.html#method.stringify_execute_doc) it, by it's main purpose for logging and it could mess with the output to hide sensitive data for example
- in order to create ExecutableDocument we need to do different types of conversion into types that used in this type to at the end convert it back to string. It could be simpler and faster just convert it to string. And I think we can't do this upfront as selection_set of fields could be different. 
- also in current implementation we couldn't easily use variables here because to use it we need to know different arguments types to specify query like this `query(arg: String, arg2: Int) { user(id: $arg2) { name } }` and it's not a very straightforward. Previously, it was done by using upstream instrospection, but I removed it and started to pass arguments explicitly with `path_graphql`

Overall, I'm not sure it's worth the effort

src/http/data_loader_request.rs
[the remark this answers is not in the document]
meskill: we may want to do batching right?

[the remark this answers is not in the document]
sujeetsr: For http,  the dataloader is used only for GET requests. For graphql, we do want batching and dedupication of queries that have the same body

tests/graphql_mock/mod.rs
```diff
@@ -3,18 +3,24 @@ use std::fs;
 use regex::Regex;
 use serde_json::Value;
 
+#[must_use]
 pub fn start_mock_server() -> mockito::Server {
```
tusharmath, line 7: As discussed on discord, unless there is something specific that we can't test via http_spec, let's not add another way to write a test. It becomes quite confusing for contributors to write tests when there are multiple ways to do the same thing.

meskill: removed this and test that used it. 

put my thoughts on http_spec to https://github.com/tailcallhq/tailcall/issues/702

tusharmath (approved):

PR #662, opened 2023-11-19
Feature: Http 2

### LINKED ISSUE #605: Support HTTP2 on Tailcall WebServer
Currently tailcall only support HTTP 1.1. We need a configuration to enable HTTP 2 also. 

It should be configured via the `version` field inside of `@server`. By default it should be HTTP1.
```graphql
schema @server(version: HTTP2) {
  query: Query
}

type Query {
  user: User
}
```

ologbonowiwi:
**Summary:**  

**Issue Reference(s):**  
Fixes #605
/claim #605

**Build & Testing:**

**Checklist:**


examples/jsonplaceholder.graphql
```diff
@@ -1,5 +1,13 @@
...
-  @server(port: 8000, enableGraphiql: true, enableQueryValidation: false, hostname: "0.0.0.0")
+  @server(
+    port: 3000
+    enableGraphiql: true
+    enableQueryValidation: false
+    hostname: "0.0.0.0"
+    version: HTTP2
+    cert: "./examples/example.crt"
+    key: "./examples/example.key"
+  )
```
tusharmath, line 10: Can we have a separate example for HTTP2.

src/http/server.rs
```diff
@@ -66,30 +73,103 @@ fn create_allowed_headers(headers: &HeaderMap, allowed: &BTreeSet<String>) -> He
 
   new_headers
 }
+
+fn load_cert(filename: &str) -> Result<Vec<rustls::Certificate>, std::io::Error> {
```
tusharmath, line 77: Can we use the async implementation of tokio to read file asynchronously

src/http/server.rs
```diff
@@ -66,30 +73,103 @@ fn create_allowed_headers(headers: &HeaderMap, allowed: &BTreeSet<String>) -> He
...
+
+  let key = keys.into_iter().find(|key| {
+    matches!(
+      key,
+      rustls_pemfile::Item::RSAKey(_) | rustls_pemfile::Item::ECKey(_) | rustls_pemfile::Item::PKCS8Key(_)
+    )
+  });
+
+  if let Some(key) = key {
+    log::info!("🔑 Loaded private key");
```
tusharmath, line 104:
```suggestion
    log::debug!("🔑 Loaded private key");
```

src/http/server.rs
```diff
@@ -66,30 +73,103 @@ fn create_allowed_headers(headers: &HeaderMap, allowed: &BTreeSet<String>) -> He
...
+
+    Ok(match key {
+      rustls_pemfile::Item::RSAKey(key) => PrivateKey(key),
+      rustls_pemfile::Item::ECKey(key) => PrivateKey(key),
+      rustls_pemfile::Item::PKCS8Key(key) => PrivateKey(key),
+      _ => unreachable!(),
+    })
+  } else {
+    Err(CLIError::new("No private key found").into())
```
tusharmath, line 113: I think the error message needs to be different here. The key was actually found but doesn't match any of the above options.

ologbonowiwi: This method can also be used to read other stuff that is not private key (e.g., `X509Certificate` – a certificate – and `Crl` – a certificate revocation list –), so if did not matched the other condition the key was not found

src/http/server.rs
ologbonowiwi: Sure; I need to keep the `BufReader` from the `std::io` though, because `rustls_pemfile` needs a `std::io::BufRead` to read from

tusharmath:
@ologbonowiwi Can you look at the conflicts please.

tusharmath (approved):

PR #664, opened 2023-11-20
refactor: remove enable prefix

amitksingh1490:


src/config/server.rs
```diff
@@ -56,19 +56,19 @@ impl Server {
...
   }
   pub fn enable_introspection(&self) -> bool {
-    self.enable_introspection.unwrap_or(true)
+    self.introspection.unwrap_or(true)
   }
   pub fn enable_query_validation(&self) -> bool {
-    self.enable_query_validation.unwrap_or(true)
+    self.query_validation.unwrap_or(false)
   }
   pub fn enable_batch_requests(&self) -> bool {
```
tusharmath, line 70: All methods also need to be renamed.

PR #673, opened 2023-11-21
Added support for Human-Readable Flattened GraphQL Queries

### LINKED ISSUE #560: Support for Human-Readable Flattened GraphQL Queries
**Issue Description**:

To enhance the readability and potential CDN-friendliness of our GraphQL queries, I propose supporting a flattened, human-readable style of GraphQL queries sent as GET request parameters.

### Sample GraphQL Schema:

```graphql
type User {
  id: Int!
  name: String!
  age: Int!
  address: Address!
}

type Address {
  city: String!
  state: String!
}

type Mutation {
  updateUser(id: Int!, name: String): User!
}

type Query {
  users: [User]
}
```

### Sample GET Request for Query:

Using the sample schema, the traditional GraphQL request would look something like:

```graphql
{
  user(id: 123) {
    name
    age
    address {
      city
      state
    }
  }
}
```

With the flattened approach, a curl request would look like:

```bash
curl "https://example.com/api/user?id=123&select=name,age,address.city,address.state"
```
- The name of the field at query is moved into the route as `kebab-case`.
- Arguments such as `id` to the query is passed a query params.
- A special `select` query param is added to control what fields need to be queried for.

### Sample POST Request for Mutation:

For mutations, consider the following GraphQL mutation:

```graphql
mutation {
  updateUser(id: 123, name: "John Doe") {
    name
  }
}
```

In the flattened style, using a POST request:

```bash
curl -X POST "https://example.com/api/update-user?id=123&select=name&name=John%20Doe"
```
Similar to `query` in `mutation` 
- The name of the field at mutation is moved into the route as `kebab-case`.
- Arguments such as `id` to the query is passed a query params.
- A special `select` query param is added to control what fields need to be queried for.

### Comparison with Persisted Queries:

Persisted queries involve sending a generated ID instead of the entire GraphQL query to the server. This improves performance as smaller requests are sent over the network. However, a significant drawback of persisted queries is that they're not reversible. Once the client sends the persisted query ID, the server must have a mechanism to match the ID to the actual query. If there's a mismatch or if the query associated with the ID is lost, it can't be recreated from the ID alone.

In contrast, our proposed flattened query parameter approach retains human-readability. It makes the content of the request transparent, while still being concise. This human-readable format has potential advantages in debugging and understanding traffic logs, without having to look up persisted query IDs.

### Integration with CDN:

CDNs can be set up to cache GET requests based on URL. The flattened, human-readable style allows for effective caching of GraphQL queries at the CDN level. Since the queries are in the URL, CDNs can easily cache the response associated with each unique URL. This is an advantage over POST requests, which are typically not cached by CDNs, leading to potential performance benefits.

### Technical Requirements
- This approach should be controlled via a setting on `@server` 
- It should have 100% code coverage for tests.

### Considerations
- The `select` param can conflict with a param that the query needs. A possible solution could be to use a `$` prefix for select eg: `$select` or just `$`.
- Query Params could get very long because each field needs to be is specified. A possible solution could be to use `*` for eg: `$select=*` or `$select=*,address.*`. By default `address` might not be resolved because it's nested unless it's explicitly asked for.

---

I believe this approach can be a valuable addition, providing a balance between performance, readability, and CDN compatibility. Feedback and further suggestions are welcome.

ssddOnTop:
**Summary:**  
Added Support for Human-Readable Flattened GraphQL Queries.

Sample Usage: `curl "http://localhost:8000/api/posts?\$=id,user.name&id=1"`

**Issue Reference(s):**  
Fixes #560
/claim #560

**Build & Testing:**

**Checklist:**


src/http/server.rs
```diff
@@ -16,139 +16,184 @@ use crate::blueprint::Blueprint;
...
+        log::info!("🌍 Playground: http://{}", addr);
+    }
+
+    let r = if blueprint.server.enable_batch_requests {
+        builder.serve(make_svc_batch_req).await
+    } else {
+        builder.serve(make_svc_single_req).await
+    };
+
+    Ok(r.map_err(CLIError::from)?)
```
tusharmath, line 198: Add the API route handle in `http/handlers/rest.rs`

src/parser/de.rs
```diff
@@ -0,0 +1,303 @@
...
+mod de_tests {
+    use crate::async_graphql_hyper::GraphQLRequest;
+    use crate::parser::de::Parser;
+
+    #[test]
+    fn parse_t() {
+        let mut parser = Parser::from_qry("user&$m=id=123,address.city=foo&$s=name,age,address.city,address.state");
+        let x = parser.parse::<GraphQLRequest>();
+        println!("{:?}", x);
+    }
```
tusharmath, line 302: Let's use a built in query parser - should be available in hyper or reqwest I think.

src/http/server.rs
```diff
@@ -16,139 +16,184 @@ use crate::blueprint::Blueprint;
...
 }
```
tusharmath, line 199: Also add unit tests under `http_spec`

tusharmath:
@ssddOnTop the url should look like — 

**Expected**
```
curl "https://example.com/api/user?id=123&$=name,age,address.city,address.state"
```

**Actual**
```
curl "https://example.com/api?user&id=123&$=name,age,address.city,address.state"
```
ie. the operation name should be part of the route and not the query params.

ssddOnTop:
> the operation name should be part of the route and not the query params.

yes I will fix that

But apart from that I think keeping we should give key to the query part i.e. `$s` and `$r` part

I think I can remove that part but having some key will be much better to recognize

ssddOnTop:
@tusharmath can you assign someone to drop a review

tests/http/config/hello.graphql
```diff
@@ -0,0 +1,7 @@
+schema @server(port: 8080) {
+  query: Query
+}
+
+type Query {
+  hello: String! @const(data: "World!")
+}
```
tusharmath, line 7: do we need this file?

ssddOnTop: oh I accidentally committed that as well, I'll remove it

tests/http_spec.rs
```diff
@@ -302,3 +301,27 @@ async fn run(spec: HttpSpec, downstream_assertion: &&DownstreamAssertion) -> any
...
+  #[test]
+  fn t2_url_nested_qry_parser() {
+    let parser =
+      Parser::from_path("api/user?id=123,address.country=India,address.city=Foo&$=name,age,address.city,address.state");
+    assert_eq!(
+      parser.unwrap().parse::<GraphQLRequest>().unwrap().0.query,
+      "{user (id: 123,) {address (city: Foo,country: India,) ) {city state} age name}}"
+    );
+  }
+}
```
ologbonowiwi, line 327: These tests should be on the parser file itself

ologbonowiwi (changes requested):
Left a few comments, but we have a lot of "one letter variables" on parser file. We need to make it more declarative, giving explicit names since single char variables increases code complexity

examples/jsonplaceholder.graphql
```diff
@@ -22,5 +22,5 @@ type Post {
   userId: Int!
   title: String!
   body: String!
-  user: User @http(path: "/users/{{value.userId}}")
+  user: User @http(path: "/users/{{values.id}}")
```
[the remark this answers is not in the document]
ssddOnTop: yes I was just testing stuff, it was a blunder and I accidentally committed all the local changes which weren't supposed to, I'll fix that

examples/jsonplaceholder.graphql
```diff
@@ -22,5 +22,5 @@ type Post {
   userId: Int!
   title: String!
   body: String!
-  user: User @http(path: "/users/{{value.userId}}")
+  user: User @http(path: "/users/{{values.userId}}")
```
ologbonowiwi, line 25:
```suggestion
  user: User @http(path: "/users/{{value.userId}}")
```

src/http/request_handler.rs
```diff
@@ -80,7 +80,45 @@ pub async fn graphql_request<T: DeserializeOwned + GraphQLRequestLike>(
...
   }
 }
+async fn url_req<T: DeserializeOwned + GraphQLRequestLike>(
+  req: Request<Body>,
+  server_ctx: &ServerContext,
+) -> Result<Response<Body>> {
+  let req_ctx = Arc::new(create_request_context(&req, server_ctx));
+  if let Some(q) = req.uri().path_and_query() {
+    let q = q.as_str();
+    let mut parser = Parser::from_path(q)?;
```
ologbonowiwi, line 90:
```suggestion
  if let Some(query) = req.uri().path_and_query() {
    let query = query.as_str();
    let mut parser = Parser::from_path(query)?;
```

src/http/request_handler.rs
```diff
@@ -80,7 +80,45 @@ pub async fn graphql_request<T: DeserializeOwned + GraphQLRequestLike>(
...
+    let request = parser.parse::<T>();
+    match request {
+      Ok(request) => {
+        let response = request.data(req_ctx.clone()).execute(&server_ctx.schema).await;
+        let mut response = update_cache_control_header(response, server_ctx, req_ctx).to_response()?;
+        update_response_headers(&mut response, server_ctx);
+        Ok(response)
+      }
+      Err(err) => {
+        log::error!("Failed to parse request: {q}",);
```
ologbonowiwi, line 100: did not saw this one

```suggestion
        log::error!("Failed to parse request: {query}",);
```

src/parser/de.rs
ologbonowiwi, line 1: Since we only have a `de.rs` file on this module, we could make a `parser.rs` as part of `http`. what do you think?

ssddOnTop: I was thinking of that but I think it would be better to be part of `parser` module instead of `http` and will just keep parsing part in `de.rs`

src/http/request_handler.rs
```diff
@@ -80,7 +80,45 @@ pub async fn graphql_request<T: DeserializeOwned + GraphQLRequestLike>(
     }
   }
 }
+async fn url_req<T: DeserializeOwned + GraphQLRequestLike>(
```
tusharmath, line 83: create a new file `api_request_handler` and move this file there.

src/http/request_handler.rs
```diff
@@ -80,7 +80,45 @@ pub async fn graphql_request<T: DeserializeOwned + GraphQLRequestLike>(
...
+    log::error!("Failed to parse request, invalid url",);
+    let mut response = async_graphql::Response::default();
+    let server_error = ServerError::new("Unexpected GraphQL Request.".to_string(), None);
+    response.errors = vec![server_error];
+
+    Ok(GraphQLResponse::from(response).to_response()?)
+  }
+}
+
+async fn custom_url_req(req: Request<Body>, server_ctx: &ServerContext) -> Result<Response<Body>> {
```
tusharmath, line 119:
```suggestion
async fn api_request(req: Request<Body>, server_ctx: &ServerContext) -> Result<Response<Body>> {
```

src/parser/de.rs
```diff
@@ -0,0 +1,309 @@
...
+  #[test]
+  fn t2_url_nested_qry_parser() {
+    let parser =
+      Parser::from_path("api/user?id=123,address.country=India,address.city=Foo&$=name,age,address.city,address.state");
+    assert_eq!(
+      parser.unwrap().parse::<GraphQLRequest>().unwrap().0.query,
+      "{user (id: 123,) {address (city: Foo,country: India,) ) {city state} age name}}"
+    );
+  }
+}
```
tusharmath, line 309: Add support for `*` also which expands on all the fields, refer to the original issue for more details.

src/parser/de.rs
```diff
@@ -0,0 +1,309 @@
...
+use serde::de::{DeserializeOwned, Error};
+use serde_json::{Map, Value};
+
+use crate::async_graphql_hyper::GraphQLRequestLike;
+
+type PosValHolder = HashMap<String, Vec<(String, String)>>;
+
+#[derive(Debug, serde::Deserialize, Default)]
+pub struct Parser {
+  root: Option<String>,
```
tusharmath, line 12:
```suggestion
  operation: String,
```
Operation can't be optional

src/parser/de.rs
```diff
@@ -0,0 +1,309 @@
...
+  arguments: Option<String>,
```
tusharmath, line 13:
```suggestion
  arguments: HashMap<String, Vec<String>>,
```
Shouldn't arguments should be a hashmap.

src/parser/de.rs
tusharmath: I agree with @ologbonowiwi because this isn't going to be used anywhere else apart from `http`. For a first timer, it might feel like the parser module is for the .graphql files and configurations.

src/parser/de.rs
ssddOnTop: In the struct it just stores arguments string to later be parsed to hashmap check https://github.com/ssddOnTop/tailcall/blob/883118025e83f5ddca9ae4a9abc1ff5490414d13/src/parser/de.rs#L101

ssddOnTop: Storing the map in struct will just consume more memory when we know that we only need the funal `Value` and it's going to be parsed only once

src/http/request_handler.rs
```diff
@@ -80,7 +80,45 @@ pub async fn graphql_request<T: DeserializeOwned + GraphQLRequestLike>(
     }
   }
 }
+async fn url_req<T: DeserializeOwned + GraphQLRequestLike>(
+  req: Request<Body>,
+  server_ctx: &ServerContext,
+) -> Result<Response<Body>> {
+  let req_ctx = Arc::new(create_request_context(&req, server_ctx));
```
tusharmath, line 87: Create context only when required.

src/http/request_handler.rs
```diff
@@ -80,7 +80,45 @@ pub async fn graphql_request<T: DeserializeOwned + GraphQLRequestLike>(
...
+        Ok(response)
+      }
+      Err(err) => {
+        log::error!("Failed to parse request: {query}",);
+
+        let mut response = async_graphql::Response::default();
+        let server_error = ServerError::new(format!("Unexpected GraphQL Request: {}", err), None);
+        response.errors = vec![server_error];
 
+        Ok(GraphQLResponse::from(response).to_response()?)
```
tusharmath, line 106: There are different kinds of errors and since this is a RESTish API, we should stick to those standards. Eg: if the operation name is not provided a `404` should be issued. If the query params can't be parsed `400` should be issued etc.

tests/http/batching.yml
```diff
@@ -72,3 +72,19 @@ assert:
               - column: 1
                 line: 1
             message: " --> 1:1\n  |\n1 | FOO\n  | ^---\n  |\n  = expected executable_definition"
+  # Sending a bad request
+  - request:
+      method: GET
+      url: http://localhost:8080/api/posts\?\$\=title,user.name
```
meskill, line 78: does yaml requires escaping all that symbols?

even if so, I think it could be fixed by wrapping the whole string in quotes

tests/http/batching.yml
```diff
@@ -72,3 +72,19 @@ assert:
...
+    response:
+      status: 200
+      body:
+        data: {}
+  - request:
+      method: GET
+      url: http://localhost:8080/api/user \?id=1\&\$=id
```
meskill, line 85: with space in url? how should it work?

src/parser/de.rs
```diff
@@ -0,0 +1,95 @@
+use std::collections::HashMap;
+
+use serde_json::Value;
+type PosValHolder = HashMap<String, Vec<(String, String)>>;
+
+pub fn de_kebab(qry: &str) -> String {
```
meskill, line 6: can we use something like https://crates.io/crates/heck to do the case conversion?

src/http/request_handler.rs
```diff
@@ -24,14 +25,14 @@ fn not_found() -> Result<Response<Body>> {
   Ok(Response::builder().status(StatusCode::NOT_FOUND).body(Body::empty())?)
 }
 
-fn create_request_context(req: &Request<Body>, server_ctx: &ServerContext) -> RequestContext {
+pub fn create_request_context(req: &Request<Body>, server_ctx: &ServerContext) -> RequestContext {
```
meskill, line 28: as we have now different handlers and we have shared logic between shared handlers, it might be more suitable to have split it to 3 files: "api", "graphql", "shared". wdyt?

src/http/api_request_handler.rs
```diff
@@ -0,0 +1,61 @@
...
+    let response = Response::builder()
+      .status(StatusCode::BAD_REQUEST)
+      .body(Body::from(make_error_json(
+        "Failed to parse request, invalid url".to_string(),
+      )));
+    Ok(response?)
+  }
+}
+
+fn make_error_json(err: String) -> String {
```
meskill, line 53: you may check async_graphql::Error struct to recreate graphql error

src/http/parser.rs
```diff
@@ -0,0 +1,290 @@
...
+  operation: String,
+  arguments: Option<String>,
+  selections: Option<String>,
+}
+
+impl Parser {
+  pub fn from_path(path: &str) -> anyhow::Result<Parser> {
+    let path = urlencoding::decode(&de_kebab(path)).unwrap().replace('\\', "");
+    let mut root = String::new();
+    let split = path.split('?').collect::<Vec<&str>>();
```
meskill, line 21: better accept hyper::Uri instead of raw str that already has parsed uri and do not parse it yourself

src/http/parser.rs
```diff
@@ -0,0 +1,290 @@
...
+        parser.operation = root;
+        if let Some(q) = p.get("$") {
+          if let Some(q) = q.as_str() {
+            parser.selections = Some(q.to_string());
+          }
+        }
+        let arguments = p
+          .iter()
+          .filter(|(key, _)| !key.starts_with("api") && !key.starts_with("/api") && !key.starts_with('$'))
+          .map(|(k, v)| format!("{k}={}", v.as_str().unwrap()))
```
meskill, line 53: a lot of `unwrap` in this function. At least bubble the error with `?` operator

src/http/parser.rs
```diff
@@ -0,0 +1,290 @@
...
+      Ok(p) => {
+        let mut parser = Self::default();
+        let rootname = split.first().unwrap();
+        let mut to_camel = false;
+        for char in rootname.chars().skip(4) {
+          match char {
+            '/' => (),
+            '-' => {
+              to_camel = true;
+            }
```
meskill, line 33: is that de_kebab logic again?

src/http/parser.rs
```diff
@@ -0,0 +1,290 @@
...
+    if stk > 0 {
+      return Err(serde_json::Error::custom("Unexpected token {"));
+    }
+    Ok(s)
+  }
+}
+
+fn is_scalar_type(type_name: &str) -> bool {
+  ["String", "Int", "Float", "Boolean", "ID", "JSON"].contains(&type_name)
+}
```
meskill, line 244: I think that function already exists somewhere other place

src/http/parser.rs
```diff
@@ -0,0 +1,290 @@
+use std::collections::{HashMap, LinkedList};
```
meskill, line 1: too much is going on in that file. Can we try split it to modules and add unit tests?

tests/http/parser-extra.yml
```diff
@@ -0,0 +1,132 @@
+---
+name: Testing a GraphQL query
```
meskill, line 2: could you add test for a query with arguments?

src/parser/de.rs
```diff
@@ -0,0 +1,95 @@
...
+            if v.is_object() { k } else { root_node },
+            level + 1,
+          ),
+        );
+      }
+    }
+  }
+}
+
+pub fn to_json_str(value: &Value) -> String {
```
meskill, line 62: serde_json::Value already could be easily converted to json there is no need to specific function.

meskill (changes requested):

src/parser/de.rs
ssddOnTop: @meskill this will just increase compile time to add dependencies, we just need to convert it to lower camel case, and according to original issue, `foo-bar` should be converted to `fooBar`, but actually `-` is not recognized as separator, using heck we can only convert something like `foo_bar` to `fooBar` and `de_kebab` function also remove unwanted blank spaces to avoid further iterations

src/http/api_request_handler.rs
ssddOnTop: @meskill Tushar asked me to throw specific [http errors](https://github.com/tailcallhq/tailcall/pull/673#discussion_r1404735381)

src/http/parser.rs
ssddOnTop: ah yes I forgot that I already switched to a function, I'll remove it

src/parser/de.rs
ssddOnTop: @meskill this function name is misleading, sorry about that but this function is actually converting the `serde_json::Value` to queries/selections.

src/parser/de.rs
meskill: even simple things as conversion could have range of gotchas that you need to consider when you're going to implement this. And such libraries usually have automated tests and behaviour verified by users. 
Anyway, we already using another crate `convert_case` in the repo if you concern about the build time

src/http/api_request_handler.rs
meskill: here you're creating json object to represent error by Graphql spec - better use already existing types that handles it. It doesn't relate to http protocol

src/http/request_handler.rs
ssddOnTop: @meskill with shared you mean the common methods between `api` and `graphql` right?

meskill: yes

src/http/parser.rs
```diff
@@ -0,0 +1,159 @@
...
+
+#[derive(Debug, serde::Deserialize, Default)]
+pub struct Parser {
+  operation: String,
+  arguments: Option<String>,
+  selections: Option<String>,
+}
+
+impl Parser {
+  pub fn from_path(uri: &Uri) -> anyhow::Result<Parser> {
```
meskill, line 21:
```suggestion
  pub fn from_uri(uri: &Uri) -> anyhow::Result<Parser> {
```

meskill (changes requested):

src/async_graphql_hyper.rs
```diff
@@ -126,6 +126,23 @@ impl GraphQLResponse {
     Ok(response)
   }
 
+  pub fn to_response_with_status_code(self, status_code: StatusCode) -> Result<Response<hyper::Body>> {
```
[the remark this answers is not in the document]
ssddOnTop: @meskill how about migrating from the function above to the new function instead?

And I added that function because I was asked asked to do so [here](https://github.com/tailcallhq/tailcall/pull/673#discussion_r1404735381)

src/parser/de.rs
```diff
@@ -0,0 +1,219 @@
...
+use serde_json::{Map, Value};
+type PosValHolder = HashMap<String, Vec<(String, String)>>;
+
+pub fn parse_operation(path: &str) -> String {
+  if path.len() < 6 {
+    return String::new();
+  }
+  let path = de_kebab(path);
+  let mut s = String::new();
+  for char in path.chars().skip(5) {
```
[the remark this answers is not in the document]
ssddOnTop: it's known that the route starts with `/api/`so we need to ignore that part to extract the operation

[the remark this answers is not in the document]
ssddOnTop: so there must be at least 6 chars (operation can be of single char) for it to be valid query

src/http/parser.rs
```diff
@@ -0,0 +1,159 @@
...
+  de_kebab, next_token, parse_args, parse_arguments_string, parse_operation, parse_selections, parse_selections_string,
+  to_json, value_to_graphql_selections,
+};
+
+#[derive(Debug, serde::Deserialize, Default)]
+pub struct Parser {
+  operation: String,
+  arguments: Option<String>,
+  selections: Option<String>,
+}
```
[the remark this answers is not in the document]
ssddOnTop: The objective of this pr was to parse the queries to `GraphQLRequestLike`, there is no where in the existing code that we are interacting with `async_graphql::parser::ExecutableDocument` directly.

I understand there is too much to interpret here, as you suggested I'll try my best to split the code and try to add comments all over the code to explain functionality

src/parser/de.rs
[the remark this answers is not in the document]
ssddOnTop: I've tried my best to avoid unwraps, I have only kept unwraps where I am sure that it won't throw an exception.

src/parser/de.rs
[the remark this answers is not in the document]
meskill: either use some explicit prefix removal function or describe details why such constants are used

src/parser/de.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+pub fn de_kebab(qry: &str) -> String {
+  let qry = urlencoding::decode(qry).unwrap_or_default().replace(['\\', ' '], "");
+  let converter = convert_case::Converter::new();
+  let converter = converter.to_case(convert_case::Case::Camel);
+  converter.convert(qry)
+}
+
+pub fn to_json(value: &Value, result: &mut HashMap<usize, PosValHolder>, prl: (Option<String>, &String, usize)) {
```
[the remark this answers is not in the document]
ssddOnTop: I added some commends to the code, and rename the function to `deserialize_to_levelwise_args`, you may check that

tests/http/batching.yml
```diff
@@ -72,3 +72,20 @@ assert:
...
+      method: GET
+      url: http://localhost:8080/api/posts?$=title,user.name
+    response:
+      status: 200
+      body:
+        data: {}
+  # Bad formatting with spaces
+  - request:
+      method: GET
+      url: http://localhost:8080/api/user ?id=1&$=id
```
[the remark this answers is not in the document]
ssddOnTop: Well there's no harm in checking the spaces...

src/http/parser.rs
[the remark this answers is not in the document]
ssddOnTop: I researched about `ExecutableDocument` but as far as I understand, it's just internally used, and it won't make sense to de-serialize values to that struct and then again serialize it to a string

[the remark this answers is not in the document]
meskill: ok, thanks for checking this

tests/http/batching.yml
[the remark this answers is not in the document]
meskill: well, if you put space in any place in url for current tab, it won't work =)

It is not obvious why should we have any custom logic that in practice shouldn't be used and in case of using spaces they should be properly encoded for http

ologbonowiwi (changes requested):

PR #680, opened 2023-11-21
fix: logging on server start

amitksingh1490:


amitksingh1490:
@ologbonowiwi Please check this, Simplified the htt1 and htt2 start please check, as it did not need the rt.spawn

src/http/http_1.rs
```diff
@@ -8,39 +8,31 @@ use crate::cli::CLIError;
 
 pub async fn start_http_1(sc: Arc<ServerConfig>) -> std::prelude::v1::Result<(), anyhow::Error> {
   let addr = sc.addr();
-  Ok(
-    sc.tokio_runtime()?
-      .spawn(async move {
-        let sc_cloned = sc.clone();
+  let sc_cloned = sc.clone();
```
ologbonowiwi, line 11: since we're not moving `sc`, we don't need to clone it; we can use the original value

```suggestion
  let sc_cloned = sc.clone();
```

src/http/http_1.rs
```diff
@@ -8,39 +8,31 @@ use crate::cli::CLIError;
...
-        let make_svc_single_req = make_service_fn(|_conn| {
-          let state = Arc::clone(&sc_cloned);
-          async move {
-            Ok::<_, anyhow::Error>(service_fn(move |req| {
-              handle_single_request(req, state.server_context.clone())
-            }))
-          }
-        });
+  let make_svc_single_req = make_service_fn(|_conn| {
+    let state = Arc::clone(&sc_cloned);
```
ologbonowiwi, line 14:
```suggestion
    let state = Arc::clone(&sc);
```

src/http/http_1.rs
```diff
@@ -8,39 +8,31 @@ use crate::cli::CLIError;
...
-          builder.serve(make_svc_single_req).await
-        };
-
-        log_launch(sc.as_ref());
-        server.map_err(CLIError::from)
-      })
-      .await??,
-  )
+  let make_svc_batch_req = make_service_fn(|_conn| {
+    let state = Arc::clone(&sc_cloned);
```
ologbonowiwi, line 23:
```suggestion
    let state = Arc::clone(&sc);
```

src/http/http_1.rs
```diff
@@ -8,39 +8,31 @@ use crate::cli::CLIError;
...
+    async move {
+      Ok::<_, anyhow::Error>(service_fn(move |req| {
+        handle_batch_request(req, state.server_context.clone())
+      }))
+    }
+  });
+  let builder = hyper::Server::try_bind(&addr).map_err(CLIError::from)?;
+  log_launch(sc.as_ref());
+  let server: std::prelude::v1::Result<(), hyper::Error> = if sc_cloned.blueprint.server.enable_batch_requests {
```
ologbonowiwi, line 32:
```suggestion
  let server: std::prelude::v1::Result<(), hyper::Error> = if sc.blueprint.server.enable_batch_requests {
```

ologbonowiwi (approved):
left a feel comments but LGTM

src/http/http_1.rs
```diff
@@ -8,39 +8,29 @@ use crate::cli::CLIError;
...
+  let make_svc_batch_req = make_service_fn(|_conn| {
+    let state = Arc::clone(&sc);
+    async move {
+      Ok::<_, anyhow::Error>(service_fn(move |req| {
+        handle_batch_request(req, state.server_context.clone())
+      }))
+    }
+  });
+  let builder = hyper::Server::try_bind(&addr).map_err(CLIError::from)?;
+  log_launch(sc.as_ref());
```
ologbonowiwi, line 29: On `log_launch`, can we check for the used protocol also? The current implementation uses `https://{addr}` even for HTTP 1

amitksingh1490: Ya let me fix that

ologbonowiwi: image

The expected output here would be 
```
[2023-11-21T16:56:53Z INFO  tailcall::http] 🌍 Playground: http://0.0.0.0:8000
```
since we're in a non-HTTPS server

ologbonowiwi (approved):
LGTM

PR #683, opened 2023-11-22
Fix: Cache-Control header cacheability handling

### LINKED ISSUE #670: Cache Control Bug with no-store
The bug is pointed to in this PR

https://github.com/tailcallhq/tailcall/pull/669/files#diff-3b02f0ed8b516a325866c0d66fe5898cb708590994d2452158fdda5cbba46163R83

A-N-uraag:
**Summary:**  
Fixed public, private and no-cache directive handling in Cache-Control behavior

**Issue Reference(s):**  
Fixes #670 
/claim #670 

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/async_graphql_hyper.rs
+++ b/src/async_graphql_hyper.rs
@@ -129,27 +129,27 @@ impl GraphQLResponse {
   /// Sets the `cache_control` for a given `GraphQLResponse`.
   ///
   /// The function modifies the `GraphQLResponse` to set the `cache_control` `max_age`
-  /// to the specified `min_cache` value and `public` flag to `!cache_private`
+  /// to the specified `min_cache` value and `public` flag to `cache_public`
   ///
   /// # Arguments
   ///
   /// * `res` - The GraphQL response whose `cache_control` is to be set.
   /// * `min_cache` - The `max_age` value to be set for `cache_control`.
-  /// * `cache_private` - The negation of `public` flag to be set for `cache_control`.
+  /// * `cache_public` - The negation of `public` flag to be set for `cache_control`.
   ///
   /// # Returns
   ///
   /// * A modified `GraphQLResponse` with updated `cache_control` `max_age` and `public` flag.
-  pub fn set_cache_control(mut self, min_cache: i32, cache_private: bool) -> GraphQLResponse {
+  pub fn set_cache_control(mut self, min_cache: i32, cache_public: bool) -> GraphQLResponse {
     match self.0 {
       BatchResponse::Single(ref mut res) => {
         res.cache_control.max_age = min_cache;
-        res.cache_control.public = !cache_private;
+        res.cache_control.public = cache_public;
       }
       BatchResponse::Batch(ref mut list) => {
         for res in list {
           res.cache_control.max_age = min_cache;
-          res.cache_control.public = !cache_private;
+          res.cache_control.public = cache_public;
         }
       }
     };
--- a/src/async_graphql_hyper.rs
+++ b/src/async_graphql_hyper.rs
@@ -126,27 +126,30 @@ impl GraphQLResponse {
     Ok(response)
   }
 
-  /// Sets the `cache_control` max_age for a given `GraphQLResponse`.
+  /// Sets the `cache_control` for a given `GraphQLResponse`.
   ///
   /// The function modifies the `GraphQLResponse` to set the `cache_control` `max_age`
-  /// to the specified `min_cache` value.
+  /// to the specified `min_cache` value and `public` flag to `!cache_private`
   ///
   /// # Arguments
   ///
   /// * `res` - The GraphQL response whose `cache_control` is to be set.
   /// * `min_cache` - The `max_age` value to be set for `cache_control`.
+  /// * `cache_private` - The negation of `public` flag to be set for `cache_control`.
   ///
   /// # Returns
   ///
-  /// * A modified `GraphQLResponse` with updated `cache_control` `max_age`.
-  pub fn set_cache_control(mut self, min_cache: i32) -> GraphQLResponse {
+  /// * A modified `GraphQLResponse` with updated `cache_control` `max_age` and `public` flag.
+  pub fn set_cache_control(mut self, min_cache: i32, cache_private: bool) -> GraphQLResponse {
     match self.0 {
       BatchResponse::Single(ref mut res) => {
         res.cache_control.max_age = min_cache;
+        res.cache_control.public = !cache_private;
       }
       BatchResponse::Batch(ref mut list) => {
         for res in list {
           res.cache_control.max_age = min_cache;
+          res.cache_control.public = !cache_private;
         }
       }
     };
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -43,10 +43,12 @@ pub fn max_age(res: &Response) -> Option<Duration> {
 }
 
 pub fn cache_visibility(res: &Response) -> String {
-  match cache_policy(res) {
-    Some(value) if value.cachability == Some(Cachability::Public) => "public".to_string(),
-    Some(value) if value.cachability == Some(Cachability::Private) => "private".to_string(),
-    Some(value) if value.cachability == Some(Cachability::NoCache) => "no-cache".to_string(),
+  let cachability = cache_policy(res).and_then(|value| value.cachability);
+
+  match cachability {
+    Some(Cachability::Public) => "public".to_string(),
+    Some(Cachability::Private) => "private".to_string(),
+    Some(Cachability::NoCache) => "no-cache".to_string(),
     _ => "".to_string(),
   }
 }
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -14,7 +14,7 @@ mod server_context;
 
 use std::time::Duration;
 
-use cache_control::CacheControl;
+use cache_control::{Cachability, CacheControl};
 pub use client::*;
 pub use data_loader::*;
 pub use data_loader_request::*;
@@ -28,11 +28,27 @@ pub use server_context::ServerContext;
 
 use self::server_config::ServerConfig;
 
-pub fn max_age(res: &Response) -> Option<Duration> {
+pub fn cache_policy(res: &Response) -> Option<CacheControl> {
   let header = res.headers.get(CACHE_CONTROL)?;
   let value = header.to_str().ok()?;
-  let policy = CacheControl::from_value(value)?;
-  policy.max_age
+
+  CacheControl::from_value(value)
+}
+
+pub fn max_age(res: &Response) -> Option<Duration> {
+  match cache_policy(res) {
+    Some(value) => value.max_age,
+    None => None,
+  }
+}
+
+pub fn cache_visibility(res: &Response) -> String {
+  match cache_policy(res) {
+    Some(value) if value.cachability == Some(Cachability::Public) => "public".to_string(),
+    Some(value) if value.cachability == Some(Cachability::Private) => "private".to_string(),
+    Some(value) if value.cachability == Some(Cachability::NoCache) => "no-cache".to_string(),
+    _ => "".to_string(),
+  }
 }
 
 /// Returns the minimum TTL of the given responses.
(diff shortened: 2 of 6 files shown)


src/async_graphql_hyper.rs
```diff
@@ -126,27 +126,30 @@ impl GraphQLResponse {
...
   /// * `res` - The GraphQL response whose `cache_control` is to be set.
   /// * `min_cache` - The `max_age` value to be set for `cache_control`.
+  /// * `cache_private` - The negation of `public` flag to be set for `cache_control`.
   ///
   /// # Returns
   ///
-  /// * A modified `GraphQLResponse` with updated `cache_control` `max_age`.
-  pub fn set_cache_control(mut self, min_cache: i32) -> GraphQLResponse {
+  /// * A modified `GraphQLResponse` with updated `cache_control` `max_age` and `public` flag.
+  pub fn set_cache_control(mut self, min_cache: i32, cache_private: bool) -> GraphQLResponse {
```
ologbonowiwi, line 143: We could use `cache_public` here since we're doing `!cache_private` where the variable is used.

It'd make the core more concise and readable.

src/http/mod.rs
```diff
@@ -28,11 +28,27 @@ pub use server_context::ServerContext;
...
+    None => None,
+  }
+}
+
+pub fn cache_visibility(res: &Response) -> String {
+  match cache_policy(res) {
+    Some(value) if value.cachability == Some(Cachability::Public) => "public".to_string(),
+    Some(value) if value.cachability == Some(Cachability::Private) => "private".to_string(),
+    Some(value) if value.cachability == Some(Cachability::NoCache) => "no-cache".to_string(),
+    _ => "".to_string(),
```
ologbonowiwi, line 50:
```suggestion
  let cachability = cache_policy(res).and_then(|value| value.cachability);

  match  {
    Some(Cachability::Public) => "public".to_string(),
    Some(Cachability::Private) => "private".to_string(),
    Some(Cachability::NoCache) => "no-cache".to_string(),
    _ => "".to_string(),
}
```

src/http/request_context.rs
```diff
@@ -27,20 +29,34 @@ impl Default for RequestContext {
...
 
   pub async fn execute(&self, req: reqwest::Request) -> anyhow::Result<Response> {
     self.http_client.execute(req).await
   }
-  fn set_min_max_age_conc(&self, min_max_age: u64) {
+  fn set_min_max_age_conc(&self, min_max_age: i64) {
     *self.min_max_age.lock().unwrap() = Some(min_max_age);
   }
-  pub fn get_min_max_age(&self) -> Option<u64> {
+  pub fn get_min_max_age(&self) -> Option<i64> {
```
ologbonowiwi, line 48: why change it from unsigned? will we ever have negative values here?

src/http/request_context.rs
```diff
@@ -27,20 +29,34 @@ impl Default for RequestContext {
...
     *self.min_max_age.lock().unwrap()
   }
 
-  pub fn set_min_max_age(&self, max_age: u64) {
+  pub fn set_cache_private_true(&self) {
+    *self.cache_private.lock().unwrap() = Some(true);
+  }
+  pub fn is_cache_private(&self) -> Option<bool> {
```
ologbonowiwi, line 55:
```suggestion

  pub fn is_cache_private(&self) -> Option<bool> {
```

src/http/request_handler.rs
```diff
@@ -37,9 +37,9 @@ fn update_cache_control_header(
   req_ctx: Arc<RequestContext>,
 ) -> GraphQLResponse {
   if server_ctx.blueprint.server.enable_cache_control_header {
-    if let Some(ttl) = req_ctx.get_min_max_age() {
-      return response.set_cache_control(ttl as i32);
-    }
+    let ttl = req_ctx.get_min_max_age().unwrap_or(0);
+    let cache_private_flag = req_ctx.is_cache_private().unwrap_or(false);
+    return response.set_cache_control(ttl as i32, cache_private_flag);
```
ologbonowiwi, line 42:
```suggestion
    let ttl: i32 = req_ctx.get_min_max_age().unwrap_or(0);
    let cache_private_flag = req_ctx.is_cache_private().unwrap_or(false);
    return response.set_cache_control(ttl, cache_private_flag);
```

src/lambda/expression.rs
```diff
@@ -122,7 +123,13 @@ impl Expression {
                   .unwrap_or_default();
                 if ctx.req_ctx.server.get_enable_cache_control() && resp.status.is_success() {
                   if let Some(max_age) = max_age(&resp) {
-                    ctx.req_ctx.set_min_max_age(max_age.as_secs());
+                    ctx.req_ctx.set_min_max_age(max_age.as_secs() as i64);
+                  }
+                  if let Some(policy) = cache_policy(&resp) {
+                    ctx.req_ctx.set_cache_visibility(&policy.cachability);
+                    if Some(Cachability::NoCache) == policy.cachability {
+                      ctx.req_ctx.set_min_max_age(-1);
```
ologbonowiwi, line 131: why not use 0 here instead?

src/lambda/expression.rs
```diff
@@ -144,7 +151,13 @@ impl Expression {
...
               if ctx.req_ctx.server.get_enable_cache_control() && res.status.is_success() {
                 if let Some(max_age) = max_age(&res) {
-                  ctx.req_ctx.set_min_max_age(max_age.as_secs());
+                  ctx.req_ctx.set_min_max_age(max_age.as_secs() as i64);
+                }
+                if let Some(policy) = cache_policy(&res) {
+                  ctx.req_ctx.set_cache_visibility(&policy.cachability);
+                  if Some(Cachability::NoCache) == policy.cachability {
+                    ctx.req_ctx.set_min_max_age(-1);
+                  }
```
ologbonowiwi, line 160: looks like we have a duplicated code here, could you extract this to a helper function?

src/http/request_context.rs
A-N-uraag: Yes, @ologbonowiwi. In case of no-cache, it is set to -1.

src/lambda/expression.rs
A-N-uraag: async_graphql expects the max_age to be set to `-1` if there is `no-cache` directive to convert the `cache_control` attribute of `response` accordingly.

src/async_graphql_hyper.rs
A-N-uraag: Yes, I understand. Updated it.

src/lambda/expression.rs
A-N-uraag: Refactored to a new method `set_cache_control`.

src/http/request_handler.rs
A-N-uraag: This threw an error while running tests. So, I updated everything to just directly use `i32` since that's the type we use at the end.

tests/http/cache-control.yml
```diff
@@ -43,7 +43,7 @@ mock:
         test: test
     response:
       headers:
-        Cache-Control: "max-age=7200"
+        Cache-Control: "max-age=7200, private"
```
ologbonowiwi, line 46: We're lacking tests to public cache on this file

A-N-uraag:
Hi @b4s36t4, yeah, I already noted this and shared that with the team on discord. I realized it would've been better to keep it on the PR. 

> I started with fixing it for all the cases including no-store, proxy-revalidate, must-revalidate. I later learnt that async_graphql has support for just public, private and no-cache. So, I made updates to handle just those cases.

src/http/request_context.rs
```diff
@@ -93,4 +129,18 @@ mod test {
...
+    let req_ctx = RequestContext::default();
+    req_ctx.set_cache_visibility(&Some(Cachability::Private));
+    assert_eq!(req_ctx.is_cache_public(), Some(false));
+  }
+
+  #[test]
+  fn test_update_cache_visibility_public() {
+    let req_ctx = RequestContext::default();
+    req_ctx.set_cache_visibility(&Some(Cachability::Public));
+    assert_eq!(req_ctx.is_cache_public(), None);
```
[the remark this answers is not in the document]
A-N-uraag: Hi @ologbonowiwi , by default, the cache is assumed to be `public`. I wanted to avoid locking for updating the default field, so considered both `Some(true)` and `None` as `public`. Would that be a fair tradeoff to avoid unnecessary locking or do you think it's better to keep it restricted to just `Some(true)`?

tests/http/cache-control.yml
A-N-uraag: Oh, yes. Missed it. Updating...

A-N-uraag: Since the cache is by default `public`, async_graphql is not setting the `public` flag. This would still adhere to the standards. Do you want me to update the behavior to show the `public` directive if it's present?

src/http/request_context.rs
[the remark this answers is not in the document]
ologbonowiwi: imo we should have three statuses:

- public: `Some(true)`
- private: `Some(false)`
- no cache: `None`

but it'd be nice to hear @tusharmath and/or @amitksingh1490 opinion on this one

tests/http/cache-control.yml
amitksingh1490: @A-N-uraag what @ologbonowiwi meant was in this file there were two test
1. both cache control headers are public, expectation is to set the mimimum of two.
2. no-cache should in one request should convert the resultant header to no store.
you updated the first one to test private so there is not test which tests the both public scenario.

Expected tests
1. A - Public B - Public. result Min of A and B Public
2. A - Public B - Private result min of A and B Private
3. A - Public B - No cache result No cache
4. A - Private B - Nocache result No cache

src/http/request_context.rs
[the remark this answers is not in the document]
amitksingh1490: @A-N-uraag @ologbonowiwi , we could keep public as default to avoid locking.

tests/http/cache-control.yml
A-N-uraag: Ah, got it! Added the other scenarios. A small note @amitksingh1490: Case 4 would result in `No-Cache, Private` since No-Cache still indicates that a validation before re-use is required. But private should still be present to indicate that only private caches are allowed to store this response.

amitksingh1490 (approved):

src/http/mod.rs
```diff
@@ -28,11 +28,29 @@ pub use server_context::ServerContext;
...
+}
+
+pub fn max_age(res: &Response) -> Option<Duration> {
+  match cache_policy(res) {
+    Some(value) => value.max_age,
+    None => None,
+  }
+}
+
+pub fn cache_visibility(res: &Response) -> String {
```
meskill, line 45: Is this fn used anywhere other than tests?

src/http/request_context.rs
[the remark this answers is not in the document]
meskill: do we even have the situation when we set it to Some(true)? if no, it could become just a boolean flag. And in case you'll need 3 states consider using enum instead of Option(bool)

meskill (changes requested):

src/http/mod.rs
A-N-uraag: No, @meskill. Just the tests. Would you suggest moving it somewhere else?

src/http/request_context.rs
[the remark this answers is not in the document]
A-N-uraag: Hi @meskill, I tried to avoid locking at all in the first place and use a simple bool(since there's no "race condition" here technically). I am not very familiar with `rust` and this was resulting in a mutability issue which meant changing a lot of functions to use mutability. So, I ended up using something similar to `max_age`. If you have suggestions on improving this, I would love to work on them. Thanks.

src/http/mod.rs
meskill: yeah, you could put such things into the tests module if you need it only there

another thing to consider: do we need this at all? The Cachability struct could be compared directly without need to convert it to string

src/http/request_context.rs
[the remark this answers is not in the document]
meskill: that the mutex that provides internal mutability for you, not Option. So, when you have mutex you can change the value, and Option in that case in just excessive wrapper for value

PR #685, opened 2023-11-23
feat/support type of nested values

ologbonowiwi:
- chore: add nested_value example
- feat: get nested types
- refactor: use `get_nested_type` only when the tail is long

**Summary:**  
This PR aims to support nested values on queries.

On the current release, we have `{{value.userId}}`, but after this PR, we'll be able to do `{{value.user.id}}` also.

It fixes the bug mentioned on https://discord.com/channels/1044859667798568962/1176915094584688731/1177307810149695498

**Build & Testing:**

**Checklist:**


examples/nested_value.graphql
```diff
@@ -0,0 +1,41 @@
+schema @server(port: 8000, graphiql: true) @upstream(baseURL: "https://api-web.nhle.com") {
```
meskill, line 1: better put as test config to /tests

src/blueprint/from_config.rs
```diff
@@ -349,6 +350,18 @@ impl<'a> MustachePartsValidator<'a> {
...
+
+  fn get_nested_type(&self, tail: &[String], type_of: &config::Type) -> Option<Type> {
+    if tail.len() == 1 {
+      return get_value_type(type_of, &tail[0]);
+    }
+
+    let head = tail[0].as_str();
+    let field = type_of.fields.get(head)?;
+    let type_of = self.config.find_type(&field.type_of)?;
+    self.get_nested_type(&tail[1..], type_of)
```
meskill, line 362: take a look in [example](https://github.com/tailcallhq/tailcall/blob/741a1cc27e54e75132f840a68b4862fc73516bef/src/lambda/evaluation_context.rs#L125) for doing something like this without recursion

src/blueprint/from_config.rs
```diff
@@ -363,17 +376,28 @@ impl<'a> MustachePartsValidator<'a> {
...
-          // Queries can use optional values
-          if !is_query && val_type.is_nullable() {
-            return Valid::fail(format!("value '{tail}' is a nullable type"));
-          }
+        let val_type = if tail_is_long {
+          self.get_nested_type(&parts[1..], type_of)
         } else {
-          return Valid::fail(format!("no value '{tail}' found"));
+          get_value_type(type_of, tail)
+        };
```
meskill, line 385: just do it as a single function call without condition, just make sure get_nested_type handles both cases.

meskill (changes requested):

src/blueprint/from_config.rs
ssddOnTop: I fixed this

src/blueprint/from_config.rs
ssddOnTop: fixed

tusharmath:
@ologbonowiwi can you describe what is this PR for?

ologbonowiwi:
Fix the bug mentioned on https://discord.com/channels/1044859667798568962/1176915094584688731/1177307810149695498

tusharmath:
@ologbonowiwi can you update the PR description with the problem fixed.

Need to add tests for the fix.

amitksingh1490:
@ologbonowiwi Please rebase this once.

examples/nested_value.graphql
ologbonowiwi: done

ologbonowiwi:
@meskill @tusharmath @amitksingh1490, we've implemented all requested changes/tweaks. I've also added the description on the PR as requested.

Can you review this again?

src/blueprint/from_config/operators/http.rs
```diff
@@ -30,6 +30,32 @@ impl<'a> MustachePartsValidator<'a> {
...
+          _ => {
+            if val_type.is_nullable() {
+              return Err(format!("value '{}' is a nullable type", item.as_str()));
+            }
+          }
+        }
+      } else {
+        return Err(format!("no value '{}' found", item));
+      }
+      let field = type_of.fields.get(item).ok_or(format!("no value '{}' found", item))?;
```
meskill, line 50: `ok_or` part is basically already checked in the `get_value_type`. And as `get_value_type` is used only in this function we can change it or remove to simplify that function and to not to duplicate the logic

src/blueprint/from_config/operators/http.rs
```diff
@@ -30,6 +30,32 @@ impl<'a> MustachePartsValidator<'a> {
...
+      if let Some(val_type) = get_value_type(type_of, item) {
+        match len {
+          1 => {
+            return Ok(val_type);
+          }
+          _ => {
+            if val_type.is_nullable() {
+              return Err(format!("value '{}' is a nullable type", item.as_str()));
+            }
+          }
```
meskill, line 45: if-else wouldn't be simpler here?

src/blueprint/from_config/operators/http.rs
```diff
@@ -30,6 +30,32 @@ impl<'a> MustachePartsValidator<'a> {
...
+        return Err(format!("no value '{}' found", item));
+      }
+      let field = type_of.fields.get(item).ok_or(format!("no value '{}' found", item))?;
+      type_of = self
+        .config
+        .find_type(&field.type_of)
+        .ok_or(format!("no value '{}' found", item))?;
+      len -= 1;
+    }
+    unreachable!()
```
meskill, line 57: I think the code could be written without that:
- to resolve the value you need to pass the whole tail array actually, so if we store current resolved value in some mutable variable outside the loop, after the loop is finished we can just return that value
- if while iterating we can't resolve some value just return Err

tests/graphql/test-nested-value.graphql
```diff
@@ -0,0 +1,75 @@
+#> server-sdl
```
meskill, line 1: isn't it too many types to do the testing? it could be simplified I think.

ologbonowiwi: agree! done :smile:

src/blueprint/from_config/operators/http.rs
ologbonowiwi: agree; done!

src/blueprint/from_config/operators/http.rs
ologbonowiwi: I returned an error instead; imo we can keep the return inside the loop as it is instead of create a variable outside of it

src/blueprint/from_config/operators/http.rs
ologbonowiwi: I've removed the `ok_or`, but I'll let `get_value_type` as it is to avoid more nested conditions. if you prefer to leave this logic on this method, let me know, and I can delete `get_value_type`.

src/blueprint/from_config/operators/http.rs
```diff
@@ -30,6 +30,28 @@ impl<'a> MustachePartsValidator<'a> {
...
     Self { type_of, config, field }
   }
+  fn get_nested_type(&self, tail: &[String], mut type_of: &'a config::Type) -> Result<(Type, String), String> {
+    let mut len = tail.len();
+    for item in tail {
+      let val_type = get_value_type(type_of, item).ok_or_else(|| format!("no value '{}' found", item))?;
+
+      if len == 1 {
+        return Ok((val_type, item.to_owned()));
+      } else if val_type.is_nullable() {
```
meskill, line 40: in the last step in `validate` function we check for nullable only for !is_query. Should it be checked the same way here?

src/blueprint/from_config/operators/http.rs
```diff
@@ -30,6 +30,28 @@ impl<'a> MustachePartsValidator<'a> {
...
+      if let Some(field) = type_of.fields.get(item) {
+        if let Some(_type_of) = self.config.find_type(&field.type_of) {
+          type_of = _type_of;
+        }
+      }
+
+      len -= 1;
+    }
+
+    Err(format!("no value '{}' found", tail.join(".")))
```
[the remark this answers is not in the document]
meskill: ok, if we can prove it, we can mark it as unreachable and leave a comment why it is

src/blueprint/from_config/operators/http.rs
```diff
@@ -30,6 +30,28 @@ impl<'a> MustachePartsValidator<'a> {
...
+        return Ok((val_type, item.to_owned()));
+      } else if val_type.is_nullable() {
+        return Err(format!("value '{}' is a nullable type", item.as_str()));
+      }
+
+      if let Some(field) = type_of.fields.get(item) {
+        if let Some(_type_of) = self.config.find_type(&field.type_of) {
+          type_of = _type_of;
+        }
+      }
```
meskill, line 48: this is not obvious actually that `type_of` will be updated although in practice it should. Because the first condition is checked already in `get_value_type` and the second should be validated prior to this validation. 

Let's actually handle the case when `type_of` is not updated and return error to prevent issues in the future in case some changes will be made to other code that will break that assumptions

src/blueprint/from_config/operators/http.rs
ologbonowiwi: done

src/blueprint/from_config/operators/http.rs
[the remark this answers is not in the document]
ologbonowiwi: I've removed the `ok_or_else` now the code is reachable.

src/blueprint/from_config/operators/http.rs
ologbonowiwi: I've deleted `get_value_type` and now only check once for `type_of.fields.get`. I'm using `ok_or_else` to throw a `Path [{}] does not exist`, but the code never comes to this inexistent path, as not found will not even enter on the if.

If there are any more needed changes, let me know. We should specify a test case to cover it, but I could not reach that code here. It'd be nice get a hand on it.

meskill: is is a `Path` or more like `Type` in that case?

I think if you'll try to replicate this the validation will fail with another error since it's will be validated out by another code. I think we don't need to care about it here, so it's ok to not have reproduction in tests for this, but our code should be reliable anyway and we should the handle all the cases

src/blueprint/from_config/operators/http.rs
```diff
@@ -19,19 +19,36 @@ struct MustachePartsValidator<'a> {
...
+          return Ok((val_type, item.to_owned()));
+        } else if !is_query && val_type.is_nullable() {
+          return Err(format!("value '{}' is a nullable type", item.as_str()));
+        }
+
+        type_of = self
+          .config
+          .find_type(&field.type_of)
+          // this code never runs because we validate the config before
+          .ok_or_else(|| format!("Path [{}] does not exist", field.type_of.as_str()))?;
```
ologbonowiwi, line 43:
```suggestion
          .ok_or_else(|| format!("Type [{}] does not exist", field.type_of.as_str()))?;
```

src/blueprint/from_config/operators/http.rs
ologbonowiwi: Changed it to `Type`

ologbonowiwi:
@meskill, we already have a test that checks for not found: [test schema](https://github.com/tailcallhq/tailcall/pull/685/files#diff-77fd1a9f78c8c2224775b798b6b29355231d003314f88a8b083b3c8f4ac4b589R36-R37), [test assertion](https://github.com/tailcallhq/tailcall/pull/685/files#diff-77fd1a9f78c8c2224775b798b6b29355231d003314f88a8b083b3c8f4ac4b589R55-R59).

But this case will be handled by the `no value found` case, [here](https://github.com/tailcallhq/tailcall/pull/685/files#diff-ad4b6cd1504ee254a5e6f1998d82cf02edfb9ba256af7187ca231fa9cd47f944R49).

As you said is not required to find tests to this we're good to merge it

meskill:
> @meskill, we already have a test that checks for not found: [test schema](https://github.com/tailcallhq/tailcall/pull/685/files#diff-77fd1a9f78c8c2224775b798b6b29355231d003314f88a8b083b3c8f4ac4b589R36-R37), [test assertion](https://github.com/tailcallhq/tailcall/pull/685/files#diff-77fd1a9f78c8c2224775b798b6b29355231d003314f88a8b083b3c8f4ac4b589R55-R59).
> 
> But this case will be handled by the `no value found` case, [here](https://github.com/tailcallhq/tailcall/pull/685/files#diff-ad4b6cd1504ee254a5e6f1998d82cf02edfb9ba256af7187ca231fa9cd47f944R49).
> 
> As you said is not required to find tests to this we're good to merge it

but have you tried it?

ologbonowiwi:
> but have you tried it?

now I've tried, I see the error.

I've added it as a test case + added a few more cases that were covered after. If you find anything else, let me know :smile:

ologbonowiwi:
> but have you tried it?

now I've tried, I see the error.

I've added it as a test case + added a few more cases that were covered after. If you find anything else, let me know @meskill :smile:

src/blueprint/from_config/operators/http.rs
```diff
@@ -19,19 +19,43 @@ struct MustachePartsValidator<'a> {
...
+        // get tail so far to be logged
+        let item = tail[0..tail.len() - len + 1].join(".");
+
+        return Err(format!("no value '{}' found", item.as_str()));
+      }
+    }
+
+    let item = tail[0..tail.len() - len + 1].join(".");
+
+    Err(format!("no value '{}' found", item.as_str()))
```
ologbonowiwi, line 55: as we can see on the codecov report, this code is unreachable (we have a lot of tests without value, but all of them are being handled on the loop above), but as recommended, I'm keeping the error return, just in case

meskill (changes requested):

src/blueprint/from_config/operators/http.rs
```diff
@@ -44,17 +65,21 @@ impl<'a> MustachePartsValidator<'a> {
...
+        match val_type {
+          Ok((val_type, tail)) => {
+            if !is_scalar(val_type.name()) {
+              return Valid::fail(format!("value '{tail}' is not of a scalar type"));
+            }
+
+            // Queries can use optional values
+            if !is_query && val_type.is_nullable() {
+              return Valid::fail(format!("value '{tail}' is a nullable type"));
+            }
```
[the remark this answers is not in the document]
ologbonowiwi: This made me think that we should do this here instead on the `get_nested_type` (for being explicit, and also to keep only one responsibility on `get_nested_type`).

src/blueprint/from_config/operators/http.rs
ologbonowiwi: this did not work as the variable would start uninitialized :confused:

do you have any other suggestions?

ologbonowiwi: I've found a solution, I'm resolving this one

ologbonowiwi:
@meskill all changes addressed and 100% of code coverage, feel free to review it again when available :smile:

If you have any other test suggestions to stress the implementation, let me know and I'll add it 😄

meskill (approved):

PR #693, opened 2023-11-26
Enhencement/recursive check resolver

### LINKED ISSUE #564: Allow Empty Wrappers
Empty wrapping of resolvers should be allowed. However it should fail if no resolver is found anywhere on the to it's leaf nodes. 

**Config**
```graphql
schema @server(port: 8000) @upstream(baseURL: "http://jsonplaceholder.typicode.com") {
  query: Query
}

type Query {
  posts: InPost
}

type InPost {
  get: [Post] @http(path: "/posts")
}
```
**Actual**
```bash
Error: Invalid Configuration
Caused by:
  • No resolver has been found in the schema [at Query.posts]
```

**Expected**
Server should be able to start successfully. 

---
The following should continue to throw errors because no resolver is found to resolve Post

**Config**
```graphql
schema @server(port: 8000) @upstream(baseURL: "http://jsonplaceholder.typicode.com") {
  query: Query
}

type Query {
  posts: InPost
}

type InPost {
  get: [Post]
}

type Post {
  id: Int!
  userId: Int!
  title: String!
  body: String!
}

```

**Expected Error**
```
Error: Invalid Configuration
Caused by:
  • No way to resolve fields in Post from Query ❯ InPost ❯ Post [at Post]
```

Shylock-Hg:
**Summary:**  
Check resolver in recursive.
**Issue Reference(s):**  
Close #564
/claim #564
**Build & Testing:**

**Checklist:**


amitksingh1490:
Add a test for execution also for details check this folder
`tailcall/tests/graphql/passed`

src/blueprint/from_config/schema.rs
```diff
@@ -50,9 +70,11 @@ fn validate_mutation(config: &Config) -> Valid<(), String> {
...
     };
 
-    Valid::from_iter(mutation.fields.iter(), validate_field_has_resolver)
-      .trace(mutation_type_name)
-      .unit()
+    Valid::from_iter(mutation.fields.iter(), |i| {
+      validate_field_has_resolver(i, &config.types)
+    })
+    .trace(mutation_type_name)
+    .unit()
```
meskill, line 77: it could be replaced like this here? 
```suggestion
    validate_field_of_type_has_resolver(mutation_type_name, mutation, &config.types)
```
and for query as well

src/blueprint/from_config/schema.rs
```diff
@@ -14,14 +14,34 @@ fn validate_query(config: &Config) -> Valid<(), String> {
...
+        if let Some(ty2) = types.get(f_type) {
+          let res = validate_field_of_type_has_resolver(f_type, ty2, types);
+          return !res.is_succeed();
+        } else {
+          return true;
+        }
+      }
+      {
+        false
+      }
```
meskill, line 43:
```suggestion
    false
```

meskill (changes requested):

src/blueprint/from_config/definitions.rs
```diff
@@ -293,13 +293,28 @@ fn validate_field_type_exist(config: &Config, field: &Field) -> Valid<(), String
...
+    update_args()
+      .and(update_http().trace("@http"))
+      .and(update_unsafe().trace("@unsafe"))
+      .and(update_const_field().trace("@const"))
+      .and(update_modify().trace("@modify"))
+      .and(update_ref_field())
+      .try_fold(&(config, field, type_of, name), FieldDefinition::default())
+  };
+
+  let add_to_field = |name: &String, field: &Field| {
```
meskill, line 312: this closure is literally the same closure as to_field from above. Could we have just one closure with maybe additional argument to do this?

src/lambda/expression.rs
```diff
@@ -168,6 +172,20 @@ impl Expression {
             }
           }
         }
+        Self::Merge(subs) => {
```
meskill, line 175: could you provide an example of problem you're solving with this? I'm not sure this is part of the original issue

meskill (changes requested):

src/blueprint/from_config/operators/ref_field.rs
```diff
@@ -0,0 +1,40 @@
...
+        let resolvers = to_fields(ty, config).map(|i| {
+          i.iter()
+            .filter(|fd| fd.resolver.is_some())
+            .map(|fd| (fd.name.clone(), fd.resolver.clone()))
+            .collect::<IndexMap<String, Option<Expression>>>()
+        });
+        let mut mut_field = b_field;
+        resolvers.map(|i| {
+          if !i.is_empty() {
+            mut_field.resolver = Some(Merge(i));
```
meskill, line 27: if we replace it like this

```
mut_field.resolver = Some(Expression::Literal(serde_json::Value::Object(Default::default())));
```

the tests still pass. This is due how async_graphql handles inner fields, so the Merge could be removed. wdyt?

src/blueprint/from_config/operators/ref_field.rs
```diff
@@ -0,0 +1,35 @@
...
+  TryFold::<(&Config, &Field, &config::Type, &str), FieldDefinition, String>::new(
+    move |(config, field, _, _), b_field| {
+      if is_add_field {
+        return Valid::succeed(b_field);
+      }
+      if !field.has_resolver() {
+        let field_type = config.types.get(&field.type_of);
+        if let Some(ty) = field_type {
+          let mut mut_field = b_field;
+          let has_resolver = ty.fields.iter().any(|f| f.1.has_resolver());
```
meskill, line 20: I think here could be a problem since we are not checking fields recursively here. Can we add another test for resolving value with more deep nesting? similar to the `tests/graphql/passed/test-ref-other.graphql` to verify it?

Something like this:
```
type User {
  name: String @http(path: "/users/1")
}

type User1 {
  user1: User
}

type Query {
  firstUser: User1
}
```

meskill (changes requested):

src/blueprint/from_config/operators/ref_field.rs
```diff
@@ -0,0 +1,48 @@
+use std::collections::BTreeMap;
+
+use crate::blueprint::*;
+use crate::config;
+use crate::config::{Config, Field};
+use crate::lambda::Expression;
+use crate::try_fold::TryFold;
+use crate::valid::Valid;
+
+fn fields_has_resolver(ty: &config::Type, types: &BTreeMap<String, config::Type>) -> bool {
```
meskill, line 10: isn't this almost the save function as `validate_field_of_type_has_resolver` below? at least they have the same purpose to check nested resolvers.

So, I'm thinking since `update_ref_field` returns `Valid` we can have that logic for validation and setting resolver completely in it and there is no need to do the same inside `src/blueprint/from_config/schema.rs`

src/blueprint/from_config/operators/ref_field.rs
meskill: right, thanks

src/blueprint/from_config/operators/ref_field.rs
meskill: we don't have specific separation for updating and validation as it usually easier to make both together since they usually involve the same kind of operations to do. Take a look at other fields like `src/blueprint/from_config/operators/modify.rs` to confirm this.

meskill (approved):

PR #694, opened 2023-11-26
test: add server start

### LINKED ISSUE #687: Add a test for server start
Write a test that does the following

- Starts a GraphQL server on port 8000.
- Make 100 requests to the server.
- Assert the shape of the response received.
- Kill the server and end the test.

**Technical Requirements**
- Use the following GraphQL Config
  ```graphql
  schema @server(port: 8000) {
    query: Query
  }
  
  type Query {
    greet: String @const(data: "Hello World!")
  }
  ```
- Test should run under 100ms.

neo773:
Adds test for server start with the spec below:

**Issue Reference(s):**  
Fixes #687

**Build & Testing:**

image

/claim #687


tests/server_spec.rs
```diff
@@ -0,0 +1,36 @@
...
+  use serde_json::json;
+
+  tokio::spawn(async {
+    let file_paths = vec!["tests/server/config/server-start.graphql".to_string()];
+    let config = Config::from_file_or_url(file_paths.iter()).await.unwrap();
+    start_server(config).await.unwrap();
+  });
+
+  // Give the server some time to start
+  tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
```
ologbonowiwi, line 16: If we don't spawn a new thread to the server we can await for tokio start instead of sleeping (which slow down automated tests)

neo773: Does not work it gets stuck infinitely when awaiting directly

ologbonowiwi, line 16: something like

```suggestion
  let file_paths = vec!["tests/server/config/server-start.graphql".to_string()];
  let config = Config::from_file_or_url(file_paths.iter()).await.unwrap();
  start_server(config).await.unwrap();
```

ologbonowiwi: We don't have a way to await the spawn then? What bothers me is having to sleep for 5 secs, this does not match the issue requirements of running under 100ms

ologbonowiwi: maybe this works?

```suggestion
  tokio::spawn(async {
    let file_paths = vec!["tests/server/config/server-start.graphql".to_string()];
    let config = Config::from_file_or_url(file_paths.iter()).await.unwrap();
    start_server(config).await.unwrap();
  })
  .await;
```

neo773: Fixed it, needed to remove await from server.
Averaging at around 40 ms now
Thanks

neo773: Nevermind that was flaky, I used a different approach

tests/server_spec.rs
```diff
@@ -0,0 +1,38 @@
...
+async fn server_start() {
+  use reqwest;
+  use reqwest::Client;
+  use serde_json::json;
+
+  tokio::spawn(async {
+    let file_paths = vec!["tests/server/config/server-start.graphql".to_string()];
+    let config = Config::from_file_or_url(file_paths.iter()).await.unwrap();
+    start_server(config).await.unwrap();
+  });
```
ologbonowiwi, line 14: According to [tokio docs](https://tokio.rs/tokio/tutorial/spawning#tasks), we can 
```suggestion
  tokio::spawn(async {
    let file_paths = vec!["tests/server/config/server-start.graphql".to_string()];
    let config = Config::from_file_or_url(file_paths.iter()).await.unwrap();
    start_server(config).await.unwrap()
  }).await.unwrap();
```

In this way we don't need to loop until the server starts

neo773: Like I mentioned before this results in infinite loading

ologbonowiwi: @meskill @tusharmath @amitksingh1490, guys, can you think about an alternative here?

I've also tried to save the spawn in a variable and then loop until `spawn.is_finished()` became true, but this does not work either.

neo773:
@amitksingh1490 
Pushed the changes, it's down to 10ms now

image

tests/server_spec.rs
tusharmath: I would recommend using a `oneshot::channel()` to perform this synchronization. `start_server` should take a `oneshot` and set the value once the handler is registered. In the test we should await on the reader.

tests/server_spec.rs
```diff
@@ -0,0 +1,37 @@
...
+
+  let mut tasks = vec![];
+  for _ in 0..100 {
+    let client = client.clone();
+    let query = query.clone();
+    let task = tokio::spawn(async move {
+      let send_request = || async { client.post("http://localhost:8000/graphql").json(&query).send().await };
+      let response = send_request().await.expect("Failed to send request");
+      let response_body: serde_json::Value = response.json().await.expect("Failed to parse response body");
+      assert_eq!(response_body["data"]["greet"], "Hello World!");
```
tusharmath, line 33:
```suggestion
      assert_eq!(response_body, json!({data: {greet: "Hello World!"}});
```
This might work

meskill (changes requested):

meskill:
codecov is unhappy with coverage for http2. Could we have test for http2 server as well? it should look the same just with different server config

src/http/http_1.rs
```diff
@@ -27,6 +31,7 @@ pub async fn start_http_1(sc: Arc<ServerConfig>) -> std::prelude::v1::Result<(),
     }
   });
   let builder = hyper::Server::try_bind(&addr).map_err(CLIError::from)?;
+  tx.send(true).ok();
```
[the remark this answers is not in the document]
meskill: yes, this will be useful, but do you think it should be in scope of this pr?

neo773:
@meskill 
Added the changes including http2 coverage
server signals seems to be working well but cargo test is way too fast and there isn't enough time for server to shutdown before cargo starts running the next test.
Not sure what do to here kinda lost at this.

meskill:
> @meskill Added the changes including http2 coverage server signals seems to be working well but cargo test is way too fast and there isn't enough time for server to shutdown before cargo starts running the next test. Not sure what do to here kinda lost at this.

Great!
Actually, the problem not with how fast the tests are executed but in fact that they run in parallel and both of your tests are trying to start the server on the same port. If you specify different ports it'll be fine

meskill (approved):

PR #704, opened 2023-11-29
standardises the check/start commands to use the same functions

alexanderjophus:
**Summary:**  
Following on from the other simplification PR #701 , this has check and start go through the same function calls so there's a smaller chance of split when refactoring changes in the future.

**Issue Reference(s):**  
None, just minor cleanup

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/cli/tc.rs
+++ b/src/cli/tc.rs
@@ -11,6 +11,7 @@ use tokio::runtime::Builder;
 use super::command::{Cli, Command};
 use crate::blueprint::Blueprint;
 use crate::cli::fmt::Fmt;
+use crate::cli::CLIError;
 use crate::config::Config;
 use crate::http::start_server;
 use crate::print_schema;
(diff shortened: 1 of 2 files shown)


src/cli/tc.rs
```diff
@@ -23,35 +24,37 @@ pub fn run() -> Result<()> {
...
     }
+    Command::Check { file_path, n_plus_one_queries, schema } => match parse_blueprint(file_path, n_plus_one_queries) {
+      Ok(blueprint) => {
+        if schema {
+          display_schema(&blueprint);
+        }
+        Ok(())
+      }
+      Err(e) => Err(e),
+    },
```
meskill, line 47: Consider simplification like this

```suggestion
    Command::Check { file_path, n_plus_one_queries, schema } => 
    parse_blueprint(file_path, n_plus_one_queries)
    .map(|blueprint| {
        if schema {
          display_schema(&blueprint);
        }
    }),
```

and it could be applied to `Command::Start` also

src/cli/tc.rs
```diff
@@ -23,35 +24,37 @@ pub fn run() -> Result<()> {
...
     Command::Init { file_path } => Ok(tokio::runtime::Runtime::new()?.block_on(async { init(&file_path).await })?),
   }
 }
 
+fn parse_blueprint(file_path: Vec<String>, n_plus_one_queries: bool) -> Result<Blueprint> {
+  let config = tokio::runtime::Runtime::new()?.block_on(async { Config::from_file_or_url(file_path.iter()).await })?;
+  display_config(&config, n_plus_one_queries);
```
meskill, line 54: will show that info for any command. Should it be like this?

alexanderjophus: I was hoping to chat about that bit actually, it was one of the inconsistencies that made me open this PR.
I'll have another look at it tomorrow as on second look it feels weird logging something out there anyway. In fact this change might still need some work

src/cli/tc.rs
alexanderjophus: Does this not mask/hide any errors that may occur?

meskill: map will only be called if there is no error and if there is an error it will be returned

src/cli/tc.rs
meskill: I think it's ok the either way - that log by default is pretty small and not annoying

PR #706, opened 2023-11-30
feat: support operation validation via `check` command

### LINKED ISSUE #416: Pass `operations` to `check` command

tobihans:
**Summary:**  
This PR add ability to check that the schema is able to execute some operations. It follows the second approach outlined in the [issue](https://github.com/tailcallhq/tailcall/issues/416#issue-1925530012).

It's worth noting that the validation depends on the `queryValidation` parameter of the `@server` directive, as the query is executed against the schema (with resolvers that don't actually do anything).

Given a the sample schema `jsonplaceholder.graphql` and two operations `posts.graphql` and `empty.graphql`, we get the output below.
```bash
# tc check jsonplaceholder.graphql --operations posts.graphql empty.graphql

N + 1: 1

Error: Validation Error
• Unknown field "userOd" on type "Post". Did you mean "userId", "user"? [examples/operations/posts.graphql, line 5 column 5]
• Unknown field "userAd" on type "Post". Did you mean "userId", "user"? [examples/operations/posts.graphql, line 6 column 5]
•  --> 1:1
  |
1 | 
  | ^---
  |
  = expected executable_definition [examples/operations/empty.graphql, line 1 column 1]


```

- `jsonplaceholder.graphql`
```graphql
schema
  @server(port: 8000, graphiql: true, hostname: "0.0.0.0", queryValidation: true)
  @upstream(baseURL: "http://jsonplaceholder.typicode.com", httpCache: true) {
  query: Query
}

type Query {
}

type User {
  id: Int!
  name: String!
  username: String!
  email: String!
  phone: String
  website: String
}

type Post {
  id: Int!
  userId: Int!
  title: String!
  body: String!
}
```
- `posts.graphql`
```graphql
query {
  posts {
    id
    body
    userOd
    userAd
  }
}
```
- `empty.graphql`: well, it's empty

**Issue Reference(s):**  
Fixes #416 

**Build & Testing:**

**Checklist:**

**Fun fact:**

I mistakenly deleted the branch, which closed the first [PR](https://github.com/tailcallhq/tailcall/pull/692) I opened.

/claim #416


src/cli/command.rs
```diff
@@ -42,6 +42,10 @@ pub enum Command {
     /// Display schema
     #[arg(short, long)]
     schema: bool,
+
+    /// Operations to check
+    #[arg(short, long, num_args = 1..)]
+    operations: Vec<String>,
```
b4s36t4, line 48: Add [value_delimiter](https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_delimiter) this to the arg so that even `,` comma separated values get captured.

b4s36t4 (changes requested):

examples/jsonplaceholder.graphql
```diff
@@ -1,5 +1,5 @@
 schema
-  @server(port: 8000, graphiql: true, hostname: "0.0.0.0")
+  @server(port: 8000, graphiql: true, hostname: "0.0.0.0", queryValidation: true)
```
ologbonowiwi, line 2:
```suggestion
  @server(port: 8000, graphiql: true, hostname: "0.0.0.0")
```

src/blueprint/into_schema.rs
```diff
@@ -39,26 +39,29 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
-                  a => FieldValue::from(a),
-                };
-                Ok(Some(p))
-              }
-            }
-          })
+            })
+          } else {
+            FieldFuture::new(async { Ok(None::<FieldValue>) })
+          }
```
ologbonowiwi, line 64: Is this an actual change, or was it committed by mistake?

src/cli/tc.rs
```diff
@@ -104,11 +108,37 @@ pub async fn init(file_path: &str) -> Result<()> {
...
+
+fn validate_operations(blueprint: &Blueprint, operations: Vec<String>) -> Result<()> {
+  tokio::runtime::Runtime::new()?.block_on(async {
+    let schema = blueprint.to_schema();
+    let mut execution = vec![];
+
+    for op in operations.iter() {
+      if let Ok(operation) = tokio::fs::read_to_string(op).await {
+        // Fake execution to get errors
+        let Response { errors, .. } = schema.execute(&operation).await;
```
ologbonowiwi, line 127: We don't have a way to validate errors without executing the schema?

src/cli/tc.rs
```diff
@@ -104,11 +108,37 @@ pub async fn init(file_path: &str) -> Result<()> {
...
   Fmt::display(Fmt::table(seq));
 }
+
+fn validate_operations(blueprint: &Blueprint, operations: Vec<String>) -> Result<()> {
+  tokio::runtime::Runtime::new()?.block_on(async {
+    let schema = blueprint.to_schema();
+    let mut execution = vec![];
+
+    for op in operations.iter() {
+      if let Ok(operation) = tokio::fs::read_to_string(op).await {
```
ologbonowiwi, line 125: Instead of adding an `if else` block here to check if it is ok, can you use `unwrap_or_else` instead? it'd make it more readable

src/cli/tc.rs
```diff
@@ -104,11 +108,37 @@ pub async fn init(file_path: &str) -> Result<()> {
...
+        // Fake execution to get errors
+        let Response { errors, .. } = schema.execute(&operation).await;
+        execution.push((op, errors));
+      } else {
+        return Err(anyhow!("Cannot read operation {}", op));
+      }
+    }
```
ologbonowiwi, line 132: Instead of creating a mut variable and pushing to it, we could use `.map` and assign it directly

e.g.:
```suggestion
    let mut execution = vec![];
    let execution = operations.iter()
        .map(/* the operation processing */)
```

ologbonowiwi: also, once we do this, we can map into `format_operation` directly, and on `Fmt::display`, we'll need only to `join(\n\n)` (and possibly `.collect::<Vec<String>>()`)

src/cli/command.rs
tobihans: Thank you.

src/blueprint/into_schema.rs
tobihans: That's a change. Because I want to avoid actually running the execution logic when checking the operation. Otherwise, checking will run real requests.****

src/cli/tc.rs
tobihans: Well, I tried at first. But that would just be too much work and I believe it's out of the scope of the project. `async_graphql` is already parsing well, so creating an `Operation` struct to handle all that, would be a lot of duplicated efforts. But, because the resolvers are fake when checking, the schema execution doesn't return any meaning result, yet it still performs all the required checks.

tobihans: I also thought about parsing the operation then validating the fields manually, but then, it's still repetitive.

src/cli/tc.rs
tobihans: By design, I exit if we fail to read the operation file. There is nothing to provide if we cannot read the op, I think.

src/cli/tc.rs
tobihans: But then, it will require to handle failures in reading in another way. I'll look into it. Thanks

src/cli/tc.rs
```diff
@@ -104,11 +108,37 @@ pub async fn init(file_path: &str) -> Result<()> {
 pub fn display_schema(blueprint: &Blueprint) {
   Fmt::display(Fmt::heading(&"GraphQL Schema:\n".to_string()));
   let sdl = blueprint.to_schema();
-  Fmt::display(print_schema::print_schema(sdl));
+  Fmt::display(format!("{}\n", print_schema::print_schema(sdl)));
```
ologbonowiwi, line 111: these changes are out of the scope of #416

```suggestion
  Fmt::display(print_schema::print_schema(sdl));
```

ologbonowiwi: I got that this meant to break the line, but we need to do so not here but where we'll print the other message

src/cli/tc.rs
```diff
@@ -104,11 +108,37 @@ pub async fn init(file_path: &str) -> Result<()> {
...
 }
 
 fn display_config(config: &Config, n_plus_one_queries: bool) {
-  Fmt::display(Fmt::success(&"No errors found".to_string()));
```
ologbonowiwi, line 111: We need to restore this; the message should be changed but not deleted

```suggestion
  Fmt::display(Fmt::success(&"No errors found on config".to_string()));
```

src/cli/tc.rs
ologbonowiwi: If you handle the failures as mentioned [here](https://github.com/tailcallhq/tailcall/pull/706/files#r1411120957) (with `unwrap_or_else`), you can do it with only a few tweaks.

ologbonowiwi (changes requested):
We need to find a way to write automated tests to this feature

src/cli/tc.rs
meskill: the example above is not right, `unwrap_or_else` is used to provide defaults if we have an error and will not end the function with the result. 

The current code is valid if we want to fail fast on errors. To simplify it we can use something like this:

```
let value = func().map_err(...)?;
```

src/blueprint/into_schema.rs
meskill: agree with @ologbonowiwi  - it's a good concern to not execute request for validation, but check like this is too implicit and error prone, so we would like to have some kind of explicit switch or function to have that custom logic

src/cli/tc.rs
```diff
@@ -104,11 +108,35 @@ pub async fn init(file_path: &str) -> Result<()> {
...
   let seq = vec![Fmt::n_plus_one_data(n_plus_one_queries, config)];
   Fmt::display(Fmt::table(seq));
 }
+
+fn validate_operations(blueprint: &Blueprint, operations: Vec<String>) -> Result<()> {
```
tusharmath, line 119: Can be moved to `blueprint`. Instead of result it should return a `Valid<(), String>`

src/cli/fmt.rs
```diff
@@ -40,6 +44,23 @@ impl Fmt {
...
+    let heading = Fmt::heading(&format!("Operation {op}: {meta}"));
+
+    let report = errors
+      .iter()
+      .enumerate()
+      .map(|(count, error)| Fmt::error(&format!("{}. {}", count + 1, error.message)))
+      .collect::<Vec<String>>()
+      .join("\n");
+
+    format!("{heading}\n{report}")
```
tusharmath, line 62: If you return a `Valid` it can be easily handled by the CLI internally because it's built on a ValidationError which is converted to a `CLIError` which is CLI friendly.

src/cli/tc.rs
```diff
@@ -104,11 +108,35 @@ pub async fn init(file_path: &str) -> Result<()> {
...
+  tokio::runtime::Runtime::new()?.block_on(async {
+    let schema = blueprint.to_validation_schema();
+    let mut execution = vec![];
+
+    for op in operations.iter() {
+      let operation = tokio::fs::read_to_string(op)
+        .await
+        .map_err(|_| anyhow!("Cannot read operation {}", op))?;
+        let Response { errors, .. } = schema.execute(&operation).await;
+        execution.push((op, errors));
```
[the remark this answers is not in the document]
tobihans: Yeah, it's not real execution. I create a fake schema with `blueprint.to_validation_schema`. It doesn't make real http calls. Thanks

src/cli/tc.rs
```diff
@@ -104,11 +108,35 @@ pub async fn init(file_path: &str) -> Result<()> {
...
+    }
```
ologbonowiwi, line 130: Now that we're handling errors with `map_err`, we can change this loop to a map. The code may require some tweaks to work 

```suggestion
    let execution = operations
      .iter()
      .map(async |op| {
        let operation = tokio::fs::read_to_string(op)
          .await
          .map_err(|_| anyhow!("Cannot read operation {}", op))? //This can likely be a CLIError
          
        let Response { errors, .. } = schema.execute(&operation).await;
        (op, errors)
      });
```

ologbonowiwi (changes requested):

src/blueprint/blueprint.rs
```diff
@@ -210,4 +214,29 @@ impl Blueprint {
...
+      match tokio::fs::read_to_string(op).await {
+        Ok(operation) => {
+          let Response { errors, .. } = schema.execute(&operation).await;
+          execution.push((op, errors.iter().map(|e| e.message.clone()).collect::<Vec<String>>()));
+        }
+        Err(_) => execution.push((op, vec!["Cannot read operation".to_string()])),
+      }
+    }
+
+    Valid::from_iter(execution.iter(), |(op, errors)| Fmt::format_operation(op, errors)).map(|_| ())
```
tobihans, line 240: Please, I'm very unfamiliar with the `Valid` struct and still trying to understand it. I don't know if someone can explain me why this results in duplicated errors in the console like this:
```bash
Error: Validation Error
• Operation examples/operations/posts.graphql: 2 errors found
1. Unknown field "userOd" on type "Post". Did you mean "userId", "user"?
2. Unknown field "userAd" on type "Post". Did you mean "userId", "user"? []
• Operation examples/operations/empty.graphql: 1 errors found
1. Cannot read operation []
Validation Error
• Operation examples/operations/posts.graphql: 2 errors found
1. Unknown field "userOd" on type "Post". Did you mean "userId", "user"?
2. Unknown field "userAd" on type "Post". Did you mean "userId", "user"? []
• Operation examples/operations/empty.graphql: 1 errors found
1. Cannot read operation []
```
Thanks

ologbonowiwi: You're using `Valid` here and then using `Valid::fail` on `format_operation`. Maybe that's the issue?

src/blueprint/into_schema.rs
```diff
@@ -133,6 +133,42 @@ fn create(blueprint: &Blueprint) -> SchemaBuilder {
   schema
 }
 
+pub(crate) fn create_dumb_schema(blueprint: &Blueprint) -> SchemaBuilder {
```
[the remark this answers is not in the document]
tobihans: It required to alter the behavior of the `create` function, which might lead to bugs, and updating the blueprint would still not prevent the real execution of the requests, which is why I added this method. But it reuses most of the previous code. It only makes _null_ the resolver.

tobihans:
Thanks for the branch update @tusharmath. I'm currently really sick. I hope to feel better soon and complete this. It remains to write automated tests like in `graphql_spec.rs`

src/blueprint/into_schema.rs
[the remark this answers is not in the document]
meskill: I totally get why you've added the method. 

But to express my concerns in more details: there is no code reuse here because you've actually created new code instead of using previous one. We can only hope that code acts the same way the original code works. For now it could be easy to see it's ok, but such code tends to diverge with time and it needs to be supported. Moreover, see the codecov comments in your pr? It shows this function is not covered by tests, so we probably need to create additional set of tests to make sure `create_dumb_schema` works as expected that should be same as tests for normal workflow

meskill:
@tobihans Hey, could you please update us on the status of the pr?

tobihans:
> @tobihans Hey, could you please update us on the status of the pr?

I'll ask for a review in one day or two. Thanks.

src/cli/tc.rs
ologbonowiwi: this review is still valid @tobihans

src/cli/tc.rs
ologbonowiwi: this review is still valid @tobihans

npm/gen-root.ts
```diff
@@ -73,7 +73,7 @@ async function genServerPackage(buildDefinitions: string[]) {
 
   const postInstallScriptContent = `const version = "${packageVersion}";\n${postInstallScript}`
   const preInstallScriptContent = `const optionalDependencies = ${JSON.stringify(
-    optionalDependencies,
```
ologbonowiwi, line 76: we shouldn't change this as does not relate with the issue being fixed

npm/gen-root.ts
```diff
@@ -73,7 +73,7 @@ async function genServerPackage(buildDefinitions: string[]) {
...
+    optionalDependencies
```
ologbonowiwi, line 76: we shouldn't change it as it does not relate with the issue being solved
```suggestion
    optionalDependencies,
```

src/blueprint/blueprint.rs
```diff
@@ -219,4 +221,53 @@ impl Blueprint {
...
+        Ok(operation) => {
+          let causes = Blueprint::validate_operation(&schema, &operation).await;
+          execution.push((op.to_string(), causes));
+        }
+        Err(_) => execution.push((
+          op.to_string(),
+          vec![Cause::new(format!("Cannot read file operation file {}", op))],
+        )),
+      }
+    }
```
ologbonowiwi, line 266: we should use `operations.iter().map()` rather than using for and `.push` values into execution.

src/cli/mod.rs
```diff
@@ -4,4 +4,5 @@ mod fmt;
 mod tc;
 
 pub use error::CLIError;
+pub use fmt::Fmt;
```
ologbonowiwi, line 7: we do use it outside `cli` module? if not, we don't need the `pub` identifier (we can use `pub(crate)` instead)

src/cli/tc.rs
ologbonowiwi: this review is still valid @tobihans

src/cli/tc.rs
ologbonowiwi: this review is still valid @tobihans. IMO, we shouldn't delete this line.

ologbonowiwi (changes requested):

src/blueprint/blueprint.rs
```diff
@@ -219,4 +221,53 @@ impl Blueprint {
...
+
+    Valid::from_iter(execution.iter(), |(op, causes)| {
+      Valid::<(), String>::from_vec_cause(causes.to_vec()).trace(op)
+    })
+    .map(|_: Vec<_>| ())
```
ologbonowiwi, line 271: why do we need this?

tests/operation_spec.rs
ologbonowiwi, line 1: great job finding a way to test operations!

tests/graphql/operations/jsonplaceholder.graphql
ologbonowiwi, line 1: can you add more test cases so we can increase coverage for this feature?

and congrats on what you've done so far, I believe it was quite tricky to test this

src/blueprint/blueprint.rs
tobihans: It converts the errors from all operations into one, and discard the expected `Vec<()>` for `()`.

tests/operation_spec.rs
tobihans: Thank you !

src/cli/mod.rs
tobihans: Solved. Thanks.

src/blueprint/blueprint.rs
tobihans: That allows to display a structured output with each failure linked to the appropriate file.

tests/graphql/operations/jsonplaceholder.graphql
tobihans: Yes, I'll look at the other tests and do it. Thanks

src/blueprint/blueprint.rs
tobihans: I can use `futures::stream` for that, but that will require adding a new dependency. So I think, this is ok.

npm/gen-root.ts
tobihans: Yeah, this is certainly a formatting thing from my editor. THanks

tests/operation_spec.rs
meskill, line 1: this file is very similar to the graphql_spec file nearby. 

Could this two be merged together? so we can just extend our graphql tests with operations check. Anyway it's related to graphql so it's ok not to split them

tests/graphql/operations/jsonplaceholder.graphql
```diff
@@ -0,0 +1,46 @@
...
+type Post {
+  id: Int!
+  userId: Int!
+  title: String!
+  body: String!
+  user: User @http(path: "/users/{{value.userId}}")
+}
+
+#> client-query
+query @diagnostic(count: 0) {
```
meskill, line 31: could we borrow the approach to test errors from the graphql tests like here https://github.com/tailcallhq/tailcall/blob/0295e5e16a9c13aa93864b155c3857090e9707a5/tests/graphql/errors/test-field-already-implemented-from-Interface.graphql#L21 ?

- more implicit naming for directive and arguments
- ability to test actual error

src/cli/tc.rs
```diff
@@ -37,18 +37,31 @@ pub fn run() -> Result<()> {
...
             display_schema(&blueprint);
           }
 
-          Ok(())
+          if let Some(out_file) = out_file_path {
+            tokio::runtime::Runtime::new()?.block_on(async { config.write_file(&out_file).await })?;
+            Fmt::display(Fmt::success(
+              &format!("Schema has been written to {}", out_file).to_string(),
+            ));
+          }
```
meskill, line 58: this code is not related to this pr and should be removed

src/blueprint/blueprint.rs
```diff
@@ -219,4 +221,53 @@ impl Blueprint {
...
     schema.finish().unwrap()
   }
+
+  /// Generate a dumb schema useful for validation.
+  pub fn to_validation_schema(&self) -> Schema {
+    create(self, Some(SchemaModifiers { no_resolver: true }))
+      .validation_mode(ValidationMode::Strict)
+      .disable_introspection()
+      .finish()
+      .unwrap()
```
meskill, line 231: prefer to return Result here instead of unwrap and make a validation error from this at the caller level

src/blueprint/blueprint.rs
```diff
@@ -219,4 +221,53 @@ impl Blueprint {
...
+          op.to_string(),
+          vec![Cause::new(format!("Cannot read file operation file {}", op))],
+        )),
+      }
+    }
+
+    Valid::from_iter(execution.iter(), |(op, causes)| {
+      Valid::<(), String>::from_vec_cause(causes.to_vec()).trace(op)
+    })
+    .map(|_| ())
```
meskill, line 271:
```suggestion
    .unit()
```

src/blueprint/blueprint.rs
```diff
@@ -219,4 +221,53 @@ impl Blueprint {
...
+          err
+            .locations
+            .iter()
+            .map(|loc| format!("line {} column {}", loc.line, loc.column)),
+        ))
+      })
+      .collect()
+  }
+
+  pub async fn validate_operations(&self, operations: Vec<String>) -> Valid<(), String> {
```
meskill, line 251: we need a way to see test output for this without testing it manually. Improving operation tests for this would help

meskill (changes requested):

src/blueprint/blueprint.rs
meskill: probably, using `futures_util::future::join_all` will be enough and we have that dependency.

Overall, we want to execute all the checks in parallel and not sequentially since it is expected to finish faster

src/blueprint/blueprint.rs
```diff
@@ -219,4 +221,61 @@ impl Blueprint {
     // generation of schema cannot fail.
     schema.finish().unwrap()
   }
+
+  /// Generate a dumb schema useful for validation.
+  pub fn to_validation_schema(&self) -> Result<Schema, ValidationError<String>> {
```
meskill, line 226: let's move all the validation stuff into separate file `blueprint/validation.rs`

src/blueprint/blueprint.rs
```diff
@@ -219,4 +221,61 @@ impl Blueprint {
...
+      .collect()
+  }
+
+  pub async fn validate_operations(&self, operations: Vec<String>) -> Valid<(), String> {
+    match self.to_validation_schema() {
+      Ok(schema) => {
+        let mut execution = vec![];
+
+        for op in operations.iter() {
+          match tokio::fs::read_to_string(op).await {
```
meskill, line 261: we have constraints in some environments like wasm on reading files so it's important to not have IO operations inside the core logic like blueprint. Let's move this reading out of blueprint and move it to `tc.rs` file, here just accept the operation content itself

meskill (changes requested):

ssddOnTop:
Can you resolve the conflicts

tobihans:
> Can you resolve the conflicts

Yes, I'll do it soon. Thanks

src/cli/tc.rs
tusharmath: I think we can convert this to a log message instead. So that people have an option to enable/disable it via env variables.

src/blueprint/operation.rs
```diff
@@ -0,0 +1,75 @@
...
+          .locations
+          .iter()
+          .map(|loc| format!("line {} column {}", loc.line, loc.column))
+          .collect(),
+      )
+    })
+    .collect()
+}
+
+pub async fn validate_operations(blueprint: &Blueprint, operations: Vec<String>) -> Valid<(), String> {
```
tusharmath, line 44: Anything inside of `blueprint` shouldn't be dependent on any kind of IO.
`ValidateOperation` should take in all the operations as string, parse and validate them.

IO Operations should happen within the CLI.

src/blueprint/into_schema.rs
```diff
@@ -177,20 +181,25 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
+    schema = schema.register(to_type(def, modifiers.map(|a| a.no_resolver).unwrap_or(false)));
   }
 
   schema
 }
 
 impl From<&Blueprint> for SchemaBuilder {
   fn from(blueprint: &Blueprint) -> Self {
-    create(blueprint)
+    create(blueprint, None)
```
tusharmath, line 203: Why are we making a change here?

tests/graphql/operations/jsonplaceholder.graphql
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#> client-query
+query @diagnostic(count: 1) {
+  user(id: 1) {
+    id
+    nome
+    username
+  }
+}
```
tusharmath, line 46: Tests in `/graphql` and `/http` exist because we have written that logical components that execute those bits. However here we are leveraging what async_graphql provides us. We should just test that when we call the methods on async_graphql, it returns results in the format we expect and that should be enough. 

The tests written in `operation.rs` should be enough for this usecase.

src/blueprint/into_schema.rs
tobihans: I'm reusing this method to create a validation schema that doesn't make actual calls. The method accepts an additional `Option<SchemaModifiers>` to modify the resulting schema. Normal behavior is then obtained by passing `None`.

src/cli/fmt.rs
```diff
@@ -13,10 +13,6 @@ impl Fmt {
     format!("{}", meta.yellow())
   }
 
-  pub fn success(success: &String) -> String {
-    format!("{}", success.green())
-  }
```
tusharmath, line 18: This is a library function, why delete?

src/cli/tc.rs
```diff
@@ -39,17 +40,31 @@ pub fn run() -> Result<()> {
       runtime.block_on(server.start())?;
       Ok(())
     }
-    Command::Check { file_paths, n_plus_one_queries, schema } => {
+    Command::Check { file_paths, n_plus_one_queries, schema, operations, out_file_path: _ } => {
```
tusharmath, line 43:
```suggestion
    Command::Check { file_paths, n_plus_one_queries, schema, operations } => {
```

src/blueprint/mod.rs
```diff
@@ -4,8 +4,10 @@ mod const_utils;
 mod from_config;
 mod into_schema;
 mod timeout;
+mod validation;
 
 pub use blueprint::*;
 pub use const_utils::*;
 pub use from_config::*;
 pub use timeout::GlobalTimeout;
+pub use validation::*;
```
tusharmath, line 13: Rename this to `operation`. Because this will contain functions for validating operations.

src/cli/command.rs
```diff
@@ -42,6 +42,14 @@ pub enum Command {
     /// Display schema
     #[arg(short, long)]
     schema: bool,
+
+    /// Path of the file to save final configuration in
+    #[arg(short, long("out"))]
+    out_file_path: Option<String>,
```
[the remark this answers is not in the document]
tobihans: Ok, let me remove it. It comes from me, pulling `upstream/main` to keep up to date.

src/cli/command.rs
```diff
@@ -42,6 +42,14 @@ pub enum Command {
...
+
+    /// Operations to check
+    #[arg(short('p'), long, value_delimiter=',', num_args = 1..)]
```
tusharmath, line 51:
```suggestion
    #[arg(short, long, value_delimiter=',', num_args = 1..)]
```

tusharmath (approved):

PR #718, opened 2023-12-05
Copying over documentation

rajatbarman:
**Summary:**  
Just copying over docs from tailcallhq.github.io repo for the sake of an implementation where we serve docs from this repo itself.


tusharmath:
Let's do this as the last step. Moving the documentation right now will cause confusion about where to update them.

tusharmath:
Let's open this once the website is ready.

PR #719, opened 2023-12-05
Removing Arc from Blueprint to prevent side effects

shashitnak:
**Summary:**  
_Removing Arc from Blueprint to prevent side effects._

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/lambda/expression.rs
```diff
@@ -202,8 +200,11 @@ async fn execute_request_with_dl<
 >(
   ctx: &EvaluationContext<'ctx, Ctx>,
   req: Request,
-  data_loader: &Option<Arc<DataLoader<Dl>>>,
-) -> Result<Response> {
+  data_loader_index: &Option<usize>,
```
tusharmath, line 203:
```suggestion
  data_loader_index: Option<usize>,
```

src/lambda/expression.rs
```diff
@@ -38,16 +40,12 @@ pub enum Context {
...
-    Option<GroupBy>,
-    Option<Arc<DataLoader<HttpDataLoader, NoCache>>>,
-  ),
+  Http(RequestTemplate, Option<GroupBy>, Option<usize>),
   GraphQLEndpoint {
     req_template: GraphqlRequestTemplate,
     field_name: String,
     batch: bool,
-    data_loader: Option<Arc<DataLoader<GraphqlDataLoader, NoCache>>>,
+    data_loader_index: Option<usize>,
```
tusharmath, line 48:
```suggestion
    dl_id: Option<usize>,
```

src/lambda/expression.rs
```diff
@@ -38,16 +40,12 @@ pub enum Context {
 
 #[derive(Clone)]
 pub enum Unsafe {
-  Http(
-    RequestTemplate,
-    Option<GroupBy>,
-    Option<Arc<DataLoader<HttpDataLoader, NoCache>>>,
-  ),
+  Http(RequestTemplate, Option<GroupBy>, Option<usize>),
```
tusharmath, line 43: May be we can take this opportunity to make the fields named here to be consistent with Unsafe::GraphQLEndpoint.

src/lambda/expression.rs
```diff
@@ -59,7 +57,7 @@ impl Debug for Unsafe {
         .debug_struct("Http")
         .field("req_template", req_template)
         .field("group_by", group_by)
-        .field("dl", &dl.clone().map(|a| a.clone().loader().batched.clone()))
+        .field("dl", dl)
         .finish(),
       Unsafe::GraphQLEndpoint { req_template, field_name, batch, .. } => f
         .debug_struct("GraphQLEndpoint")
```
tusharmath, line 63: I am guessing `Debug` can now be derived.

src/http/server_context.rs
```diff
@@ -12,9 +13,16 @@ pub struct ServerContext {
   pub schema: dynamic::Schema,
   pub http_client: Arc<dyn HttpClient>,
   pub blueprint: Blueprint,
+  pub http_data_loaders: Arc<Vec<DataLoader<HttpDataLoader>>>,
+  pub gql_data_loaders: Arc<Vec<DataLoader<GraphqlDataLoader>>>,
```
tusharmath, line 17: Do we need `Arc<Vec<_>>`? or can we use `Vec<Arc<_>>`? The second one might have lesser contention.

src/lambda/expression.rs
```diff
@@ -36,46 +38,22 @@ pub enum Context {
...
 pub enum Unsafe {
-  Http(
-    RequestTemplate,
-    Option<GroupBy>,
-    Option<Arc<DataLoader<HttpDataLoader, NoCache>>>,
-  ),
+  Http {
+    req_template: RequestTemplate,
+    group_by: Option<GroupBy>,
+    dl_id: Option<usize>,
```
meskill, line 46: actually usize is too much generic type. To do it more rusty way we may consider using [newtype](https://matklad.github.io/2018/06/04/newtype-index-pattern.html) pattern to have specific id type for resolving the dl. This way we won't confuse what does usize means and how to properly acquire the dataloader.

shashitnak: @meskill added a `struct DataLoaderId(usize)` to use `dl_id`. Could you please check if the current changes are the way you suggested or if there's something you would like me to change?

tusharmath:
@shashitnak There is a slight degradation in perf — https://github.com/tailcallhq/tailcall/actions/runs/7116029969/job/19373484273?pr=719#step:17:16

Compared to `main` https://github.com/tailcallhq/tailcall/commit/9b203052fa244941e9c304e149365706b04a2078#comments

shashitnak:
> @shashitnak There is a slight degradation in perf — https://github.com/tailcallhq/tailcall/actions/runs/7116029969/job/19373484273?pr=719#step:17:16
> 
> Compared to `main` [9b20305#comments](https://github.com/tailcallhq/tailcall/commit/9b203052fa244941e9c304e149365706b04a2078#comments)

@tusharmath could it be because of the conversion from Arc<Vec<_>> to Vec<Arc<_>>? because it's creating a new list with the references instead of taking the reference to the entire list

src/lambda/expression.rs
shashitnak: @meskill added the type parameter just as you suggested

src/lambda/expression.rs
```diff
@@ -36,43 +39,41 @@ pub enum Context {
...
+  }
+}
+
+impl<T> Clone for DataLoaderId<T> {
+  fn clone(&self) -> Self {
+    *self
+  }
+}
+
+impl<T> Copy for DataLoaderId<T> {}
```
meskill, line 72: does using PhantomData somehow affects `#[derive]` ?

shashitnak: yes, `#[derive]` expects the type parameter T to also implement the trait that we want to derive

src/lambda/expression.rs
```diff
@@ -36,43 +39,41 @@ pub enum Context {
...
-    data_loader: Option<Arc<DataLoader<GraphqlDataLoader, NoCache>>>,
+    dl_id: Option<DataLoaderId<GraphqlDataLoader>>,
   },
   JS(Box<Expression>, String),
 }
 
-impl Debug for Unsafe {
+pub struct DataLoaderId<T>(pub usize, PhantomData<T>);
+
+impl<T> DataLoaderId<T> {
```
tusharmath, line 60: Drop the type-param from DataLoaderId.

tusharmath (approved):

src/lambda/expression.rs
meskill: type safety in such cases helps to prevent wrong usages and helps with code understanding. While this case is pretty simple to understand, it still requires more mental energy to remember what usize is all about and how to use it - i.e. acquire it only from that function, pass it only to other. While new_type put heavy constraints on what could be done and prevents possible misuse in future, especially for devs who didn't work with that code before. 

And it's totally possible to parameterize here as we're using different dataloaders types in different expressions and therefore always know the type on compilation

src/http/request_context.rs
```diff
@@ -87,13 +93,31 @@ impl RequestContext {
   }
 }
 
+pub trait GetDataLoader<Dl> {
```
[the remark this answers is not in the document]
meskill: just a reminder that traits doesn't involve dynamic dispatch until we use trait object. In most cases traits are just the feature of type system not a runtime

PR #720, opened 2023-12-05
A very draft for the resolver plan

meskill:


sujeetsr:
@meskill are you familiar with [grafast](https://grafast.org/grafast/) ?
https://www.youtube.com/watch?v=x0FMjL5-kNI

The plan resolver looks very similar to our blueprint and expression.

Issue #402, opened 2023-10-02
Tailcall LSP integration


tusharmath:
**Problem**
Currently one needs to always go to the documentation website to read the configuration parameters and then set them up in the configuration file. We want to leverage IDE capabilities for GraphQL to provide intellisense and improve developer experience.

**Solution**
Create a new `tc init` command that bootstraps a tailcall project:
1. Creates a `.tailcallrc.graphql` like [this](https://github.com/tailcallhq/tailcall/blob/main/assets/.tailcallrc.graphql)
2. Asks the user if they want to add a file to the project if they say **no** - exit with just the `.tailcallrc.graphql` file, otherwise if they say **yes**:
  1. Asks the user to enter a the file name and create a `<file name>.graphql` file.
  2. Create a `.graphqlrc.yml` file with the following content 
    
      ```yml
      schema:
        - "./<file name>.graphql"
        - "./.tailcallrc.graphql"
      ```

PR #733, opened 2023-12-10
fix: Simplify init subcommand

### LINKED ISSUE #698: Simplify init subcommand
Following on from our discussion in discord.
- The first question doesn't make too much sense to me. If I didn't mean to call the command, I would cancel the command. It also errors if the directory doesn't exist.
- If the directory I'm targeting doesn't exist, init should create the directory
- I wasn't initially aware that init created 2 hidden files, as well as an empty file with the name of the schema I wanted

Feel free to break this up into smaller tickets, especially the last point.

- Remove the first question
- Create the directory (or at least ask if I want it created if it doesn't already exist)
- Some output of what was created would be helpful (I appreciate this is in the docs already), or maybe expand on the help section of the subcommand.

**Describe alternatives you've considered**
See above (for the or)

**Additional context**
image
Diagram should be amended to say errors if the user presses no at first and the dir doesn't exist.

neo773:
Fixes #698

/claim #698

### DIFF
--- a/src/cli/tc.rs
+++ b/src/cli/tc.rs
@@ -5,6 +5,7 @@ use anyhow::Result;
 use clap::Parser;
 use inquire::Confirm;
 use log::Level;
+use stripmargin::StripMargin;
 use tokio::runtime::Builder;
 
 use super::command::{Cli, Command};
@@ -51,30 +52,71 @@ pub fn run() -> Result<()> {
 }
 
 pub async fn init(file_path: &str) -> Result<()> {
-  let file_name = ".tailcallrc.graphql";
-  let tailcallrc = include_str!("../../examples/.tailcallrc.graphql");
-  let tailcallrc_path = Path::new(file_path).join(file_name);
+  let tailcall_rc = ".tailcallrc.graphql";
+  let graphql_rc = ".graphqlrc.yml";  
+  let tailcallrc_data = include_str!("../../examples/.tailcallrc.graphql");
+  let graphql_rc_exists = fs::metadata(format!("{}/{}", file_path, graphql_rc)).is_ok();
+  let path = Path::new(file_path).join(tailcall_rc);
 
-  if let Some(parent) = tailcallrc_path.parent() {
+  if let Some(parent) = path.parent() {
     fs::create_dir_all(parent)?;
   }
 
-  if tailcallrc_path.exists() {
-    let overwrite = Confirm::new(&format!(
-      "File {} already exists. Do you want to overwrite it?",
-      file_name
-    ))
-    .with_default(false)
-    .prompt()?;
+  if !graphql_rc_exists {
+    fs::write(format!("{}/{}", file_path, graphql_rc), "")?;
 
-    if !overwrite {
-      return Ok(());
-    }
+    let graphqlrc = r#"|schema:
+         |- './.tailcallrc.graphql'
+    "#
+    .strip_margin();
+
+    fs::write(format!("{}/.graphqlrc.yml", file_path), graphqlrc)?;
+    Fmt::display(Fmt::success(&format!("Created file .graphqlrc.yml in {}", file_path)));
   }
 
-  fs::write(tailcallrc_path, tailcallrc)?;
+  let tailcall_exists = fs::metadata(format!("{}/{}", file_path, tailcall_rc)).is_ok();
 
-  Fmt::display(Fmt::success(&format!("Created files in {}: {}", file_path, file_name,)));
+  if tailcall_exists {
+    let confirm = Confirm::new(&format!("Do you want to overwrite the file {}?", tailcall_rc))
+      .with_default(false)
+      .prompt();
+
+    match confirm {
+      Ok(true) => fs::write(format!("{}/{}", file_path, tailcall_rc), tailcallrc_data)?,
+      Ok(false) => (),
+      Err(e) => return Err(e.into()),
+    };
+  } else {
+    fs::write(format!("{}/{}", file_path, tailcall_rc), tailcallrc_data)?;
+    Fmt::display(Fmt::success(&format!("Created file .tailcallrc.graphql in {}", file_path)));
+  }
+
+  let graphqlrc_path = format!("{}/.graphqlrc.yml", file_path);
+  let graphqlrc = fs::read_to_string(&graphqlrc_path)?;
+
+  if !graphqlrc.contains(tailcall_rc) {
+    let confirm = Confirm::new(&format!("Do you want to add {} to the schema?", tailcall_rc))
+      .with_default(false)
+      .prompt();
+
+    match confirm {
+      Ok(true) => {
+        let mut schema_line = graphqlrc
+          .lines()
+          .find(|line| line.contains("schema:"))
+          .unwrap()
+          .to_string();
+
+        schema_line.push_str("\n  - './.tailcallrc.graphql'");
+
+        let updated = graphqlrc.replace("schema:", &schema_line);
+
+        fs::write(graphqlrc_path, updated)?;
+      }
+      Ok(false) => (),
+      Err(e) => return Err(e.into()),
+    }
+  }
 
   Ok(())
 }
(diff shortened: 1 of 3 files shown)


src/cli/tc.rs
```diff
@@ -55,53 +56,30 @@ pub fn run() -> Result<()> {
...
-            &file_name
-          )
-          .strip_margin();
-          fs::write(format!("{}/.graphqlrc.yml", file_path), graphqlrc)?;
-        }
-        Ok(false) => (),
-        Err(e) => return Err(e.into()),
-      }
+    if !overwrite {
+      return Ok(());
```
ologbonowiwi, line 71: We don't need `./graphqlrc.yml` anymore?

neo773: Yeah, If Im not wrong that's what I got from the issue conversation.

src/cli/tc.rs
```diff
@@ -55,53 +56,30 @@ pub fn run() -> Result<()> {
...
-        .prompt();
+  if let Some(parent) = tailcallrc_path.parent() {
+    fs::create_dir_all(parent)?;
+  }
 
-      match confirm {
-        Ok(true) => {
-          fs::write(format!("{}/{}", file_path, &file_name), "")?;
+  if tailcallrc_path.exists() {
+    let overwrite = Confirm::new("A .tailcallrc.graphql file already exists. Do you want to overwrite it?")
```
tusharmath, line 66: Keep `.tailcallrc.graphql` name in a variable.

src/cli/tc.rs
```diff
@@ -55,53 +56,30 @@ pub fn run() -> Result<()> {
...
+    if !overwrite {
+      return Ok(());
     }
-    Ok(false) => (),
-    Err(e) => return Err(e.into()),
+  } else {
+    println!(
+      "{}",
+      format!("Created .tailcallrc.graphql file in {}", file_path).blue()
+    );
```
tusharmath, line 77: Can use `fmt::success` instead of println

src/cli/tc.rs
tusharmath: @ologbonowiwi We need the file. We don't so much of ceremony around creating the necessary files.

src/cli/tc.rs
```diff
@@ -55,53 +56,30 @@ pub fn run() -> Result<()> {
 
 pub async fn init(file_path: &str) -> Result<()> {
```
tusharmath, line 57: Can you add better documentation to the `init` command, currently it just says, that it initializes the current project.

src/cli/tc.rs
ologbonowiwi: In this case, can you re-add it, @neo773?

src/cli/tc.rs
```diff
@@ -55,31 +56,47 @@ pub fn run() -> Result<()> {
 }
 
 pub async fn init(file_path: &str) -> Result<()> {
+  let file_name = ".tailcallrc.graphql";
+  let file_name_yml = ".graphqlrc.yml";
   let tailcallrc: resource::Resource<str> = resource_str!("examples/.tailcallrc.graphql");
```
neo773, line 61: About this resource_str expects a static string can't really substitute file_name dynamically

src/cli/tc.rs
neo773: Added

src/cli/tc.rs
neo773: @tusharmath Sure, will do that in docs repo

src/cli/tc.rs
```diff
@@ -54,51 +56,44 @@ pub fn run() -> Result<()> {
...
+  let tailcallrc_path = Path::new(file_path).join(file_name);
+  let graphqlrc = format!(
+    r#"|schema:
+       |- './{}'
+       |- './{}'
+    "#,
+    &file_name, file_name
```
ologbonowiwi, line 68: Since we don't create the user's `.graphql` file, we no longer need to declare it on `graphqlrc`.

```suggestion
  let graphqlrc = format!(
    r#"|schema:
       |- './{}'
    "#,
    &file_name
```

ologbonowiwi (changes requested):

src/cli/tc.rs
neo773: fixed, good catch thank ypu

src/cli/tc.rs
```diff
@@ -54,51 +56,43 @@ pub fn run() -> Result<()> {
 }
 
 pub async fn init(file_path: &str) -> Result<()> {
+  let file_name = ".tailcallrc.graphql";
+  let file_name_yml = ".graphqlrc.yml";
   let tailcallrc: resource::Resource<str> = resource_str!("examples/.tailcallrc.graphql");
```
ologbonowiwi, line 61: I know this is not part of your change, but can we parse the type here as a string or the bytes to create the file?

Since we're using it only on [88](https://github.com/tailcallhq/tailcall/pull/733/files#diff-b00853da75ac44d2390ae9c7899e6b06c03a50693ada790c3ab1e2f45ca9ffd4R88), we could get the actual type that we'll use directly instead of getting a `resource::Resource` to parse it afterward.

ologbonowiwi: this will return the string as expected

```suggestion
  let tailcallrc = include_str!("examples/.tailcallrc.graphql");
```

neo773: Wouldn't work at runtime resource_str bundles it with the binary

ologbonowiwi:
```suggestion
  let tailcallrc = include_str!("../../examples/.tailcallrc.graphql");
```

src/cli/tc.rs
```diff
@@ -54,51 +56,43 @@ pub fn run() -> Result<()> {
...
     }
-    Ok(false) => (),
-    Err(e) => return Err(e.into()),
   }
 
-  fs::write(
-    format!("{}/.tailcallrc.graphql", file_path),
-    tailcallrc.as_ref().as_bytes(),
-  )?;
+  fs::write(tailcallrc_path, tailcallrc.as_ref().as_bytes())?;
```
ologbonowiwi, line 88:
```suggestion
  fs::write(tailcallrc_path, tailcallrc)?;
```

src/cli/tc.rs
```diff
@@ -1,7 +1,9 @@
 use std::fs;
+use std::path::Path;
 
 use anyhow::Result;
 use clap::Parser;
+
 use inquire::Confirm;
 use log::Level;
 use resource::resource_str;
```
ologbonowiwi, line 9:
```suggestion
```

ologbonowiwi (changes requested):
tested the changes locally, it works as expected

ologbonowiwi (approved):
good work! :rocket:

ologbonowiwi:
Hey, @neo773, PR is approved, but I noticed a few minutes ago that `resource` is an external library, so we can likely delete it as the only place it was used is on the `init` command.

neo773:
@ologbonowiwi 
Resolved, thanks.

src/cli/tc.rs
```diff
@@ -54,51 +54,43 @@ pub fn run() -> Result<()> {
 }
 
 pub async fn init(file_path: &str) -> Result<()> {
-  let tailcallrc: resource::Resource<str> = resource_str!("examples/.tailcallrc.graphql");
+  let file_name = ".tailcallrc.graphql";
+  let file_name_yml = ".graphqlrc.yml";
```
tusharmath, line 58: As per the issue, we should just drop the graphqlrc.yml file.

meskill: the issue doesn't state that file should be removed. And it's mainly used for LSP integration that was initial intention for adding `init` command https://github.com/tailcallhq/tailcall/issues/402.

Could we revisit this? I think it shouldn't be removed.

meskill: actually, I would even add additional file like `examples/jsonplaceholder.graphql` to make init command suitable for quick bootstrapping and testing. In that case after calling `tailcall init` we can quickly start the server with example configuration

tusharmath: My bad. I meant:
1. We should NOT delete the `tailcallrc.graphql`. 
2. We should delete the `.graphql.yml`

ologbonowiwi:
@tusharmath @meskill is there anything missing to merge this one guys?

meskill:
> @tusharmath @meskill is there anything missing to merge this one guys?

There were some uncertainties about the implementation. @neo773 could you please take a look into the last comments for the issue and say what do you think about suggested implementation?

PR #736, opened 2023-12-11
Feature: open playground in the browser automatically if enabled

### LINKED ISSUE #735: Automatically open the GraphiQL IDE on Start
On `tailcall start` command, the CLI should automatically open the default browser with the configured IDE link.

**Technical Specification**
- Make sure this only happens if the graphiql setting is enabled.
- Do not write code from scratch, instead use https://github.com/amodm/webbrowser-rs or https://github.com/Seeker14491/opener and justify your reasoning in the PR.

ologbonowiwi:
- chore: install webbrowser
- chore: enable `hardened` and `disable-wsl` features
- feat: create a method to open browser
- feat: change `log_launch` to also open browser

**Summary:**  
This PR adds [webbrowser](https://github.com/amodm/webbrowser-rs) to open the default browser when the server starts.
I'm using [webbrowser](https://github.com/amodm/webbrowser-rs) instead [opener](https://github.com/Seeker14491/opener) for three main reasons:
- Option to turn off `file://` URLs, which is good for security (we only need to open `http://` and `https://`)
- Compatibility with multiple platforms
- According to them:
> **Browser guarantee** - This library guarantees that the browser is opened, even for local files - the only crate to make 
such guarantees at the time of this writing. Alternative libraries rely on existing system commands, which may lead to an editor being opened (instead of the browser) for local html files, leading to an inconsistent behaviour for users. 

**Issue Reference(s):**  
Fixes #735
/claim #735

**Build & Testing:**

**Checklist:**

- ~[ ] I have added relevant unit & integration tests.~
I have not added tests, as it'd be too painful to ensure that the browser was opened with the expected URL, and this wouldn't have that high impact that pay the effort.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ futures-channel = { version = "0.3.29" }
 futures-timer = { version = "3.0.2" }
 futures-util = { version = "0.3.29" }
 lru = { version = "0.12.1" }
-webbrowser = "0.8.12"
+webbrowser = { version = "0.8.12", features = ["hardened", "disable-wsl"] }
 
 [dev-dependencies]
 criterion = "0.5.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ futures-channel = { version = "0.3.29" }
 futures-timer = { version = "3.0.2" }
 futures-util = { version = "0.3.29" }
 lru = { version = "0.12.1" }
+webbrowser = "0.8.12"
 
 [dev-dependencies]
 criterion = "0.5.1"
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -77,15 +77,9 @@ fn log_launch_and_open_browser(sc: &ServerConfig) {
     let url = sc.graphiql_url();
     log::info!("🌍 Playground: {}", url);
 
-    open_browser(url);
-  }
-}
-
-fn open_browser(graphiql_url: String) {
-  let opened = webbrowser::open_browser(webbrowser::Browser::Default, graphiql_url.as_str());
-
-  if opened.is_ok() {
-    log::info!("{} opened on your browser", graphiql_url);
+    if let Ok(_) = webbrowser::open(url.as_str()) {
+      log::info!("{} opened on your browser", url);
+    }
   }
 }
 
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -77,9 +77,7 @@ fn log_launch_and_open_browser(sc: &ServerConfig) {
     let url = sc.graphiql_url();
     log::info!("🌍 Playground: {}", url);
 
-    if webbrowser::open(url.as_str()).is_ok() {
-      log::info!("{} opened on your browser", url);
-    }
+    let _ = webbrowser::open(url.as_str());
   }
 }
 
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -88,7 +88,7 @@ fn open_browser(graphiql_url: String) {
 
   match opened {
     Ok(_) => log::info!("{} opened on your browser", graphiql_url),
-    Err(e) => log::error!("Failed to open default browser, error: {}", e)
+    Err(e) => log::error!("Failed to open default browser, error: {}", e),
   };
 }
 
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -62,6 +62,7 @@ futures-channel = { version = "0.3.29" }
 futures-timer = { version = "3.0.2" }
 futures-util = { version = "0.3.29" }
 lru = { version = "0.12.1" }
+webbrowser = { version = "0.8.12", features = ["hardened", "disable-wsl"] }
```
ologbonowiwi, line 65: as of [their docs page](https://docs.rs/webbrowser/latest/webbrowser/):
- `hardened` - this disables handling of non-http(s) urls (e.g. file:///) as a hard security precaution
- `disable-wsl` - this disables WSL file implementation (http still works)

I've turned on these features as we don't want/need to open files on browser.

src/http/mod.rs
```diff
@@ -70,14 +70,28 @@ pub fn min_ttl<'a>(res_vec: impl Iterator<Item = &'a Response>) -> i32 {
...
-fn log_launch(sc: &ServerConfig) {
+fn log_launch_and_open_browser(sc: &ServerConfig) {
   let addr = sc.addr().to_string();
   log::info!("🚀 Tailcall launched at [{}] over {}", addr, sc.http_version());
   if sc.graphiql() {
-    log::info!("🌍 Playground: {}", sc.graphiql_url());
+    let url = sc.graphiql_url();
+    log::info!("🌍 Playground: {}", url);
+
+    log::info!("🌍 Opening playground on your default browser");
```
tusharmath, line 80:
```suggestion
```
We don't need this log.

src/http/mod.rs
```diff
@@ -70,14 +70,28 @@ pub fn min_ttl<'a>(res_vec: impl Iterator<Item = &'a Response>) -> i32 {
...
   }
 }
 
+fn open_browser(graphiql_url: String) {
+  let opened = webbrowser::open_browser(webbrowser::Browser::Default, graphiql_url.as_str());
+
+  match opened {
+    Ok(_) => log::info!("{} opened on your browser", graphiql_url),
+    Err(e) => log::error!("Failed to open default browser, error: {}", e),
+  };
```
tusharmath, line 92: This should silently fail.

src/http/mod.rs
```diff
@@ -70,11 +70,22 @@ pub fn min_ttl<'a>(res_vec: impl Iterator<Item = &'a Response>) -> i32 {
...
+    open_browser(url);
+  }
+}
+
+fn open_browser(graphiql_url: String) {
+  let opened = webbrowser::open_browser(webbrowser::Browser::Default, graphiql_url.as_str());
+
+  if opened.is_ok() {
+    log::info!("{} opened on your browser", graphiql_url);
   }
```
tusharmath, line 89:
```suggestion
```
Not required. It should just open in the browser

src/http/mod.rs
```diff
@@ -70,14 +70,22 @@ pub fn min_ttl<'a>(res_vec: impl Iterator<Item = &'a Response>) -> i32 {
...
-fn log_launch(sc: &ServerConfig) {
+fn log_launch_and_open_browser(sc: &ServerConfig) {
   let addr = sc.addr().to_string();
   log::info!("🚀 Tailcall launched at [{}] over {}", addr, sc.http_version());
   if sc.graphiql() {
-    log::info!("🌍 Playground: {}", sc.graphiql_url());
+    let url = sc.graphiql_url();
+    log::info!("🌍 Playground: {}", url);
+
+    open_browser(url);
```
tusharmath, line 80: we can inline the function here.

src/http/mod.rs
```diff
@@ -70,11 +70,16 @@ pub fn min_ttl<'a>(res_vec: impl Iterator<Item = &'a Response>) -> i32 {
...
   let addr = sc.addr().to_string();
   log::info!("🚀 Tailcall launched at [{}] over {}", addr, sc.http_version());
   if sc.graphiql() {
-    log::info!("🌍 Playground: {}", sc.graphiql_url());
+    let url = sc.graphiql_url();
+    log::info!("🌍 Playground: {}", url);
+
+    if let Ok(_) = webbrowser::open(url.as_str()) {
+      log::info!("{} opened on your browser", url);
+    }
```
tusharmath, line 82: No need for these logs actually. Just open silently and let it fail also silently.

ologbonowiwi: done

tusharmath (approved):

PR #741, opened 2023-12-12
docs: add documentation for .tailcallrc.graphql

### LINKED ISSUE #740: Add documentation to each operator and it fields
`.tailcallrc.graphql` doesn't have any documentation. We should add some documentation using `"""` for each operator and field to make sure that there is some helpful text available with auto-complete.

neo773:
Adds documentation for directives properties in .taillrc.graphql

https://github.com/tailcallhq/tailcall/assets/62795688/441b9109-e019-40dc-8a42-f52f69330585

/fixes #740
/claim #740

### DIFF
--- a/examples/.tailcallrc.graphql
+++ b/examples/.tailcallrc.graphql
@@ -226,27 +226,27 @@ directive @const(data: JSON) on FIELD_DEFINITION
 The @graphQL operator allows to specify GraphQL API server request to fetch data from.
 """
 directive @graphQL(
-"""
-This refers to the base URL of the API. If not specified, the default base URL is the one specified in the `@upstream` operator.
-"""
+  """
+  This refers to the base URL of the API. If not specified, the default base URL is the one specified in the `@upstream` operator.
+  """
   baseURL: String
-"""
-Specifies the root field on the upstream to request data from. This maps a field in your schema to a field in the upstream schema. When a query is received for this field, Tailcall requests data from the corresponding upstream field.
-"""
+  """
+  Specifies the root field on the upstream to request data from. This maps a field in your schema to a field in the upstream schema. When a query is received for this field, Tailcall requests data from the corresponding upstream field.
+  """
   name: String
-"""
-Named arguments for the requested field. More info [here](https://tailcall.run/docs/guides/operators/#args)
-"""
+  """
+  Named arguments for the requested field. More info [here](https://tailcall.run/docs/guides/operators/#args)
+  """
   args: [KeyValue]
-"""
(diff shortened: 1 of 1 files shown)


ologbonowiwi (changes requested):
Nice work documenting the properties! It looks really good.

Although, it's lacking docs for the operators itself (e.g.: `@server`, `@upstream`), and also to the operations and fields below line [194](https://github.com/tailcallhq/tailcall/pull/741/files#diff-df81d69338b9c82f79a6c2b67791c6865740f1c1b8c6bac6af6b8021e2399dc3R194).

neo773:
@ologbonowiwi  @tusharmath 
Hmm do we really need it? seemed self explanatory

ologbonowiwi:
IMO, we need it. I usually see newcomers have confusion about `@server` and `@upstream`.

As for the other fields, they are self-explanatory to us who are familiar with the project, but I'm not so sure about people who are having their first experience with tailcall

neo773:
> I agree @ologbonowiwi - we should add documentation for everything. Secondly and more importantly, the documentation should be a little more detailed. Currently it looks like GPT generated low effort work (sorry 🙈 )

The original version looked way too verbose for the tiny detail box, that's why I went with a concise version
Will rewrite it

neo773:
@tusharmath 
Pushed a new commit, rewrote it all

examples/.tailcallrc.graphql
```diff
@@ -22,35 +123,312 @@ enum HttpVersion {
...
+}
+
+type User {
+  name: String
+  age: Int
+}
+```
+"""
 directive @const(data: JSON) on FIELD_DEFINITION
 directive @graphQL(
```
ologbonowiwi, line 433: I guess we need docs also for these (and it's respective fields)

ologbonowiwi (changes requested):

examples/.tailcallrc.graphql
```diff
@@ -1,19 +1,120 @@
...
+  """
+  `responseHeaders` appends headers to all server responses, aiding cross-origin requests or extra headers for downstream services.
+
+  The responseHeader is a key-value pair array. These headers are included in every server response. Useful for headers like Access-Control-Allow-Origin for cross-origin requests, or additional headers like X-Allowed-Roles for downstream services.
+
+  ```graphql
+  schema @server(responseHeaders: [{key: "X-Allowed-Roles", value: "admin,user"}]) {
+    query: Query
+    mutation: Mutation
+  }
```
tusharmath, line 81: Can drop the code from the documentation.

examples/.tailcallrc.graphql
tusharmath: Agree.

tusharmath (approved):

PR #742, opened 2023-12-12
fix: disable batching when delay is set to 0

### LINKED ISSUE #737: There shouldn't be any batching when batch.delay is set to 0
### Prerequisites

### Describe the bug
When `batch: {delay: 0}` is set in the config no batching should occur.

### Steps to reproduce

1. Run the benchmarks with and without `batch: {delay: 0}` added in the config
2. Run the benchmarks with and without `batch: {delay: 1}` added in the config

### Expected behavior

There shouldn't be any difference in performance between both the benchmarks in the first step.
In the second step, the one with batching enabled should be faster.

### Actual behavior

Both `batch: {delay: 0}` and `batch: {delay: 1}` give similar performance

### Environment information:

- **Ubuntu 22.04 LTS**
- **Tailcall version v0.17.0**
- **rustc 1.74.1**
- **wrk/4.1.0-3**

adelinaenache:
**Summary:**  
Make multiple requests when `upstream.batch.delay` is 0. 
If the graphql schema expects a single entry for a field but the result from the endpoint is an array with a single value, extract the value from the array.

**Issue Reference(s):**  
Fixes #737 
/claim #737 
**Build & Testing:**

**Checklist:**


src/blueprint/into_schema.rs
```diff
@@ -48,13 +48,20 @@ fn to_type(def: &Definition) -> dynamic::Type {
             }
             Some(expr) => {
               let expr = expr.to_owned();
+              let field_of_type = type_ref.clone();
```
tusharmath, line 51: Why do we need these changes?

src/http/request_context.rs
```diff
@@ -91,6 +93,10 @@ impl RequestContext {
       self.set_min_max_age(-1);
     }
   }
+
+  pub fn is_batching_enabled(&self) -> bool {
+    return self.upstream.batch.is_some() && self.upstream.get_delay() != 0;
```
tusharmath, line 98: Enable only if delay is >= 1 or max_size >= 1

src/http/request_context.rs
```diff
@@ -25,9 +25,11 @@ pub struct RequestContext {
 impl Default for RequestContext {
   fn default() -> Self {
     let config = config::Config::default();
+    let mut upstream = config.upstream.clone();
+    upstream.batch = Some(Batch::default());
     //TODO: default is used only in tests. Drop default and move it to test.
     let server = Server::try_from(config.server.clone()).unwrap();
-    RequestContext::new(Arc::new(DefaultHttpClient::default()), server, config.upstream.clone())
+    RequestContext::new(Arc::new(DefaultHttpClient::default()), server, upstream)
```
tusharmath, line 32: Why do we need these changes?

tusharmath:
@adelinaenache can you please add integration tests in the `/tests/http` directory also.

src/blueprint/into_schema.rs
adelinaenache: The problem was that for the '/users&id=?` request the endpoint was returning an array but the GQL schema was expecting a single object, therefore returning an error to the GQL query.
 Before making this change, the request would go through the batched DataLoader that would take the array and process it, but now the response would only be processed by this future field resolver that did not expect to receive an array. 
 
My approach to this was to check the expected field type, and if was a single object but the response from the REST API was an array, extract the object from the array. Hope this is the right place to do it, if not please let me know how to approach this, thanks! 

This specific line was added because i needed the type of the field in a `match` context.

src/http/request_context.rs
adelinaenache: You are right, this was not needed.

tests/http/batching-disabled.yml
```diff
@@ -0,0 +1,51 @@
+---
+config: !file tests/http/config/batching-disabled.graphql
```
tusharmath, line 2: can use `!inline` and inline the YML here.

tests/http/batching-disabled.yml
```diff
@@ -0,0 +1,82 @@
...
+    response:
+      body:
+        - id: -1
+          name: "THIS SHOULD NOT BE CALLED"
+  - request:
+      url: http://jsonplaceholder.typicode.com/users?id=2&id=1
+    response:
+      body:
+        - id: -1
+          name: "THIS SHOULD NOT BE CALLED"
```
amitksingh1490, line 68: you can drop this as if the mock is not present the test will automatically fail

src/blueprint/into_schema.rs
amitksingh1490: Ideally this should not be done. It should fail as wrong URL is configured.

src/blueprint/into_schema.rs
```diff
@@ -48,13 +48,20 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
+
               FieldFuture::new(async move {
                 let ctx = EvaluationContext::new(req_ctx, &ctx);
                 let const_value = expr.eval(&ctx).await?;
                 let p = match const_value {
-                  ConstValue::List(a) => FieldValue::list(a),
+                  ConstValue::List(a) => match field_of_type {
+                    dynamic::TypeRef::Named(_) => FieldValue::from(a[0].clone()),
+                    dynamic::TypeRef::NonNull(_) => FieldValue::from(a),
+                    dynamic::TypeRef::List(_) => FieldValue::list(a),
```
amitksingh1490, line 60: This is also not needed

amitksingh1490 (approved):

PR #748, opened 2023-12-14
Support multiple upstream directives

### LINKED ISSUE #729: Support multiple upstream settings
## Is your feature request related to a problem? Please describe.

`@upstream` directive is used to define some settings for the upstream server that will affect all of the requests made by tailcall. Currently, it's only possible to define one upstream setup that will be shared by all of the resolvers. Current approach, doesn't allow to specify different settings for different upstream and the limits setup ability for cases of tailcall to multiple non-equivalent apis.

## Describe the solution you'd like

Able to specify multiple `@upstream` directives for different apis that could be referenced by id for resolvers. 

Or rethink the way the upstream settings are defined.

## Describe alternatives you've considered

We can put all of the required settings to the resolver level, but it will require a lot of duplication in logic for multiple resolver operators and also duplication for settings for resolvers to the same upstream from different fields.

sujeetsr:
**Summary:**  
Adds support for multiple named upstreams

**Issue Reference(s):**  
Fixes #729 

**Build & Testing:**

**Checklist:**


amitksingh1490:
Will redisgn this from scratch

PR #749, opened 2023-12-14
Adding response caching

shashitnak:
**Summary:**  
Caching the responses in memory to avoid making redundant network calls

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/blueprint/blueprint.rs
```diff
@@ -133,6 +134,7 @@ pub struct FieldDefinition {
   pub resolver: Option<Expression>,
   pub directives: Vec<Directive>,
   pub description: Option<String>,
+  pub cache: Cache,
```
tusharmath, line 137:
```suggestion
  pub cache: Option<Cache>,
```

src/blueprint/blueprint.rs
```diff
@@ -86,6 +86,7 @@ pub struct ObjectTypeDefinition {
   pub fields: Vec<FieldDefinition>,
   pub description: Option<String>,
   pub implements: BTreeSet<String>,
+  pub cache: Cache,
```
tusharmath, line 89:
```suggestion
  pub cache: Option<Cache>,
```

src/blueprint/into_schema.rs
```diff
@@ -48,9 +75,30 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
+                let mut args_list: Vec<_> = ctx.graphql_ctx.args.iter().collect();
+                args_list.sort_by(|(key1, _), (key2, _)| key1.cmp(key2));
+                args_list.iter().for_each(|(key, value)| {
+                  key.hash(state);
+                  hash_const_value(value.as_value(), state);
+                });
+
+                let key = hasher.finish();
+                let const_value = res_cache.fetch(&ctx, &expr, key).await?;
+
```
tusharmath, line 101: Move the logic to read from cache into a separate function `read_from_cache`

src/graphql/res_cache.rs
```diff
@@ -0,0 +1,116 @@
...
+use ttl_cache::TtlCache;
+
+use crate::config::Cache;
+use crate::lambda::{EvaluationContext, Expression, ResolverContextLike};
+
+const DEFAULT_MAX_AGE: u64 = 30;
+
+#[derive(Clone)]
+pub struct ResCache {
+  cache: Cache,
```
tusharmath, line 14: Remove cache settings from here. Pass it as an argument to `insert`.

src/graphql/res_cache.rs
```diff
@@ -0,0 +1,116 @@
...
+use async_graphql_value::ConstValue;
+use ttl_cache::TtlCache;
+
+use crate::config::Cache;
+use crate::lambda::{EvaluationContext, Expression, ResolverContextLike};
+
+const DEFAULT_MAX_AGE: u64 = 30;
+
+#[derive(Clone)]
+pub struct ResCache {
```
tusharmath, line 13: Move the file to `/src`. `ChronoCache`

src/graphql/res_cache.rs
```diff
@@ -0,0 +1,116 @@
...
+  cache: Cache,
+  data: Arc<RwLock<TtlCache<u64, ConstValue>>>,
```
tusharmath, line 15:
```suggestion
  data: Arc<RwLock<TtlCache<u64, A>>>,
```
Keep data polymorphic.

src/graphql/res_cache.rs
```diff
@@ -0,0 +1,116 @@
...
+    )
+  }
+}
+
+impl ResCache {
+  pub fn new(cache: Cache) -> Self {
+    ResCache { cache, data: Arc::new(RwLock::new(TtlCache::new(1000))) }
+  }
+
+  fn insert(&self, key: u64, value: &ConstValue) -> Option<ConstValue> {
```
tusharmath, line 34:
```suggestion
  fn insert(&self, value: &A) -> Option<bool> where A: Hasher {
```
Could this work?

tests/http/config/caching.graphql
```diff
@@ -0,0 +1,20 @@
...
+  id: Int! @cache(maxAge: 15)
+  name: String! @cache(maxAge: 15)
+  bar: Bar @http(path: "/bars", query: [{key: "fooId", value: "{{value.id}}"}], groupBy: ["fooId"])
+}
+
+type Bar @cache(maxAge: 5) {
+  id: Int! @cache(maxAge: 10)
+  fooId: Int! @cache(maxAge: 20)
+  foo: [Foo] @http(path: "/foos", query: [{key: "id", value: "{{value.fooId}}"}], groupBy: ["id"])
+}
```
meskill, line 20: seems like we can specify `@cache` both on types and on fields. How will it work in case we specify also cache here for `foo`, considering it's specified for the type `Foo`?

shashitnak: the `@cache` provided at the type level will take affect for all the fields inside `Foo` but for the field `foo` the value provided at field level will take affect. I'll do some tests and to get a clearer picture of what's happening

src/chrono_cache.rs
```diff
@@ -0,0 +1,89 @@
...
+  #[allow(dead_code)]
+  pub fn remove(&self, key: &K) -> Option<V> {
+    self.data.write().unwrap().remove(key)
+  }
+}
+
+#[cfg(test)]
+mod tests {
+
+  // #[test]
```
meskill, line 48: do we need that tests? Looks like we have a thin wrapper to library that is probably has its tests

src/chrono_cache.rs
```diff
@@ -0,0 +1,89 @@
...
+  pub fn insert(&self, key: K, value: V, ttl: u64) -> Option<V> {
+    let ttl = Duration::from_secs(ttl);
+    self.data.write().unwrap().insert(key, value, ttl)
+  }
+
+  pub fn get(&self, key: &K) -> Option<V> {
+    self.data.read().unwrap().get(key).cloned()
+  }
+
+  #[allow(dead_code)]
```
meskill, line 39: is this still relevant?

src/blueprint/into_schema.rs
```diff
@@ -30,53 +37,144 @@ fn to_type_ref(type_of: &Type) -> dynamic::TypeRef {
   }
 }
 
+fn get_cache_key<'a, H: Hasher>(ctx: &'a EvaluationContext<'a, ResolverContext<'a>>, mut hasher: H) -> u64 {
```
meskill, line 40: will be great if we have some kind of tests to make sure there is no collision in cache keys for different configuration since there is much going on here.

ci-benchmark/benchmark.graphql
```diff
@@ -4,7 +4,7 @@ schema
   query: Query
 }
 
-type Query {
+type Query @cache(maxAge: 80) {
```
tusharmath, line 7: should be in milli seconds.

src/blueprint/into_schema.rs
```diff
@@ -30,14 +35,46 @@ fn to_type_ref(type_of: &Type) -> dynamic::TypeRef {
...
+  state.finish()
+}
+
 fn to_type(def: &Definition) -> dynamic::Type {
   match def {
     Definition::ObjectTypeDefinition(def) => {
       let mut object = dynamic::Object::new(def.name.clone());
+      let mut hasher = DefaultHasher::new();
+      // Hash on type name
+      def.name.hash(&mut hasher);
```
tusharmath, line 69: object name's hash can be pre-computed into the hasher.

src/blueprint/blueprint.rs
```diff
@@ -86,6 +86,7 @@ pub struct ObjectTypeDefinition {
   pub fields: Vec<FieldDefinition>,
   pub description: Option<String>,
   pub implements: BTreeSet<String>,
+  pub cache: Option<Cache>,
```
tusharmath, line 89: Create a `blueprint::Cache`. 
- Use it **only on fields** and not on types.
- Cache should also have a default hasher inside it.

```rust

#[derive(Clone, Debug)]
pub struct Cache {
  pub max_age: NonZeroU64,
  pub hasher: DefaultHasher,
}
```

tusharmath, line 89:
```suggestion
```

Keep cache information in fields only.

src/blueprint/into_schema.rs
```diff
@@ -30,14 +35,46 @@ fn to_type_ref(type_of: &Type) -> dynamic::TypeRef {
...
 
+fn get_cache_key<'a, H: Hasher>(ctx: &'a EvaluationContext<'a, ResolverContext<'a>>, mut hasher: H) -> u64 {
+  let state = &mut hasher;
+  // Hash on parent value
+  if let Some(const_value) = ctx
+    .graphql_ctx
+    .parent_value
+    .as_value()
+    // TODO: handle _id, id, or any field that has @key on it.
+    .and_then(|data| data.get_key("id"))
```
tusharmath, line 46: - [ ] Should keep going up till you keep finding a parent.
- [ ] What happens if `id` has not been queried?

src/blueprint/into_schema.rs
shashitnak: created a method in `RequestContext` named `cache_read` for reading from cache since it holds the `cache` object now

src/graphql/res_cache.rs
shashitnak: made the `value` generic but the return type is same as the type of `value` since the `TtlCache` library returns the previously existing value, so I'm putting the responsibility on the caller to use or ignore the value

src/blueprint/into_schema.rs
shashitnak: Not handling the first case as of now. For the second one, if parent's `id` isn't queried then not caching the value

src/chrono_cache.rs
shashitnak: removed this part

src/chrono_cache.rs
shashitnak: removed the tests

src/config/config.rs
```diff
@@ -165,6 +167,12 @@ impl Type {
   }
 }
 
+#[derive(Clone, Debug, Default, PartialEq, Deserialize, Serialize, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct Cache {
+  pub max_age: Option<u64>,
```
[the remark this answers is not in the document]
shashitnak: changed `max_age` to type `NonZeroU64`

amitksingh1490 (approved):

PR #753, opened 2023-12-15
feature: add `--out` option to `check` command

### LINKED ISSUE #752: Add an `--out` param to the `check` command
The `check` command currently composes multiple files into one configuration and then validates the composed file.

We want to extend this capability to also output the composed file. 

**Example usage**

```bash
tailcall check a.yml b.graphql c.json --out=abc.graphql
```
In the above example because the output file's extension is `.graphql` we can print the generated configuration in .graphql format.

**Other Usage:**
This command can also be used to convert file from one format to another. 

```bash
tailcall check a.yml --out=a.graphql
```

**Additional Note**
This will turn out quite handy when we have many files that we wish to compose and review.

A-N-uraag:
**Summary:**  
Added --out option to `check` command to write config to an output file

**Issue Reference(s):**  
Fixes #752 
/claim #752

**Build & Testing:**

**Checklist:**


src/cli/tc.rs
```diff
@@ -111,3 +114,19 @@ fn display_config(config: &Config, n_plus_one_queries: bool) {
...
+  let contents = match filetype {
+    Source::GraphQL => config.to_sdl(),
+    Source::Json => config.to_json().unwrap_or_default(),
+    Source::Yml => config.to_yaml().unwrap_or_default(),
+  };
+
+  Fmt::display(Fmt::success(
+    &format!("Schema has been written to {}", filename).to_string(),
+  ));
+  fs::write(filename, contents).unwrap();
```
ologbonowiwi, line 130:
```suggestion
  fs::write(filename, contents).unwrap();
  Fmt::display(Fmt::success(
    &format!("Schema has been written to {}", filename).to_string(),
  ));
```

ologbonowiwi (changes requested):
IMO we should use `.expect` instead of `.unwrap` to at least give a message on the errors. 

Using `?` operator can also work.

src/cli/tc.rs
A-N-uraag: Ah, yes. Missed the causality😅. Fixing it. Thanks.

src/cli/command.rs
```diff
@@ -42,6 +42,10 @@ pub enum Command {
     /// Display schema
     #[arg(short, long)]
     schema: bool,
+
+    /// Name of the file to save configuration in
```
tusharmath, line 46:
```suggestion
    /// Path of the file to save final configuration in
```

src/cli/tc.rs
```diff
@@ -111,3 +114,19 @@ fn display_config(config: &Config, n_plus_one_queries: bool) {
   let seq = vec![Fmt::n_plus_one_data(n_plus_one_queries, config)];
   Fmt::display(Fmt::table(seq));
 }
+
+fn write_output_file(filename: &String, config: &Config) -> Result<()> {
+  let filetype = Source::detect(filename)?;
+
+  let contents = match filetype {
```
tusharmath, line 121: checkout `source.rs` it has a lot of helper methods that can simplify this code.

src/cli/tc.rs
```diff
@@ -42,6 +42,9 @@ pub fn run() -> Result<()> {
           if schema {
             display_schema(&blueprint);
           }
+          if let Some(out_file) = out_file_name {
+            write_output_file(&out_file, &config)?;
```
tusharmath, line 46: Like `config::Reader` we should add `config::Writer` that can write the file.

A-N-uraag: Hi @tusharmath, added config writer. I hope this is what you were expecting?

src/cli/tc.rs
A-N-uraag: Hi @tusharmath, I have already used the Detect method from there. I could not see any other methods that could be used from there. Please let me know if there is something you see that could be re-used. Thanks.

src/config/config.rs
```diff
@@ -123,6 +124,12 @@ impl Config {
 
     Self { server, upstream, types, schema, unions }
   }
+
+  pub async fn to_file(self, filename: &String) -> Result<()> {
```
tusharmath, line 128:
```suggestion
  pub async fn write(self, filename: &String) -> Result<()> {
```

tusharmath:
Tried to create `jsonplaceholder.yml` and `jsonplaceholder.json` from `jsonplaceholder.graphql` file. It worked fine but a lot of default values were inserted so had to make some change to clean that up a little bit.

tusharmath:
@A-N-uraag Can you raise a PR in  https://github.com/tailcallhq/tailcallhq.github.io to update documentation also.

tusharmath (approved):

A-N-uraag:
Hi @tusharmath, here is the doc update PR: https://github.com/tailcallhq/tailcallhq.github.io/pull/59

PR #762, opened 2023-12-18
chore: compare micro benchmarks

alankritdabral:
**Summary:**  
Compare Criterion benchmarks with the cache_benchmarks 
Added instructions to compare the benchmarks locally
run:
`rust-script benches/criterion_compare <benchmarks1.json> <benchmarks2.json> check` to check the degradation in benchmarks
`rust-script benches/criterion_compare <benchmarks1.json> <benchmarks2.json> table` to get the comparison table
`rust-script benches/criterion_compare <benchmarks1.json> <benchmarks2.json> ` to get the comparison table and then check the degradation.

**Issue Reference(s):**  
Fixes #... _(Replace "..." with the issue number)_

**Build & Testing:**

- [ ] I ran `cargo test` successfully.
- [ ] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [ ] I have added relevant unit & integration tests.
- [ ] I have updated the [documentation website] accordingly.
- [ ] I have performed a self-review of my own code.

[documentation website]: https://github.com/tailcallhq/tailcallhq.github.io

PR #765, opened 2023-12-19
added option in (benches/parse_json_benchmark.rs) to save results in output file 

alankritdabral:
@tusharmath @amitksingh1490 
**Summary:**  
Earlier the benches/parse_json_benchmark.rs used to print the benchmarks directly using `pretty table `
Now it doesnt uses pretty table and we can print the benchmarks directly to any file we want

**Issue Reference(s):**  
#762

**Build & Testing:**

**Checklist:**


alankritdabral:
Why we need  this change:|
Earlier we want to see if the benchmarks are being printed by parsing the json benchmark, so i have use pretty table to  print on the CLI
But now wee need to print that on Pr with every commit and the pretty table does not look great on pr comment so i have made this edit in the script file

PR #776, opened 2023-12-22
feat: type checking for composition with `@graphQL` 

### LINKED ISSUE #756: Type checking for GraphQL composition
This is an invalid schema, `wrongNode` is marked as returning a `User` but the resolver makes a call to `posts`, which doesn't return `User`. This instead fails at run time.

```graphql
type Query {
  wrongNode: User @graphQL(name: "posts")
}
```
Error:
```json
{
  "data": null,
  "errors": [
    {
      "message": "IOException: Request error: HTTP status client error (404 Not Found) for url (http://jsonplaceholder.typicode.com/)",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```

`tc check` should perform some type checking to avoid these issues. Maybe add type checks to `validate_field_has_resolver` as well.
https://github.com/tailcallhq/tailcall/blob/5c479a955b06e654e46f145403c91fb4a0c48db2/src/blueprint/from_config/schema.rs#L32

mayant15:
**Summary:**  
Adds basic type-checking for `@graphQL` directives, given an upstream schema.

This directive fetches data and composes data from a given upstream endpoint, which is why we need a schema for what tailcall *receives* in addition to what tailcall *returns*. Upstream schema can be generated through an introspection query, but for this PR for now I ask the user to define it too. If a schema is not defined type-checking is skipped.

**Issue Reference(s):**  
Fixes #756 

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/blueprint/from_config/from_config.rs
+++ b/src/blueprint/from_config/from_config.rs
@@ -3,6 +3,7 @@ use std::collections::{BTreeMap, HashMap};
 use super::{Server, TypeLike};
 use crate::blueprint::compress::compress;
 use crate::blueprint::*;
+use crate::blueprint::from_config::upstream;
 use crate::config::{Arg, Batch, Config, Field};
 use crate::json::JsonSchema;
 use crate::lambda::{Expression, Unsafe};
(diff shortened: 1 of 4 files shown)


src/blueprint/from_config/from_config.rs
```diff
@@ -29,14 +29,86 @@ pub fn config_blueprint<'a>() -> TryFold<'a, Config, Blueprint, String> {
...
+        }
+      }
+      _ => Valid::succeed(()), // TODO: what should happen for other cases?
+    }
+  } else {
+    log::warn!("No schema provided for upstream, skipping type check for @graphQL");
+    Valid::succeed(())
+  }
+}
+
```
tusharmath, line 111: shouldn't we add graphql validations in `blueprint::from_config::operators::graphql`?

src/blueprint/from_config/from_config.rs
```diff
@@ -29,14 +29,86 @@ pub fn config_blueprint<'a>() -> TryFold<'a, Config, Blueprint, String> {
...
   server
     .and(schema)
     .and(definitions)
     .and(upstream)
+    .and(typecheck)
     .update(apply_batching)
     .update(compress)
 }
 
+fn typecheck_definition(def: &Definition, blueprint: &Blueprint) -> Valid<(), String> {
```
tusharmath, line 51: Should add tests in `test/graphql`

meskill:
Thank you for the pr.

Could you describe a little more about the approach you've chosen? I see that we need to define schema once again for the upstream, but I don't quite get what the difference between this and original schema? Graphql schema as is quite describes the shape of data that is loaded already. 

Put some thoughts into the issue about this

src/blueprint/from_config/from_config.rs
mayant15: Those validations run field-by-field, but for type-checking we need the context of all types defined in the schema, which isn't available till all fields are done parsing

mayant15:
Hey @meskill and @tusharmath thanks for the comments! I've updated the PR description and the original issue with more details.

PR #787, opened 2023-12-26
fix: open localhost in browser in case of using unspecified host

meskill:
**Summary:**  

Fix the problem when in config defined `hostname: "0.0.0.0"` and therefore graphql playground shows address `0.0.0.0` in the logs and open the browser on zeros ip. Instead of using broken address replace unspecified address with localhost.

**Build & Testing:**

**Checklist:**


tusharmath:
I think this was done because we were not able to deploy on `fly` without it. Maybe we should relook at that too.

tusharmath (approved):

PR #800, opened 2023-12-27
feat: set the log level using env variable

### LINKED ISSUE #791: Log Level should be passed as an environment variable
**Problem:**
Currently log-level can only be specified in the `start` command via the `--log-level` parameter. There is no way to configure the log level in other commands. 

**Proposal**
Instead of a cli param, we should set the log level using environment variables. 

```bash
# Option 1
TAILCALL_LOG_LEVEL=DEBUG tailcall start ./main.graphql

# Option 2
TC_LOG_LEVEL=DEBUG tailcall start ./main.graphql
```

**Technical Requirements**
- Log level should be read from `TAILCALL_LOG_LEVEL` and then from `TC_LOG_LEVEL`.
- Drop the `--log-level` from the `start` command

adelinaenache:
**Summary:**  
Use either `TAILCALL_LOG_LEVEL` or `TC_LOG_LEVEL` as a place to set the log level. 

**Issue Reference(s):**  
Fixes #791 
/claim #791 

**Build & Testing:**

**Checklist:**

---- there seems to be no tests of the log level, so nothing to be updated
----- Can't find where/how. Let me know where I can make this change, if I can 

### DIFF
--- a/src/cli/tc.rs
+++ b/src/cli/tc.rs
@@ -18,20 +18,8 @@ use crate::print_schema;
 
 pub fn run() -> Result<()> {
   let cli = Cli::parse();
-  // set the log level
-  const LONG_ENV_FILTER_VAR_NAME: &str = "TAILCALL_LOG_LEVEL";
-  const SHORT_ENV_FILTER_VAR_NAME: &str = "TC_LOG_LEVEL";
 
-  // Check if TAILCALL_LOG_LEVEL is defined, otherwise use TC_LOG_LEVEL
-
-  let filter_env_name = env::var(LONG_ENV_FILTER_VAR_NAME)
-    .map(|_| LONG_ENV_FILTER_VAR_NAME)
-    .unwrap_or_else(|_| SHORT_ENV_FILTER_VAR_NAME);
-
-  // use the log level from the env if there is one, othwerise use the default.
-  let env = Env::new().filter_or(filter_env_name, "info");
-
-  env_logger::Builder::from_env(env).init();
+  logger_init();
 
   match cli.command {
     Command::Start { file_paths } => {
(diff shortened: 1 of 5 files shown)


src/cli/tc.rs
```diff
@@ -20,10 +20,20 @@ pub fn run() -> Result<()> {
...
-    Command::Start { file_paths, log_level } => {
-      env_logger::Builder::new()
-        .filter_level(log_level.unwrap_or(Level::Info).to_level_filter())
-        .init();
+    Command::Start { file_paths } => {
+      let long_env_filter_var_name = "TAILCALL_LOG_LEVEL";
+      let short_env_filter_var_name = "TC_LOG_LEVEL";
+
+      // Check if TAILCALL_LOG_LEVEL is defined, otherwise use TC_LOG_LEVEL
+      let filter_env_name = match env::var(long_env_filter_var_name) {
```
tusharmath, line 28: You can use `or_else` operator to make this simpler.

src/cli/tc.rs
```diff
@@ -20,10 +20,20 @@ pub fn run() -> Result<()> {
...
+        Ok(_) => long_env_filter_var_name,
+        Err(_) => short_env_filter_var_name,
+      };
+
+      // use the log level from the env if there is one, othwerise use the default.
+      let env = Env::new().filter_or(filter_env_name, "info");
+
+      env_logger::Builder::from_env(env).init();
```
tusharmath, line 36: The level is still being set only in the START command. We should set this for all commands. Let's move this logic up, before we match on the command.

src/cli/tc.rs
```diff
@@ -18,12 +18,24 @@ use crate::print_schema;
 
 pub fn run() -> Result<()> {
   let cli = Cli::parse();
+  // set the log level
+  let long_env_filter_var_name = "TAILCALL_LOG_LEVEL";
+  let short_env_filter_var_name = "TC_LOG_LEVEL";
```
tusharmath, line 23: Define the env variables as const.

src/cli/tc.rs
```diff
@@ -18,12 +18,24 @@ use crate::print_schema;
...
+
+  // Check if TAILCALL_LOG_LEVEL is defined, otherwise use TC_LOG_LEVEL
+  let filter_env_name = if env::var(long_env_filter_var_name).is_ok() {
+    long_env_filter_var_name
+  } else {
+    short_env_filter_var_name
```
tusharmath, line 29: We should use the operators that are available on Result and Option here to make this concise.

src/cli/tc.rs
```diff
@@ -18,12 +18,23 @@ use crate::print_schema;
...
+  // Check if TAILCALL_LOG_LEVEL is defined, otherwise use TC_LOG_LEVEL
+
+  let filter_env_name = env::var(LONG_ENV_FILTER_VAR_NAME)
+    .map(|_| LONG_ENV_FILTER_VAR_NAME)
+    .unwrap_or_else(|_| SHORT_ENV_FILTER_VAR_NAME);
+
+  // use the log level from the env if there is one, othwerise use the default.
+  let env = Env::new().filter_or(filter_env_name, "info");
+
+  env_logger::Builder::from_env(env).init();
```
tusharmath, line 34: Can we create a function `logger_init()` and move this logic there.

tusharmath (approved):

PR #818, opened 2023-12-28
feat: adding rate limiter at entity level

shashitnak:
**Summary:**  
Adding the functionality for Rate Limiting at the field level

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/http/server_context.rs
```diff
@@ -23,6 +25,7 @@ pub struct ServerContext {
   pub gql_data_loaders: Arc<Vec<DataLoader<DataLoaderRequest, GraphqlDataLoader>>>,
   pub cache: ChronoCache<u64, ConstValue>,
   pub grpc_data_loaders: Arc<Vec<DataLoader<grpc::DataLoaderRequest, GrpcDataLoader>>>,
+  pub num_requests_fetched: Arc<Mutex<HashMap<String, NumRequestsFetched>>>,
```
tusharmath, line 28: Use a trait that does rate limit. That looks something like —

```rust
trait RateLimiter<K> {
  async fn allow(key: K) -> Result<Stats, Error>
}

struct Stats {
  bucket_size: usize,
  filled: usize,
}

```

shashitnak: Instead of a trait, I have created a `struct RateLimiter` for now, since in this change there is only one type of rate limiter, once another one is added I will extract out the common elements from both in a `trait` as you have suggested

ssddOnTop:
Can you resolve the conflicts

tusharmath:
Changing the scope of rate limits to upstream only and not on entities.

PR #821, opened 2023-12-28
docs: logging

### LINKED ISSUE #809: Write a guide on Logging
Write an engaging document on the new logging feature that's been added to tailcall. Ensure that there is enough depth in the content. The current document isn't detailed enough.

**Requirements**
- Update the doc written here https://github.com/tailcallhq/tailcall/blob/main/docs/guides/logging.md
- Add examples of how it can be used.
- Use [excalidraw](https://excalidraw.com/) if you need to create images to explain something.

melsonic:
**Summary:**  
In this PR, detailed documentation on logging is added. I have added information about different log levels and how to configure them. I have also added one diagram.

**Issue Reference(s):**  
Fixes #809
Claim #809 


tusharmath:
Just pass the doc thru grammarly. There are a few punctuation issues. 

A image can be added, but it should be from Tailcall. How about a hero image?

tusharmath:
I like the overall content. It could though emphasize a little bit on how Tailcall leverages these levels to improve devex.

melsonic:
> Just pass the doc thru grammarly. There are a few punctuation issues.
> 
> A image can be added, but it should be from Tailcall. How about a hero image?

Sure, I will correct those punctuation errors.
And for the image part, I have already added one called logging.png. Can you check that one and let me know if that's ok?

tusharmath:
@melsonic Yes, the image is good. We generally design artwork in `canvas.excalidraw`. Would be even better if you could use that though I would not mind the author take some creative liberty here :-)

melsonic:
> @melsonic Yes, the image is good. We generally design artwork in `canvas.excalidraw`. Would be even better if you could use that though I would not mind the author take some creative liberty here :-)

I created the image using Excalidraw but was unsure how to directly link it to the document. So, I exported it as a JPG file and added the image in that logging.md file.

I have added the diagram to `canvas.excalidraw` file. I will add that in the next commit. Are there any additional steps I should take to properly link the image?

tusharmath:
I think the image in SVG doesn't get the right fonts. Let's use `png` in this case.

melsonic:
> I think the image in SVG doesn't get the right fonts. Let's use `png` in this case.

Added `png` image :+1: 
Please confirm if something still needs to be changed/added.

tusharmath (approved):

PR #823, opened 2023-12-28
feat: support for reading environment variables

### LINKED ISSUE #790: Support for reading environment variables
In our configuration, wherever we support mustache templates, we should support the ability to read environment variables.

**Example Usage:**
```graphql
  schema 
    @server(vars: [{key: "auth", value: "{{env.AUTH_TOKEN}}"}])
    @upstream(baseURL: "http://{{env.UPSTREAM_HOST_NAME}}:{{env.UPSTREAM_PORT}}") {
  query: Query
}
```

**Technical Requirements:**
- Environment variables should be read at the time of starting the server.
- Tests should be added in `tests/http` in yml format.

ologbonowiwi:
- feat: add `headers` and `env`
- test: write yml test to env feature
- chore: set env values

**Summary:**  

**Issue Reference(s):**  
Fixes #790
/claim #790

**Build & Testing:**

**Checklist:**

- [x] I have updated the [documentation] accordingly. ~I was about to update but appears that we already have it covered [here](https://github.com/tailcallhq/tailcall/blob/main/docs/guides/context.md#env)~


src/path.rs
```diff
@@ -70,6 +70,7 @@ impl<'a, Ctx: ResolverContextLike<'a>> PathString for EvaluationContext<'a, Ctx>
       "args" => convert_value(ctx.arg(tail)?),
       "headers" => ctx.header(tail[0].as_ref()).map(|v| v.into()),
       "vars" => ctx.var(tail[0].as_ref()).map(|v| v.into()),
+      "env" => std::env::var(tail[0].as_ref()).ok().map(|v| v.into()),
```
tusharmath, line 73: Access env via ctx. Which internally accesses it via server context. That will make it easier to test.

src/mustache.rs
```diff
@@ -78,15 +78,19 @@ impl Mustache {
...
+  let alphanumeric_or_underscore = nom::multi::many0(nom::branch::alt((
+    nom::character::complete::alphanumeric1,
+    nom::bytes::complete::tag("_"),
+  )));
+
+  let parser = nom::sequence::tuple((spaces, alpha, alphanumeric_or_underscore, spaces));
+
+  nom::combinator::map(parser, |(_, a, b, _)| {
+    let b: String = b.into_iter().collect();
+    format!("{}{}", a, b)
```
tusharmath, line 92: What are these changes for? Write unit test for moustache separately, instead of depending on http spec for it.

tusharmath (changes requested):

src/mustache.rs
ologbonowiwi: If you use `POST_ID,` the mustache will return an error or only the `POST` name without these changes. I needed to parse `alphanumeric1` merged with the tag `_`, wrapped into a `many0` (which selects 0 or many of alphanumerics OR `_` character.

ologbonowiwi: Tests added.

src/path.rs
ologbonowiwi: done.

src/lambda/resolver_context_like.rs
```diff
@@ -8,6 +8,9 @@ pub trait ResolverContextLike<'a> {
   fn args(&'a self) -> Option<&'a IndexMap<Name, Value>>;
   fn field(&'a self) -> Option<SelectionField>;
   fn add_error(&'a self, error: ServerError);
+  fn env_var(&'a self, key: &str) -> Option<String> {
+    std::env::var(key).ok()
+  }
```
tusharmath, line 13: This shouldn't be a default implementation.

tusharmath (changes requested):

src/lambda/evaluation_context.rs
```diff
@@ -49,6 +49,10 @@ impl<'a, Ctx: ResolverContextLike<'a>> EvaluationContext<'a, Ctx> {
     value.to_str().ok()
   }
 
+  pub fn env_var(&self, key: &str) -> Option<String> {
```
tusharmath, line 52:
```suggestion
  pub fn env_var<'a>(&'a self, key: &'a str) -> Option<&'a str> {
```

Can we do this?

tests/http_spec.rs
```diff
@@ -157,12 +161,17 @@ impl HttpSpec {
...
       Source::Json => anyhow::Ok(serde_json::from_str(&contents)?),
       Source::Yml => anyhow::Ok(serde_yaml::from_str(&contents)?),
       _ => Err(anyhow!("only json and yaml are supported")),
-    };
-    anyhow::Ok(spec?)
+    }?;
+
+    spec.env.iter().for_each(|(key, value)| {
+      std::env::set_var(key, value);
+    });
```
tusharmath, line 172: We shouldn't need this anymore. Set it `http_spec::run` inside of the server_context.

src/lambda/evaluation_context.rs
ologbonowiwi: image

I see the error when implementing the lifetimes, but I changed the return to `&str` and it worked. I've committed it.

tests/http_spec.rs
ologbonowiwi: done

tests/http/env-value-with-multiple-underscores.yml
```diff
@@ -0,0 +1,54 @@
+---
+name: Env value with underscores
```
meskill, line 2: can we merge all http_spec tests into single one? No need to make many very similar tests

tests/http_spec.rs
```diff
@@ -95,6 +95,9 @@ struct HttpSpec {
   #[serde(default)]
   mock: Vec<Mock>,
 
+  #[serde(default)]
+  env: BTreeMap<String, String>,
```
meskill, line 99: why BTreeMap and not HashMap as it is specified in server_context?

ologbonowiwi: changed it

tests/http/env-value-with-multiple-underscores.yml
ologbonowiwi: done

meskill (approved):

tusharmath (approved):

amitksingh1490 (approved):

PR #833, opened 2023-12-29
refactor: move entity caching into expression

### LINKED ISSUE #801: Create an expression for caching
This is a simple refactoring that needs to be done such that we don't have any caching logic inside of `into_schema`. The implementation in `into_schema` only should call the resolvers that are specified for a field.

**Proposal**
- Create a new `Expression` called `Cache`

    ```rust 
    struct Cache {
      max_age: i32,
      source: Box<Expression>
    }
    ```

- Move all the caching logic inside of `into_schema` to `Expression::eval`

**Technical Requirements**
- Make sure all existing tests pass.
- Make sure there is not degradation in performance.
- Drop `cache` from Blueprint

Shylock-Hg:
**Summary:**  
Move cache logical to expression layer.

**Issue Reference(s):**  
/claim #801 
Close #801 

**Build & Testing:**

**Checklist:**


src/http/server_context.rs
```diff
@@ -117,4 +118,13 @@ impl ServerContext {
...
   }
+
+  fn with_cached(field: &mut blueprint::FieldDefinition) {
+    if let Some(cache) = &field.cache {
+      field.resolver = Some(Expression::Cache(Cache::new(
+        cache.max_age,
+        Box::new(field.resolver.as_ref().unwrap().clone()),
+      )));
+    };
+  }
```
tusharmath, line 129: Move this logic into `from_config::operators::cache`

src/blueprint/into_schema.rs
```diff
@@ -84,29 +48,9 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
-                      let const_value = expr.eval(&ctx).await?;
-                      log::info!("Writing to cache. key = {key}");
-                      // Write value to cache
-                      ctx.req_ctx.cache_insert(key, const_value.clone(), ttl);
-                      const_value
-                    }
-                  }
-                  _ => expr.eval(&ctx).await?,
-                };
-
```
tusharmath, line 109: nice cleanup. `blueprint.rs` also has some remains of unused cache types/logic.

tusharmath: Use config to create it.

src/http/server_context.rs
tusharmath: That shouldn't matter. We shouldn't modify blueprint in server_context.

src/http/server_context.rs
```diff
@@ -121,4 +82,84 @@ impl ServerContext {
...
+  #[allow(clippy::too_many_arguments)]
+  fn check_resolver(
+    field: &mut blueprint::FieldDefinition,
+    expr: &Expression,
+    bt: Batch,
+    universal_http_client: Arc<dyn HttpClient>,
+    http2_only_client: Arc<dyn HttpClient>,
+    http_data_loaders: &mut Vec<DataLoader<DataLoaderRequest, HttpDataLoader>>,
+    gql_data_loaders: &mut Vec<DataLoader<DataLoaderRequest, GraphqlDataLoader>>,
+    grpc_data_loaders: &mut Vec<DataLoader<grpc::DataLoaderRequest, GrpcDataLoader>>,
```
meskill, line 95: instead of passing all that options to the function consider this:
- make this function associated fn that accepts self
- self will have most of that options already
- in new functions first create the ServerContext then call this function to validate

src/http/server_context.rs
```diff
@@ -121,4 +82,84 @@ impl ServerContext {
...
+        cache.source(),
+        bt,
+        universal_http_client,
+        http2_only_client,
+        http_data_loaders,
+        gql_data_loaders,
+        grpc_data_loaders,
+      );
+      field.resolver = Some(Expression::Cache(Cache::new(cache.max_age(), Box::new(new_expr))));
+      field.resolver.clone().unwrap()
```
meskill, line 160: I think the call for `check_resolver` will already change the resolver `field` so we don't need to do this again. Could you verify this?

src/lambda/expression.rs
```diff
@@ -181,11 +206,76 @@ impl Expression {
             result
           }
         },
+        Expression::Cache(cache) => {
+          if let Some(key) = get_cache_key(ctx, DefaultHasher::new()) {
```
meskill, line 210: I don't think here could be simple use of `DefaultHasher`as we lose a lot of information on parent types. Take a look into `src/blueprint/into_schema.rs` file and how is hasher is used there. We need to separate fields with the same name from different types I believe.

Anyway, we need a help from @shashitnak to verify this for cache

meskill (changes requested):

src/lambda/expression.rs
shashitnak: yes `DefaultHasher::new()` shouldn't be be used here. This hasher needs to be cloned from `cache` field in `FieldDefinition`. It has a `DefaultHasher` which has already hashed the name of the def and the field

meskill: @shashitnak Could you also write the steps that are important for hashing here for reference.
Is it smth like `TypeName -> FieldName -> get_cache_key`?

shashitnak: yes, that is correct. And `get_cache_key` then hashes on the value of `id` field if present in the parent of field and on the name and value of arguments of the query

src/http/server_context.rs
```diff
@@ -119,6 +61,92 @@ impl ServerContext {
...
+          }
+        }
+      }
+    }
+    ctx.schema = bp.to_schema();
+    ctx.blueprint = bp;
+    ctx
+  }
+
+  fn check_resolver_of_field(&mut self, field: &blueprint::FieldDefinition, expr: &Expression) -> Option<Expression> {
```
meskill, line 80: I think it could accept `&mut FieldDefinition` and the field.resolver could be just updated inside the function instead of returning new expr

src/http/server_context.rs
```diff
@@ -119,6 +61,92 @@ impl ServerContext {
...
+    let bt = self.blueprint.upstream.batch.clone().unwrap_or_default();
+    let universal_http_client = self.universal_http_client.clone();
+    let http2_only_client = self.http2_only_client.clone();
+    if let Expression::Unsafe(expr_unsafe) = &expr {
```
meskill, line 84: can we split some of the arms to separate functions to make it more readable overall?

meskill (changes requested):

src/http/server_context.rs
meskill: yes, but we will keep this code private inside constructor function to prevent misusage

src/http/server_context.rs
meskill: I can't agree with that because this method is already using mutation with `&mut self` and you can't call this immutable approach. 

I'd like to use consistent approach. If you still thinks current fn is better feel free to leave it as is

meskill: Actually, I found an error in current code:

in `check_unsafe_expr` the last wildcard match arm returns None which will remove the existing resolver on this line `field.resolver = ctx.check_resolver_of_field(field, expr)`

That's easy to make such kind of mistakes when rewriting from mutable to immutable version of code =)

src/http/server_context.rs
```diff
@@ -119,6 +61,91 @@ impl ServerContext {
...
+
+  fn check_resolver_of_field(&mut self, field: &blueprint::FieldDefinition, expr: &Expression) -> Option<Expression> {
+    if let Expression::Unsafe(expr_unsafe) = &expr {
+      self.check_unsafe_expr(expr_unsafe, field)
+    } else if let Expression::Cache(cache) = &expr {
+      let new_expr = self.check_resolver_of_field(field, cache.source());
+      new_expr.map(|ne| Expression::Cache(Cache::new(cache.hasher().clone(), cache.max_age(), Box::new(ne))))
+    } else {
+      Some(expr.clone())
+    }
```
meskill, line 88: `match` is more suitable for that kind of conditions

meskill (changes requested):

src/lambda/expression.rs
meskill: @shashitnak tests were passed for this. Could you check the latest changes of the pr to verify it's ok? or maybe you can suggest how to test it properly and what kind of test should be added to catch this?

shashitnak: This is a test which can catch this bug. I have tested it with the caching implementation in the main branch and it fails when I provide `DefaultHasher::new()` to `get_cache_key` and passes when I pass the hasher from the `Cache` object.

`.graphql` file
```
schema @upstream(baseURL: "http://example.com", batch: {delay: 1, maxSize: 1000}) {
  query: Query
}

type Query {
  foo: Foo @http(path: "/foo")
}

type Foo {
  id: Int!
  bar1: Bar @http(path: "/bar1") @cache(maxAge: 1000)
  bar2: Bar @http(path: "/bar2") @cache(maxAge: 1000)
}

type Bar {
  id: Int!
}
```

`.yml` file
```
config: !file tests/http/config/caching-gql-collision.graphql
name: Caching Graphql args
mock:
  - request:
      url: http://example.com/foo
    response:
      body:
        id: 0
  - request:
      url: http://example.com/bar1
    response:
      body:
        id: 1
  - request:
      url: http://example.com/bar2
    response:
      body:
        id: 2
assert:
  - request:
      method: POST
      url: http://localhost:8080/graphql
      body:
        query: "query { foo { id bar1 { id } bar2 { id } } }"
    response:
      body:
        data:
          foo:
            id: 0
            bar1:
              id: 1
            bar2:
              id: 2
```
The idea is since name of the type and the field is not used, `bar1` and `bar2` result in the same hash and the result of `bar1` is returned for `bar2` from the cache, hence the test fails.

@Shylock-Hg Please add this in the tests. If you find any mistake in the test or if it is not catching this bug correctly let me know

meskill: I've extended the test file with more check, please, take a look https://github.com/tailcallhq/tailcall/pull/833/files#diff-4ae3c7a1b1832e15a94d554d6610f6b51427fff9c090ef7ab77b7c8f2172406b

src/http/server_context.rs
```diff
@@ -123,6 +65,94 @@ impl ServerContext {
...
+      Expression::Unsafe(expr_unsafe) => self.check_unsafe_expr(expr_unsafe, field),
+      Expression::Cache(cache) => {
+        self.check_resolver_of_field(field, cache.source());
+        field.resolver = field.resolver.as_ref().map(|ne| {
+          Expression::Cache(Cache::new(
+            cache.hasher().clone(),
+            cache.max_age(),
+            Box::new(ne.clone()),
+          ))
+        });
```
meskill, line 95: do we need to do this? it look like a clone for cache expression without any changes

meskill: check this again, here is definitely something wrong. The `ne` here is already a Cache instance leading to wrapping cache inside cache. You can check that by putting debug logs here or inside the expression's eval for cache

src/lambda/expression.rs
meskill: I didn't get the question, could you rephrase it?

meskill: ok, got it. 

Since that test file didn't catch the problem during working on that pr I have doubts that it is useful at all. Also, that was a big file that was hard to read and understand what is going on so it decrease its value. New test catches several cases and it is useful already at least

tusharmath:
@Shylock-Hg can you resolve conflicts. I closed the PR by mistake.

shashitnak:
@Shylock-Hg can you please resolve the conflicts

tusharmath:
@Shylock-Hg We are reimplementing this feature here in #1078 with a few more optimizations and design changes. Will close this for now. Thank you for your time and effort 🙌

PR #837, opened 2023-12-29
fix: enhance` jsonSchema` to support recursive schema definitions

### LINKED ISSUE #758: Stackoverflow Errors on recursive types
Across the Config to Blueprint compiler the current implementation leverages recursion, which breaks  whenever there are recursive types. There is no one strategy we can use everywhere to fix this.

**Example**

```graphql
schema {
  query: Query
}

type Query {
  list: List
}

type List {
  head: Int!
  tail: List
}
```

```bash
tailcall start ./app.graphql
```

**Additional Information**
- JsonSchema might need a bit of re-write. It doesn't support recursive types at the moment.

A-N-uraag:
**Summary:**  
Fixed stackOverflow error during JsonSchema construction and validation

**Issue Reference(s):**  
Fixes #758 - Part 2/2
/claim #758 

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/json/json_schema.rs
+++ b/src/json/json_schema.rs
@@ -11,6 +11,7 @@ pub enum JsonSchema {
   Obj(HashMap<String, JsonSchema>),
   Arr(Box<JsonSchema>),
   Opt(Box<JsonSchema>),
+  Ref(String),
   Str,
   Num,
   Bool,
@@ -34,7 +35,11 @@ impl Default for JsonSchema {
 
 impl JsonSchema {
   // TODO: validate `JsonLike` instead of fixing on `async_graphql::Value`
-  pub fn validate(&self, value: &async_graphql::Value) -> Valid<(), &'static str> {
+  pub fn validate(
+    &self,
+    value: &async_graphql::Value,
+    schema_map: &HashMap<String, JsonSchema>,
+  ) -> Valid<(), &'static str> {
     match self {
       JsonSchema::Str => match value {
         async_graphql::Value::String(_) => Valid::succeed(()),
@@ -52,7 +57,7 @@ impl JsonSchema {
         async_graphql::Value::List(list) => {
           // TODO: add unit tests
           Valid::from_iter(list.iter().enumerate(), |(i, item)| {
-            schema.validate(item).trace(i.to_string().as_str())
+            schema.validate(item, schema_map).trace(i.to_string().as_str())
           })
           .unit()
         }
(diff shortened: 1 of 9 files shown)


src/json/json_schema.rs
```diff
@@ -34,7 +35,11 @@ impl Default for JsonSchema {
 
 impl JsonSchema {
   // TODO: validate `JsonLike` instead of fixing on `async_graphql::Value`
-  pub fn validate(&self, value: &async_graphql::Value) -> Valid<(), &'static str> {
+  pub fn validate(
+    &self,
+    value: &async_graphql::Value,
+    schema_map: &HashMap<String, JsonSchema>,
+  ) -> Valid<(), &'static str> {
```
tusharmath, line 42: I think it would be better to match on `value` instead of `self`.

src/json/json_schema.rs
```diff
@@ -11,6 +11,7 @@ pub enum JsonSchema {
   Obj(HashMap<String, JsonSchema>),
   Arr(Box<JsonSchema>),
   Opt(Box<JsonSchema>),
+  Ref(String),
```
tusharmath, line 14: How will Ref work if none of the JsonSchema types have any names?

src/json/json_schema.rs
A-N-uraag: Hi @tusharmath, I had a similar thought but did not proceed in that direction assuming the following: For most cases `Schema Definition` is a subset of `API response` i.e response will contain way more fields than defined in the schema. In that case, it might be better to check if the defined fields satisfy the schema and leave the rest untouched? I am not very familiar with the usage patterns of graphQL, so please correct me if that's not the case and I will adjust it accordingly. Thanks.

src/json/json_schema.rs
A-N-uraag: We use `Ref` only to refer to the `Obj` types which were referred again recursively. All the `Obj` types need to have a name for it to be defined recursively. So, there would always exist a type name. Is there any corner case I missed? Thanks.

src/json/json_schema.rs
tusharmath: It could happen both ways. The types could have more fields than the actual value. I think matching over the value will simplify the logic quite a lot and drop the usage of the HashMap

src/json/json_schema.rs
tusharmath: How about we clean this up a little bit as follows:

```rust
struct Name(String);
struct TypeDef(HashMap<Name, Definition>);
pub enum Definition {
  Obj(HashMap<String, Definition>),
  Arr(Box<Definition>),
  Opt(Box<Definition>),
  Str,
  Num,
  Bool,
  Ref(Name)
}

impl TypeDef {
  pub fn new() -> Self {todo!()}
  pub fn combine(self, other: TypeDefs) -> Self { todo!() }
  pub fn insert(self, def: Definition) -> Self {todo!()}
  pub fn validate(&self, value: &async_graphql::Value) -> Valid<(), &'static str> {todo!()}
}
```

Now validation is done on TypeDef which has access to set of types and can access by name.

src/json/json_schema.rs
A-N-uraag: Ah, ok. Sure. I will try that approach.

A-N-uraag:
Working on feedback received and gRPC changes.

PR #840, opened 2023-12-29
fix: disable batching by default

amitksingh1490:


ci-benchmark/benchmark.graphql
```diff
@@ -1,6 +1,6 @@
 schema
   @server(port: 8000, graphiql: true, hostname: "0.0.0.0")
-  @upstream(baseURL: "http://jsonplaceholder.typicode.com", httpCache: true, batch: {delay: 1}) {
+  @upstream(baseURL: "http://jsonplaceholder.typicode.com", httpCache: true, batch: {delay: 0}) {
```
tusharmath, line 3: We should do this in our codebase.

PR #842, opened 2023-12-29
feat(http): added support for http2 without ssl

### LINKED ISSUE #812: Support HTTP2 without SSL
Currently, tailcall's HTTP2 server only works with SSL. We wish to change this so that HTTP2 can work without SSL. As per the HTTP2 specification, SSL isn't required for HTTP2 to work.

Rutik7066:
**Summary:**  
SSL for http2 is now optional in this PR. user can spin up backend without cert and key fields 

**Issue Reference(s):**  
Fixes #812
/claim #812

**Build & Testing:**

**Checklist:**


src/http/http_2.rs
```diff
@@ -54,46 +54,85 @@ pub async fn start_http_2(
...
+    };
 
-  if let Some(sender) = server_up_sender {
-    sender.send(()).or(Err(anyhow::anyhow!("Failed to send message")))?;
-  }
+    let result = server.map_err(CLIError::from);
 
-  let server: std::prelude::v1::Result<(), hyper::Error> = if sc.blueprint.server.enable_batch_requests {
-    builder.serve(make_svc_batch_req).await
+    Ok(result?)
```
ologbonowiwi, line 93: It appears that this implementation only enables HTTP1 when there is no cert/key rather than enabling HTTP2 without SSL. imo is not the expected implementation

Rutik7066: .http2_only on builder and bind will enable http2 only, in both cases we are on http2 and i am still working on it made some changes in this implementation so this implementation is outdated now

tusharmath:
@Rutik7066 Please reopen the PR when it's in working state.

PR #843, opened 2023-12-30
feat: custom expressions in schemas

### LINKED ISSUE #839: feat: add support for `@expr`
Tailcall needs to perform logical operations while composing APIs. Currently, this isn't possible with `@http` and `@grpc`. Logical operations include performing conditionals, setting default values, handling errors etc. We will create a new operator called `@expr` which will do this for us.

**Proposal**

```graphql
schema @upstream(baseURL: "https://jsonplaceholder.typicode.com") {
  query: Query
}

type Query {
  users: [User] @expr(body: {http: {path: "/users"}})
}
```
Here `expr` just provides operators to be embedded directly in the body. In the above case we are inserting an `http` into the expr.`Expr` basically allows the composing of other operators that already exist to resolve data. The above example is morally equivalent to using `@http` directly:

```graphql
type Query {
}
```

Here is how the `config.rs` would look like
```rust
enum Expr {
  Http(Http), // reuse existing Http operator
  Grpc(Grpc), // reuse existing Grpc operator
  Const(Const), // reuse existing Const operator
}
```

Config to Blueprint should also be updated such that `config::Expr` and be converted into `Expression`.

**Future Work**

This will allow us to define more operators in the future such as conditional operators. eg:

```graphql
type User {
  name: String
  age: Int
  id: Int
  canVote: Boolean
    @expr(
      body: {
        if: {
          cond: { eq: [{ var: "args.age" }, 18] }
          then: { const: true }
          else: { const: false }
        }
      }
    )
}
```

This is how the updated version of `Expr` would look like:

```rust
enum Expr {
  Http(Http),
  Grpc(Grpc),
  Const(Const),
  If { cond: Box<Expr>, then: Box<Expr>, else, Box<Expr> },
  Eq(Box<Expr>, Box<Expr>),  
}
```

This could potentially open up a large number of possible use cases and freedom to define very powerful operators.

mayant15:
**Summary:**  
Adds support for a `@expr` directive, which takes an AST defined as the `config::ExprBody` type. This PR implements support for ASTs with one node - just a call to an existing resolver directive. Future work would add more expressions like comparison operators and branching.

**Issue Reference(s):**  
Fixes #839 
/claim #839

**Build & Testing:**

**Checklist:**


src/config/config.rs
```diff
@@ -387,6 +390,25 @@ pub struct Http {
   pub group_by: Vec<String>,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, Eq)]
+pub enum ExprBody {
+    #[default]
+    None,
```
tusharmath, line 396: Why is `None` required?

mayant15: Removed. I added it to be able to derive `Default`, but that isn't required.

src/config/config.rs
```diff
@@ -387,6 +390,23 @@ pub struct Http {
   pub group_by: Vec<String>,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+pub enum ExprNode {
```
tusharmath, line 394:
```suggestion
pub enum ExprBody {
```

tests/graphql/test-expr-http.graphql
```diff
@@ -0,0 +1,27 @@
...
+}
+
+type User {
+  id: Int
+  name: String
+}
+
+schema {
+  query: Query
+}
```
tusharmath, line 27: Tests need client SDL

src/blueprint/from_config/operators/expr.rs
```diff
@@ -0,0 +1,40 @@
...
+  operation_type: &config::GraphQLOperationType,
+) -> TryFold<'_, (&Config, &Field, &config::Type, &str), FieldDefinition, String> {
+  TryFold::<(&Config, &Field, &config::Type, &str), FieldDefinition, String>::new(
+    |(config, field, ty, name), b_field| {
+      let Some(expr) = &field.expr else {
+        return Valid::succeed(b_field);
+      };
+
+      match &expr.body {
+        ExprNode::Http(http) => {
```
tusharmath, line 17: Should leverage the `operators::http` to compile `config::Http` to `expression::Http`

src/config/config.rs
mayant15: I renamed it to Node because when we do if-expressions we'd need the same enum, something like
```rust
If {
  cond: Box<ExprNode>,
  then: Box<ExprNode>,
  els: Box<ExprNode>
}
```

src/blueprint/from_config/operators/expr.rs
mayant15: I wasn't quite sure what you meant here since I have used `update_http` etc. from `operators::http`. I've refactored the code to split the match and the operator application, is this closer to what you had in mind?

src/config/config.rs
mayant15: Renamed it back to `ExprBody`.

src/blueprint/from_config/operators/expr.rs
mayant15: It gets quite clunky when compiling basic expressions like if, I've tried to implement a sample in #862

tests/graphql/test-expr-http.graphql
```diff
@@ -0,0 +1,27 @@
+#> server-sdl
+schema @server @upstream(baseURL: "http://abc.com") {
+  query: Query
+}
+
+type Query {
+  foo(id: Int!): User @expr(body: {http: {baseURL: "http://jsonplacheholder.typicode.com", path: "/users/{{args.id}}"}})
+}
```
tusharmath, line 8: You can combine all the successful tests into one file `test-expr-success.graphql`.

Add new tests for invalid `@expr` usage with proper validation messages.

src/blueprint/from_config/operators/expr.rs
```diff
@@ -0,0 +1,37 @@
...
+
+      let field_with_expr = setup_field(field, &expr.body);
+      let field_with_resolver = update_http()
+        .trace(config::Http::trace_name().as_str())
+        .and(update_const_field().trace(config::Const::trace_name().as_str()))
+        .and(update_grpc(operation_type).trace(config::Grpc::trace_name().as_str()))
+        .and(update_graphql(operation_type).trace(config::GraphQL::trace_name().as_str()))
+        .try_fold(&(config, &field_with_expr, ty, name), b_field);
+      field_with_resolver
+    },
```
tusharmath, line 35: Ideally we should create a new struct in `lambda::expression::Expression` to handle all `Expr`.

src/blueprint/from_config/operators/expr.rs
```diff
@@ -0,0 +1,37 @@
...
+  let copy = field.clone();
+  match expr {
+    ExprBody::Http(http) => copy.http(http.clone()),
+    ExprBody::Const(const_field) => copy.const_field(const_field.clone()),
+    ExprBody::Grpc(grpc) => copy.grpc(grpc.clone()),
+    ExprBody::GraphQL(graphql) => copy.graphql(graphql.clone()),
+  }
+}
+
+pub fn update_expr(
```
tusharmath, line 18: Missing tests execution tests in `tests/http`

tusharmath:
@mayant15 We had built this feature in Scala using something called `JsonT` https://github.com/tailcallhq/tailcall/blob/main-scala/tailcall/src/main/scala/tailcall/runtime/JsonT.scala

`JsonT` is the moral equivalent of `Expr` sharing the link here, hope it helps!

src/config/config.rs
```diff
@@ -408,7 +428,7 @@ pub struct GraphQL {
   pub name: String,
   #[serde(default, skip_serializing_if = "is_default")]
   pub args: Option<KeyValues>,
-  #[serde(rename = "baseURL")]
+  #[serde(rename = "baseURL", default, skip_serializing_if = "is_default")]
```
ologbonowiwi, line 431: why do we need to change this?

src/blueprint/from_config/operators/expr.rs
```diff
@@ -0,0 +1,37 @@
+use crate::blueprint::*;
+use crate::config;
+use crate::config::{Config, ExprBody, Field};
+use crate::directive::DirectiveCodec;
+use crate::try_fold::TryFold;
+use crate::valid::Valid;
+
+fn setup_field(field: &Field, expr: &ExprBody) -> Field {
+  let copy = field.clone();
```
ologbonowiwi, line 9: here we could shadow this value 
```suggestion
  let field = field.clone();
```

src/config/config.rs
ologbonowiwi: is this a valid case, @tusharmath? I thought every config should have a base URL.

tusharmath: Base url can be specified in http or upstream or both. We only validate it during blueprint creation. It could also be inferred during composition of configs. 
So someone could logically set the base url to empty at both places and then override it from a different file.

mayant15:
Closing this in favour of #862 since that would introduce a more concrete feature (if-expressions)

PR #853, opened 2023-12-31
feat: add cli `compose` command

### LINKED ISSUE #782: Feature add `compose` command
We should create a new compose command that allows us to compose configurations. This feature is already implemented via `check` command.

**Usage for `compose`**
Print the composed configuration on the terminal using a format
```bash
tailcall compose a.yml b.graphql c.json --format=YML
```

**Improvements in `check`**
Passing `--blueprint` should print the blueprint on the CLI.
```bash
tailcall check a.yml b.graphql c.json --blueprint
```

**Additional Requirements**
- Drop the `--out` parameter from `check`

adelinaenache:
**Summary:**  
Implement "compose" command and improve "check" command.
Check command is able to print the blueprint using --`blueprint`
Compose command merges multiple config files and prints them in a specific format (YML/JSON/GQL) set by using --format para

**Issue Reference(s):**  
Fixes #782 
/claim #782 

**Build & Testing:**

**Checklist:**

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,7 @@ prettytable-rs = "^0.10"
 protox = "0.5.1"
 prost-reflect = { version = "0.12.0", features = ["serde"] }
 prost = "0.12.3"
+http-serde = "1.1.3"
 
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,13 +21,13 @@ sha2 = { version = "0.10", features = ["std"] }
 tokio = { version = "1", features = ["full"] }
 reqwest = { version = "0.11", features = [
    "json",
-   "rustls-tls"
+   "rustls-tls",
 ], default-features = false }
 hyper = { version = "0.14", features = ["full"], default-features = false }
 async-graphql = { version = "6.0.11", features = [
    "dynamic-schema",
    "dataloader",
-   "apollo_tracing"
+   "apollo_tracing",
 ] }
 async-graphql-value = "6.0.11"
 base64 = "0.21"
@@ -38,7 +38,7 @@ clap = { version = "4.4.12", features = ["derive"] }
 colored = "2"
 regex = "1.10.2"
 http-cache-reqwest = { version = "0.12.0", features = [
-   "manager-moka"
+   "manager-moka",
 ], default-features = false }
 reqwest-middleware = "0.2.4"
 async-trait = "0.1.77"
@@ -68,7 +68,6 @@ prettytable-rs = "^0.10"
 protox = "0.5.1"
 prost-reflect = { version = "0.12.0", features = ["serde"] }
 prost = "0.12.3"
-http-serde = "1.1.3"
 
 
 [dev-dependencies]
--- a/docs/guides/cli.md
+++ b/docs/guides/cli.md
@@ -66,7 +66,7 @@ tailcall compose ./path/to/file1.graphql ./path/to/file2.graphql
 
 ### --format
 
-This specifies the format of the desired composed file. It accepts `gql`,`yml`, `json` . Default is `json`.
+This specifies the format of the desired composed file. It accepts `gql` or `graphql`,`yml` or `yaml`, `json` . Default is `json`.
 
 ```bash
 tailcall compose ./path/to/file1.graphql ./path/to/file2.graphql --format gql
--- a/docs/guides/cli.md
+++ b/docs/guides/cli.md
@@ -12,7 +12,7 @@ The `check` command validates a composition spec. Notably, this command can dete
 tailcall check [options] <file>...
 ```
 
-The `check` command offers various options that control different settings, such as the display of the blueprint, endpoints, and schema of the composition spec.
+The `check` command offers various options that control different settings, such as the display of the generated schema, n + 1 issues etc.
 
 ### --n-plus-one-queries
 
@@ -42,16 +42,6 @@ Example:
 
 ```bash
 tailcall check --schema ./path/to/file1.graphql ./path/to/file2.graphql
-```
-
-### --blueprint
-
-This option prints the generated blueprint of the composed spec.
-
-Example:
-
-```bash
-tailcall check /path/to/file1.graphql ./path/to/file2.graphql --blueprint
 ```
 
 ## compose
--- a/docs/guides/cli.md
+++ b/docs/guides/cli.md
@@ -44,26 +44,32 @@ Example:
 tailcall check --schema ./path/to/file1.graphql ./path/to/file2.graphql
 ```
 
-### --out
+### --blueprint
 
-This option writes the resulting schema of the composition spec to a file. The value of the option will become the name of the file.
-
-- Type: String
-
-```bash
-tailcall check <file1> <file2> ... <fileN> --out <outfile>
-```
-
-The schema can be written either in `json`, `graphql` or `yml` formats.
+This option prints the generated blueprint of the composed spec.
 
 Example:
 
 ```bash
-# Output .graphql file
-tailcall check ./path/to/file1.graphql ./path/to/file2.graphql --out ./path/to/outfile.graphql
+tailcall check /path/to/file1.graphql ./path/to/file2.graphql --blueprint
+```
 
-# Output .json file
-tailcall check ./path/to/file1.graphql ./path/to/file2.graphql --out ./path/to/outfile.json
+## compose
+
+The `compose` merges multiple configuration files into one. To use the `compose` command, follow this format:
+
+Example:
+
+```bash
+tailcall compose ./path/to/file1.graphql ./path/to/file2.graphql
+```
+
+### --format
+
+This specifies the format of the desired composed file. It accepts `gql`,`yml`, `json` . Default is `json`.
+
+```bash
+tailcall compose ./path/to/file1.graphql ./path/to/file2.graphql --format gql
 ```
 
 ## start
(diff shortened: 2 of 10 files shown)


docs/guides/cli.md
```diff
@@ -44,26 +44,32 @@ Example:
...
+```
 
-# Output .json file
-tailcall check ./path/to/file1.graphql ./path/to/file2.graphql --out ./path/to/outfile.json
+### --format
+
+This specifies the format of the desired composed file. It accepts `gql`,`yml`, `json` . Default is `json`.
+
+```bash
+tailcall compose ./path/to/file1.graphql ./path/to/file2.graphql --format gql
```
tusharmath, line 72: format should take `gql` and `graphql` as input.

src/blueprint/from_config/server.rs
```diff
@@ -22,12 +23,13 @@ pub struct Server {
   pub port: u16,
   pub hostname: IpAddr,
   pub vars: BTreeMap<String, String>,
+  #[serde(skip)]
   pub response_headers: HeaderMap,
```
tusharmath, line 27: We can't skip any field while serializing the blueprint. We would loose out of information this way, and confuse the developer.

src/cli/command.rs
```diff
@@ -18,6 +18,26 @@ pub struct Cli {
...
+
+impl std::str::FromStr for FormatOption {
+  type Err = String;
+
+  fn from_str(s: &str) -> Result<Self, Self::Err> {
+    match s.to_lowercase().as_str() {
+      "json" => Ok(FormatOption::Json),
+      "yml" | "yaml" => Ok(FormatOption::Yml),
+      "graphql" | "gql" => Ok(FormatOption::Gql),
+      _ => Err(format!("Invalid format: {}", s)),
```
tusharmath, line 36:
```suggestion
      _ => Err(format!("Unsupported format: {}", s)),
```

src/blueprint/from_config/server.rs
tusharmath: Consider writing a custom serde function.

src/cli/command.rs
```diff
@@ -41,10 +61,22 @@ pub enum Command {
     #[arg(short, long)]
     schema: bool,
 
-    /// Path of the file to save final configuration in
-    #[arg(short, long("out"))]
-    out_file_path: Option<String>,
+    /// Display the generated blueprint in the console
```
tusharmath, line 64:
```suggestion
    /// Display the generated blueprint on the console
```

src/cli/command.rs
```diff
@@ -41,10 +61,22 @@ pub enum Command {
...
   },
+
+  /// Merge multiple configuration file into one
+  Compose {
+    /// Path for the configuration files separated by spaces if more than one
+    #[arg(required = true)]
+    file_path: Vec<String>,
+
+    /// Format of the result. Accepted values: JSON|YML|GQL. For blueprint option, only YML and JSON are allowed
+    #[clap(short, long, default_value = "json")]
```
tusharmath, line 76:
```suggestion
    #[clap(short, long, default_value = "GQL")]
```

src/cli/command.rs
```diff
@@ -41,10 +61,22 @@ pub enum Command {
...
+    blueprint: bool,
   },
+
+  /// Merge multiple configuration file into one
+  Compose {
+    /// Path for the configuration files separated by spaces if more than one
+    #[arg(required = true)]
+    file_path: Vec<String>,
+
+    /// Format of the result. Accepted values: JSON|YML|GQL. For blueprint option, only YML and JSON are allowed
```
tusharmath, line 75:
```suggestion
    /// Format of the result. Accepted values: JSON|YML|GQL.
```
Compose doesn't support blueprint anymore, right?

src/blueprint/from_config/server.rs
adelinaenache: I used http-serde to serialize the HeaderMap

src/blueprint/blueprint.rs
```diff
@@ -75,65 +81,66 @@ impl Definition {
...
   pub of_type: Type,
   pub default_value: Option<serde_json::Value>,
   pub description: Option<String>,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize)]
 pub struct Cache {
   pub max_age: NonZeroU64,
+  #[serde(skip)]
```
tusharmath, line 139: This is ok to skip. We are moving this logic out of the blueprint anyways.

adelinaenache: @tusharmath you mean the entire Cache or just what I've skipped?

tusharmath: Just what you have. We have another PR where this will be removed.

src/cli/command.rs
```diff
@@ -18,6 +18,26 @@ pub struct Cli {
   pub command: Command,
 }
 
+#[derive(Clone)]
+pub enum FormatOption {
+  Yml,
+  Gql,
+  Json,
+}
```
tusharmath, line 26: Can we use `config::source::Source` here?

src/cli/tc.rs
```diff
@@ -34,27 +34,37 @@ pub fn run() -> Result<()> {
...
     }
+    Command::Compose { file_path, format } => {
+      let config =
+        tokio::runtime::Runtime::new()?.block_on(async { Config::read_from_files(file_path.iter()).await })?;
+
+      Fmt::display(match format {
+        Source::Yml => config.to_yaml()?,
+        Source::GraphQL => config.to_sdl(),
+        Source::Json => config.to_json(true)?,
+      });
```
tusharmath, line 64:
```suggestion
      Fmt::display(source.encode(config));
```

Can we add an `encode` method that encodes the config and hides away the pattern matching? This is done so that we don't need to change code here when we start supporting other formats.


NOTE: duplicate code is found in `ConfigWriter::write` so we should refactor that also to `Source.encode`.

src/config/source.rs
```diff
@@ -15,6 +16,19 @@ const ALL: [Source; 3] = [Source::Json, Source::Yml, Source::GraphQL];
...
+
+  fn from_str(s: &str) -> Result<Self, Self::Err> {
+    match s.to_lowercase().as_str() {
+      "json" => Ok(Source::Json),
+      "yml" | "yaml" => Ok(Source::Yml),
+      "graphql" | "gql" => Ok(Source::GraphQL),
+      _ => Err(format!("Unsupported format: {}", s)),
+    }
+  }
+}
```
tusharmath, line 30: Do we need this? Can't we use `Source::detect`?

Cargo.toml
```diff
@@ -68,6 +68,7 @@ prettytable-rs = "^0.10"
 protox = "0.5.1"
 prost-reflect = { version = "0.12.0", features = ["serde"] }
 prost = "0.12.3"
+http-serde = "1.1.3"
```
tusharmath, line 71: Where is this used?

src/lambda/expression.rs
```diff
@@ -35,25 +35,29 @@ pub enum Expression {
...
     dl_id: Option<DataLoaderId>,
   },
+  #[serde(skip)]
   GraphQLEndpoint {
     req_template: graphql::RequestTemplate,
     field_name: String,
     batch: bool,
     dl_id: Option<DataLoaderId>,
   },
+  #[serde(skip)]
```
tusharmath, line 60: We can't skip anything in that's an expression. We have two options here —
1. Move that non-serializable thing out of Expression, into a blueprint companion object that doesn't need to be serializable. Typically into ServerContext.
2. Implement Serialize/Deserialize into it.

src/blueprint/blueprint.rs
```diff
@@ -17,15 +19,19 @@ use crate::lambda::{Expression, Lambda};
 /// It can only be generated from a valid Config.
 /// It allows us to choose a different GraphQL Backend, without re-writing all orchestration logic.
 /// It's not optimized for REST APIs (yet).
-#[derive(Clone, Debug, Default, Setters)]
+#[derive(Clone, Debug, Default, Setters, Serialize)]
```
tusharmath, line 22: Can add Deserialize also so that we can start the server by pointing directly to a blueprint file.

src/config/source.rs
adelinaenache: Looks like "detect" is using Souce::ends with that checks for an `.yml`, `.graphql` and `.json` (with a dot, because it checks for an extension name). It also doesn't take into account things like "gql" or "yaml". Should I modify it or keep it as it is?

Cargo.toml
adelinaenache: Serialize HeaderMap

src/lambda/expression.rs
adelinaenache: Managed to serialize Grahql and Http from this Enum, what I need to do now is to serialize grpc. The problem is that I need to serialize a `ProtobufOperation` that has 3 external types. I am blocked on this atm. Still looking for a solution.

tusharmath: @meskill is there a way we can implement a custom serde for this ?

meskill: oh, well, no simple way I think. 
The data used inside this is tied to a single `DescriptorPool` that is shared between many structures and use indexes and references to it. So, to proper serialize it we probably need to be able to recreate all that structure i.e. basically parse proto file again and create new ProtobufOperation.

tusharmath: Dropping blueprint serialization from the scope of this PR.

tusharmath (approved):

PR #862, opened 2023-12-31
feat: if-expressions for resolvers

### LINKED ISSUE #839: feat: add support for `@expr`

mayant15:
**Summary:**
Adds support for a `@expr` directive, which takes an AST defined as the `config::ExprBody` type. This PR implements support for ASTs with two kinds of nodes - calls to an existing resolver directive and branching on conditions. Future work could add more expressions.

**Issue Reference(s):**  
Fixes #839 
/claim #839 

**Build & Testing:**

**Checklist:**


docs/operators/expr.md
```diff
@@ -0,0 +1,36 @@
...
+
+Equivalent to [@grpc](#grpc)
+
+### graphQL
+
+Equivalent to [@graphQL](#graphQL)
+
+### if
+
+Allows branching based on conditions
```
tusharmath, line 36: Need better examples of `if`

src/blueprint/from_config/operators/expr.rs
```diff
@@ -0,0 +1,98 @@
...
+
+fn compile(context: &CompilationContext, expr: ExprBody) -> Valid<Expression, String> {
+  if is_effect(&expr) {
+    compile_effect(context, expr)
+  } else {
+    match expr {
+      ExprBody::If { cond: condition, then, els } => {
+        compile_if(CompileIf { context, condition, then, els }).trace("if")
+      }
+      _ => Valid::fail("unsupported expression".to_string()), // unreachable if is_effect is
```
tusharmath, line 80: should confidently panic using `unreachable!("Unsupported expression")`

tests/graphql/test-expr-success.graphql
```diff
@@ -0,0 +1,61 @@
...
+
+type Post {
+  content: String @expr(body: {graphQL: {args: [{key: "id", value: "{{value.id}}"}], name: "postContent"}})
+  id: Int!
+}
+
+type Query {
+  conditional: Post @expr(body: {if: {cond: {const: {data: true}}, else: {http: {path: "/posts/1"}}, then: {http: {path: "/posts/2"}}}})
+  greeting: String @expr(body: {const: {data: "hello from server"}})
+  newsByIdBatch(news: NewsInput!): News! @expr(body: {grpc: {body: "{{args.news}}", groupBy: ["news", "id"], method: "GetMultipleNews", protoPath: "src/grpc/tests/news.proto", service: "NewsService"}})
```
tusharmath, line 28: Do we need such a long example?

.prettierrc
```diff
@@ -11,10 +11,11 @@
         "add-field-many.graphql",
         "inline-many.graphql",
         "inline-many-list.graphql",
-        "test-grpc.graphql"
+        "test-grpc.graphql",
+        "test-expr-success.graphql"
       ],
       "options": {
-        "printWidth": 200
+        "printWidth": 216
```
tusharmath, line 18: One must find creative ways to avoid doing this.

mayant15: imo the test assertions should ignore whitespace instead of doing a straight string comparison. Maybe first minify then compare.

src/blueprint/from_config/operators/expr.rs
mayant15: simplified it entirely into a single type-safe match

examples/.tailcallrc.graphql
```diff
@@ -300,6 +300,66 @@ directive @cache(
...
+input ExprGrpc {
+  service: String!
+  method: String!
+  body: String
+  headers: [KeyValue]
+  protoPath: String
+  groupBy: [String!]
+}
+
+input ExprGraphQL {
```
ologbonowiwi, line 338: we need to add comments (`""" some doc """`) to these also

src/blueprint/from_config/operators/expr.rs
```diff
@@ -0,0 +1,66 @@
...
+  let els = input.els;
+
+  compile(context, *cond)
+    .trace("cond")
+    .map(Box::new)
+    .zip(compile(context, *then).trace("then").map(Box::new))
+    .zip(compile(context, *els).trace("else").map(Box::new))
+    .map(|((cond, then), els)| Expression::If { cond, then, els })
+}
+
```
ologbonowiwi, line 34: do you mind putting this below `compile` fn? it'll improve readability as we usually read code top-to-bottom.

examples/.tailcallrc.graphql
mayant15: There is work on auto-generating this file in #895, including doc comments

ologbonowiwi: Whoever will work on #895, it'll copy the comments that will have there and pass it to the `.rs` equivalent stuff. Since #895 does not include write new docs, it's fair that we add these now.

docs/operators/expr.md
```diff
@@ -0,0 +1,46 @@
+---
```
amitksingh1490, line 1: Drop this I have opened a PR separately as docs is moved back to website repo

tusharmath (approved):

PR #881, opened 2024-01-03
feat: create `@call` operator

### LINKED ISSUE #846: Create a `@call` operator
The `@call` operator will allow our users to call the resolver by name that is specified in the root Query and Mutation. Here is an example of how it would work.

```graphql
schema {
  query: Query
}

type Query {
  posts: [Post] @http(path: "/path")
}

type Post {
  id: Int
  title: String
  userId: Int
  user: User @call(query: "user", args: {id: "{{value.userId}}"})
}
```

The `@call` will simply call the resolver that's set in `Query.user` field with the arguments specified via `args`.

This feature gives us the ability to define functions using the Tailcall DSL, which means we can add some form of reusability and modularity.

**Technical Requirements**
- The function can only be called if the input and output type match exactly. In the above example  `Query::user` returns a `User` and `Post::user` also returns a `User` so it's compatible with its return type. Similarly, we can see that the args are also exactly the same.

ologbonowiwi:
**Issue Reference(s):**  
Fixes #846
/claim #846

**Build & Testing:**

**Checklist:**


src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,59 @@
...
+          .and_then(|field| {
+            if !field.has_resolver() {
+              return Valid::fail(format!("{} field has no resolver", field_name));
+            }
+
+            Valid::succeed(field)
+          })
+        })
+        .and_then(|field| {
+          // TO-DO: parse call.args into a way that `update_http`, `update_grpc` and `update_graphql` can use
```
ologbonowiwi, line 47: can you give a hand on it @tusharmath @amitksingh1490 @meskill?

IMO, most of it is done (besides automated testing), but I need help with the args.

What I have in mind now is to search based on the `field.args` mustache value. So if I have `value.userId`, I'd search for a type with a `userId` and the `field_name` `user` using this `@call` operator. Then I get the type from `userId` and use it to create an `Arg` that'll be set on `b_field`, but I'm not sure if this is the best approach.

ologbonowiwi: got it working :smile:

src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,82 @@
...
+
+              if let Some(http) = field.clone().http.as_mut() {
+                http.path = http.path.replace(format!("args.{}", key).as_str(), value.as_str());
+
+                let field = Field { http: Some(http.clone()), ..field.clone() };
+
+                return Valid::succeed(field);
+              }
+
+              Valid::succeed(field)
```
ologbonowiwi, line 67: the current implementation does not cover `grpc` or `graphql.` do we need to support it?

src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,82 @@
...
+
+      if validate_field_has_resolver(name, field, &config.types).is_succeed() {
+        return Valid::fail(format!(
+          "@call directive is not allowed on field {} because it already has a resolver",
+          name
+        ));
+      }
+
+      let type_and_field = if let Some(mutation) = &call.mutation {
+        Valid::succeed(("Mutation", mutation.as_str()))
```
ologbonowiwi, line 25: theoretically, this works, but I didn't write a test as I did not find a practical way to use it.

If you can provide some examples I'll be happy to include them on automated testing @tusharmath

src/blueprint/from_config/operators/call.rs
ologbonowiwi: if yes, it'd be nice to get some examples (as provided for `@http` query on the original issue), so I can use them to stress it out the implementaiton

src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,79 @@
...
+            }
+          })
+        })
+        .and_then(|(field, field_name, args)| {
+          args.fold(Valid::succeed(field.clone()), |field, (key, value)| {
+            field.and_then(|field| {
+              let value = value.replace("{{", "").replace("}}", "");
+
+              if let Some(http) = field.clone().http.as_mut() {
+                http.path = http.path.replace(format!("args.{}", key).as_str(), value.as_str());
```
meskill, line 57: replacing string parts is too intrusive and brittle. Let's think about some way to be able to override args for existing resolver i.e. some flags or wrappers that will override args resolution

src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,79 @@
...
+              Valid::fail(format!("{} field has no resolver", field_name))
+            }
+          })
+        })
+        .and_then(|(field, field_name, args)| {
+          args.fold(Valid::succeed(field.clone()), |field, (key, value)| {
+            field.and_then(|field| {
+              let value = value.replace("{{", "").replace("}}", "");
+
+              if let Some(http) = field.clone().http.as_mut() {
```
meskill, line 56: we need to think about more general solution not tied to specific resolvers. It could be Expression that represents any resolving operation

src/blueprint/from_config/operators/call.rs
ologbonowiwi: I removed it from the implementation

.github/ISSUE_TEMPLATE/guide.md
ologbonowiwi, line 1: changes on this file were made due to prettier errors

examples/.tailcallrc.graphql
```diff
@@ -300,6 +300,20 @@ directive @cache(
...
+"""
+directive @call(
+  """
+  The name of the field that has the `@http` resolver to be called.
+  """
+  query: String!
+  """
+  The arguments to be replace the `@http` resolver args.
+  """
+  args: [KeyValue]
```
tusharmath, line 314:
```suggestion
  args: JSON
```
Lets keep this a JSON and validate it while creating the blueprint.

src/config/config.rs
```diff
@@ -387,6 +389,13 @@ pub struct Http {
   pub group_by: Vec<String>,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, Eq)]
+pub struct Call {
+  #[serde(default, skip_serializing_if = "is_default")]
+  pub query: Option<String>,
+  pub args: KeyValues,
```
tusharmath, line 396: Args should be a Map.

examples/jsonplaceholder.graphql
```diff
@@ -23,5 +23,5 @@ type Post {
   userId: Int!
   title: String!
   body: String!
-  user: User @http(path: "/users/{{value.userId}}")
+  user: User @call(query: "user", args: [{key: "id", value: "{{value.userId}}"}])
```
tusharmath, line 26: This seems like a recursive call that will never end.

src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,70 @@
...
+      };
+
+      if validate_field_has_resolver(name, field, &config.types).is_succeed() {
+        return Valid::fail(format!(
+          "@call directive is not allowed on field {} because it already has a resolver",
+          name
+        ));
+      }
+
+      Valid::from_option(call.query.clone(), "call must have query".to_string())
```
tusharmath, line 24: Query or Mutation both are supported.

src/blueprint/from_config/operators/call.rs
tusharmath: Do we even need to replace this? I think just cloning the resolves should do the job.

src/blueprint/from_config/operators/call.rs
ologbonowiwi: It was supported first but I ended up deleting it for lack of use case. 

Can you provide a .graphql that would use it?

I've tried to add `@call` in a mutation inside the Mutation type, but that field was not handled by this workflow

examples/jsonplaceholder.graphql
ologbonowiwi: Why? This is the example provided on the issue.

src/blueprint/from_config/operators/call.rs
tusharmath: It would look exactly like the one for `query` except the call operator will use `mutation` instead of query.

examples/jsonplaceholder.graphql
tusharmath: My bad.

src/blueprint/from_config/operators/call.rs
ologbonowiwi: I can have a "nested" mutation?

Something like: insertUser and insertPost, but on User type I add a posts property that will hit the /posts when I trigger insertUser. Does that make sense?

tusharmath: You can, why not. It's just a function call and should be allowed. An example could be you have a list of users to create, but you call createUser mutation one by one using call and expr. I think this way we are getting closer to building a turning complete DSL :) so let's not stop!

ologbonowiwi: but does the current implementation already support it? if not, I agree that it's a valid use case, but it's not part of the issue this PR aims to solve

ologbonowiwi:
I'll wait until we merge #862 to finish this, as a bunch of code done there can be re-used here.

src/blueprint/from_config/operators/call.rs
ologbonowiwi: you mean creating a new variation under `Expression` enum that we'll handle on the resolvers? checking if I got it right.

ologbonowiwi: Is there any way to parse the `HashMap<String, String>` that currently represents `call.args` into a `BTreeMap<String, Arg>` (the type of `field.args`) so we can avoid creating a new expression?

Thinking here what would be the optimal solution to solve this issue. IMO this is one of the few things missing for this PR to be done.

CC: @tusharmath (in case you have any idea on how to solve it)

src/blueprint/from_config/operators/call.rs
ologbonowiwi: agree 100% with you that this solution is at minimun sub-optimal, but I'll resolve this and we can continue the discussion on alternatives on [this comment](https://github.com/tailcallhq/tailcall/pull/881#discussion_r1440691499) responses.

src/blueprint/from_config/operators/call.rs
ologbonowiwi: I've tried to add a use case/test here (without `@call` operator) but does I did not had success doing so.

I first got the message that `Post.user` should be an `input` (rather than a `type`), then after changing it to `UserInput` I got the message that I can't have an input on an output value. I did not tried to add `@http` operator on the `PostInput.user` type, and now I don't have the `.graphql` here anymore, but if you have a valid `.graphql` file that uses nested mutation (with only `@http` operator), I'll do my best to support it on the new `@call` operator.

In case we don't have it, we could create a new issue to add support (potentially to both the plain resolver – `@http`, `@grpc`, etc – and the `@call` operator that we'll introduce on this PR.

ologbonowiwi:
Got it working for both HTTP and GraphQL data sources. It'd be nice to get some test/example on how it could be used on gRPC. I'd say this is like 80% done.

docs/operators/call.md
```diff
@@ -0,0 +1,53 @@
...
+## args
+
+The arguments to be passed to the `@http` resolver. It is optional.
+
+```graphql showLineNumbers
+type Post {
+  userId: Int!
+  user: User @call(query: "user", args: [{key: "id", value: "{{value.userId}}"}])
+}
+```
```
tusharmath, line 53: Move this to https://github.com/tailcallhq/tailcallhq.github.io

examples/.tailcallrc.graphql
```diff
@@ -305,6 +305,20 @@ directive @cache(
   maxAge: Int!
 ) on FIELD_DEFINITION
 
+"""
+The @call operator is used to reference an `@http` operator.
```
tusharmath, line 309: This isn't correct. Refer to the original issue to describe what `call` does.

src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,94 @@
+use crate::blueprint::*;
+use crate::config;
+use crate::config::{Config, Field, GraphQLOperationType};
+use crate::directive::DirectiveCodec;
+// use crate::mustache::Mustache;
+use crate::try_fold::TryFold;
+use crate::valid::Valid;
+// use crate::valid::ValidationError;
```
tusharmath, line 8: commented out imports

src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,94 @@
...
+            Valid::succeed(b_field)
+          })
+          .and(update_http().trace(config::Http::trace_name().as_str()))
+          .and(update_grpc(operation_type).trace(config::Grpc::trace_name().as_str()))
+          .and(update_graphql(operation_type).trace(config::GraphQL::trace_name().as_str()))
+          .try_fold(&(config, &_field, type_of, name), b_field)
+        })
+    },
+  )
+}
```
tusharmath, line 94: We should just see what's the resolver set in the blueprint's query type. And just clone that here.

tests/graphql/errors/test-call-operator.graphql
```diff
@@ -0,0 +1,32 @@
...
+type User {
+  id: Int!
+}
+
+type Post {
+  userId: Int!
+  withoutResolver: User @call(query: "userWithoutResolver", args: {id: "{{value.userId}}"})
+  multipleResolvers: User @http(path: "/users/{{value.userId}}") @call(query: "user", args: {id: "{{value.userId}}"})
+  withoutOperator: User @call(args: {id: "{{value.userId}}"})
+  # invalidResolver: User @call(query: "userWithGraphQLResolver", args: {id: "{{value.userId}}"})
```
tusharmath, line 21: Needs tests for argument mismatch.

ologbonowiwi: Can I rely on `Mustache::parse` to check it?

src/blueprint/from_config/operators/call.rs
ologbonowiwi: not sure if this will be used. I was using to test `Mustache` on this implementation, but I guess we won't need it. I'll delete before we merge it.

src/blueprint/from_config/operators/call.rs
ologbonowiwi: I'm not sure on how to access it. Where do I find it on this context (with the variables I can access here)?

My instant thought it was be to try on `config.schema.query` but this is only a string with the `Query` value.

docs/operators/call.md
ologbonowiwi: I'll do once we have everything wrapped up

src/blueprint/from_config/operators/call.rs
ologbonowiwi: so, on it, I was unable to access `b_field.resolver` because it was not defined yet, so I ended up using `compile_http` and `compile_graphql` methods instead.

examples/.tailcallrc.graphql
```diff
@@ -310,6 +310,35 @@ directive @cache(
...
+"""
+An union between JSON and ExprArgs
+"""
+union Args = JSON | ExprArgs
+
+input ExprArgs {
+  http: ExprHttp
+  grpc: ExprGrpc
+  graphQL: ExprGraphQL
+}
```
ologbonowiwi, line 340:
```suggestion
  args: JSON
) on FIELD_DEFINITION
```

src/config/config.rs
```diff
@@ -538,6 +544,13 @@ pub struct Http {
   pub encoding: Encoding,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, Default, PartialEq, Eq, schemars::JsonSchema)]
+pub struct Call {
+  #[serde(default, skip_serializing_if = "is_default")]
+  pub query: Option<String>,
+  pub args: HashMap<String, String>,
+}
```
tusharmath, line 552: Add rust docs. This will be used in generating `tailcallrc.schema.json`

src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,127 @@
...
+      let Some(call) = &field.call else {
+        return Valid::succeed(b_field);
+      };
+
+      if validate_field_has_resolver(name, field, &config.types).is_succeed() {
+        return Valid::fail(format!(
+          "@call directive is not allowed on field {} because it already has a resolver",
+          name
+        ));
+      }
```
tusharmath, line 32: Should implement `resolvable_directives` in `config.rs`

ologbonowiwi: done

src/config/config.rs
ologbonowiwi: done

src/blueprint/from_config/operators/call.rs
```diff
@@ -0,0 +1,133 @@
...
+pub fn compile_call(
+  field: &Field,
+  config: &Config,
+  call: &config::Call,
+  operation_type: &GraphQLOperationType,
+) -> Valid<Expression, String> {
+  if validate_field_has_resolver(field.name(), field, &config.types).is_succeed() {
+    return Valid::fail("@call directive is not allowed on field because it already has a resolver".to_string());
+  }
+
```
ologbonowiwi, line 42:
```suggestion
```

src/blueprint/operators/call.rs
```diff
@@ -0,0 +1,202 @@
...
+              .clone()
+              .query(req_template.clone().query.iter().map(replace_mustache(&args)).collect())
+          })
+          .map(|req_template| {
+            req_template
+              .clone()
+              .headers(req_template.headers.iter().map(replace_mustache(&args)).collect())
+          })
+          .map(|req_template| Expression::IO(IO::Http { req_template, group_by, dl_id })),
+          _ => Valid::succeed(expr),
```
ologbonowiwi, line 94: this code and the others `_ => Valid::succeed(expr)` on this file are unreachable. can I delete it @tusharmath @meskill?

mogery:
The serialization counterpart of this is missing from `into_document.rs`

ologbonowiwi:
fixed, thank you @mogery

docs/operators/call.md
ologbonowiwi: done: https://github.com/tailcallhq/tailcallhq.github.io/pull/102

src/blueprint/operators/call.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    fn try_from(expr: Expression) -> Result<Self, Self::Error> {
+        match expr {
+            Expression::IO(IO::Grpc { req_template, group_by, dl_id }) => {
+                Ok(Grpc { req_template, group_by, dl_id })
+            }
+            _ => Err("not a grpc expression".to_string()),
+        }
+    }
+}
```
ologbonowiwi, line 92: This was done to avoid unreachable code. I can delete it and use `match` letting an unreachable `_ => Valid::succeed` or `_ => panic!`, let me know which is better, guys.

ologbonowiwi:
I'll open another one. This had many outdated discussions, which makes harder to follow and review

Issue #859, opened 2023-12-31
Read environmental variables from `.env` files


rajdip-b:
**Is your feature request related to a problem? Please describe.**
Most of the time we have a `.env` file in our directory that contains all the secret. Right now, there is no way to directly use this file to resolve the env keys in the mustache statement.

**Describe the solution you'd like**
It would be nice to automatically detect and parse all environment variables in our mustache templates, rather than having to explicitely specifying them in our runtime environment.

Usage: 
If we have a `.env` file with the following content in it,
```
API_ENDPOINT=myapi.com
```

Then, this block should work by default.

```graphql showLineNumbers
type Query {
  users: [User]! @http(baseUrl: "{{env.API_ENDPOINT}}", path: "/users")
}
```

ologbonowiwi:
Agree. I suggested it when working on #790, but it was outside that issue's scope.

ksh0205:
Hi @tusharmath, Can I give it a shot? Can you assign the issue to me.

ologbonowiwi:
I'd like to work on this one as I implemented part of the env thing already

ologbonowiwi:
I started working on it, and I'll use `env-file-reader` to read the env.

one question: Do we want a new operator or a new field on an existing operator?

IMO, we could add an `env_file`, `env_path`, or something like that on the `@server` operator or create an `@env` operator.

e.g.:
```graphql
# new field on server operator
@server(env_path: ".env.something")

# OR

# new env operator
@env(path: ".env.something")
```

what do you think @tusharmath?

ologbonowiwi:
As discussed [on Discord](https://discord.com/channels/1044859667798568962/1156188728474214472/1191854194051776532), we'll use `@link` for that, so we're blocked until #886 be resolved.

Bhavyajain21:
is this closed? I still this issue as open in algora

ologbonowiwi:
It's really open @Bhavyajain21. I'm assigned to this one but feel free to give it a shot if you want to

tusharmath:
fixed in https://github.com/tailcallhq/tailcall/pull/1523

PR #886, opened 2024-01-03
feat: add @link operator

### LINKED ISSUE #760: Add a `@link` operator
### Overview
Introducing the `@link` operator for linking and importing of external files in configurations.

#### Proposed GraphQL Schema Definition
Create a link operator that has `src`, `type` and `id` fields as follows.
```graphql
schema 
  @link(type: "tailcall/graphql", src: "https://abc.com/graphql")
  @link(id: "abc" type: "tailcall/yml", src: "https://abc.com/application.yml")
  @link(type: "tailcall/graphql", src: "https://abc.com/application.graphql") {
  query: Query
}
```

**Technical Requirements**
- Each file is resolved and inside of config_reader.
- Resolved files are either merged with the original config or attached as an extension in the config_set.

melsonic:
Initial draft PR
Fixes #760


src/config/resolver.rs
```diff
@@ -0,0 +1,48 @@
...
+  src: String, // Source URL for linked files
+  #[serde(default, skip_serializing_if = "is_default")]
+  id: Option<String>, // Id is used to refer at different places in the config
+  #[serde(default, skip_serializing_if = "is_default")]
+  content: Option<String>, // Stores raw content
+}
+
+impl Link {
+
+  pub async fn read_link_type(&mut self) -> Result<(), anyhow::Error> {
```
tusharmath, line 31:
```suggestion
  pub async fn resolve_recurse(self) -> Result<Link, anyhow::Error> {
```

src/config/resolver.rs
```diff
@@ -0,0 +1,48 @@
...
+      self.content = Some(resp.text().await?);
+    } else {
+      let path = &self.src.trim_end_matches('/');
+      let mut f = File::open(path).await?;
+      let mut buffer = Vec::new();
+      f.read_to_end(&mut buffer).await?;
+      self.content = Some(String::from_utf8(buffer)?);
+    };
+    
+    Ok(())
```
tusharmath, line 46: Need to check for recursive links so that it doesn't get into a loop. 

- [ ] Add unit tests for this component
- [ ] Keep File and Http as a dependency so that it's easier to test.

src/config/resolver.rs
melsonic: Resolved :+1:

src/config/resolver.rs
melsonic: @tusharmath Regarding the recursive links, the `resolver` module only fetches the file content and places it into the `content` attribute of the `Link`, as it was mentioned that 
`Focused Module: The ConfigResolver is dedicated to fetching and aggregating content. Content parsing is handled in the config-to-blueprint step, ensuring modularity.`
Since content parsing is not done in this module It doesn't know if the linked file/URL contains any additional links. If we consider this case, we need to parse the content and add the links to the `Vec<Link>`.

tusharmath: `blueprint::from_config` is not async. So we can't perform any IOs in that step. Everything will need to be resolved before it reaches that step. We could parse the files and insert it in the config as data but not validate it. Parsing could be the responsibility of the Resolver module.

melsonic: adding the unit test in the next commit :+1:

src/config/resolver.rs
```diff
@@ -0,0 +1,96 @@
...
+        return Err(anyhow!("Read over URL failed with status code: {}", resp.status()));
+      }
+      res_link.content = Some(resp.text().await?);
+    } else {
+      let path = &link.src.trim_end_matches('/');
+      let mut f = File::open(path).await?;
+      let mut buffer: Vec<u8> = Vec::new();
+      f.read_to_end(&mut buffer).await?;
+      res_link.content = Some(String::from_utf8(buffer)?);
+    };
```
ologbonowiwi, line 93: we could make the `if`/`else` block return the `content` then we override.

```suggestion
    let content = if let Ok(url) = Url::parse(&link.src) {
      let resp = reqwest::get(url).await?;
      if !resp.status().is_success() {
        return Err(anyhow!("Read over URL failed with status code: {}", resp.status()));
      }
      Some(resp.text().await?);
    } else {
      let path = &link.src.trim_end_matches('/');
      let mut f = File::open(path).await?;
      let mut buffer: Vec<u8> = Vec::new();
      f.read_to_end(&mut buffer).await?;
      Some(String::from_utf8(buffer)?);
    };
    
    // override and return
```

this would also avoid needing to create `res_link` as mutable to return it afterward.

src/config/resolver.rs
```diff
@@ -0,0 +1,96 @@
...
+        }
+        let source = if let Some(v) = resp.headers().get("content-type") {
+          if let Ok(s) = Source::detect_content_type(v.to_str()?) {
+            s
+          } else {
+            Source::detect(path.trim_end_matches('/'))?
+          }
+        } else {
+          Source::detect(path.trim_end_matches('/'))?
+        };
```
ologbonowiwi, line 57: likely we can make the condition response an `Option` enum so we can make `let source = source.or(Source::detect(path.trim_end_matches('/'))?)` instead of returning `Source::detect(path.trim_end_matches('/'))?` on `else` blocks.

ologbonowiwi:
looks like there are some tests broken @melsonic. if you need a hand or want to collaborate on it, let me know, as #859 (which I'm assigned to) will rely on `@link` to load environment variables.

src/config/resolver.rs
ologbonowiwi: another option is to put this logic into a private function.

so you can make something like
```suggestion
        let source = get_source(resp).or(Source::detect(path.trim_end_matches('/'))?;
```

ologbonowiwi: and the private function would be closer to this:
```
  fn get_source(resp: /* the type */) -> anyhow::Result<Source> {
    let content_type = resp.headers().get("content-type")?;

    Source::detect_content_type(content_type)
  }
```

note this is not a ready-to-go code, and it'll likely require a few tweaks

src/config/resolver.rs
```diff
@@ -0,0 +1,96 @@
...
+        (resp.text().await?, source)
```
ologbonowiwi, line 58: this whole `if` condition block could be a private function. IMO, it'd make easier to read it.

src/config/resolver.rs
```diff
@@ -0,0 +1,96 @@
...
+      } else {
+        let path = &curr.src.trim_end_matches('/');
+        let mut f = File::open(path).await?;
+        let mut buffer = Vec::new();
+        f.read_to_end(&mut buffer).await?;
+        (String::from_utf8(buffer)?, Source::detect(path)?)
+      };
+
+      curr.content = Some(txt.clone());
+      let config = Config::from_source(source, &txt)?;
```
ologbonowiwi, line 68: if you change the lines' order, you don't need to clone `txt`

```suggestion
      let config = Config::from_source(source, &txt)?;

      curr.content = Some(txt);
```

src/config/resolver.rs
```diff
@@ -0,0 +1,96 @@
...
+        } else {
+          Source::detect(path.trim_end_matches('/'))?
+        };
+        (resp.text().await?, source)
+      } else {
+        let path = &curr.src.trim_end_matches('/');
+        let mut f = File::open(path).await?;
+        let mut buffer = Vec::new();
+        f.read_to_end(&mut buffer).await?;
+        (String::from_utf8(buffer)?, Source::detect(path)?)
```
ologbonowiwi, line 64: these lines are repeated on `get_raw_content` else block. we could put it into a function to be re-used.

src/config/resolver.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl Link {
+  pub async fn resolve_recurse(self) -> anyhow::Result<Vec<Link>> {
+    let mut result: Vec<Link> = Vec::new();
+    if self.type_of == LinkType::Protobuf || self.type_of == LinkType::Data {
+      let link: Link = Self::get_raw_content(&self).await?;
+      result.push(link);
+      return Ok(result);
+    }
```
ologbonowiwi, line 39: since all files can be both `http://` or `file://`, why we handle `Protobuf` and `Data` differently?

melsonic:
@ologbonowiwi  I've addressed the issues with the broken tests by making the `merge_right` function asynchronous, as the newly added `merge_links` function involves I/O operations. However, I am finding a problem with the unit test in the `resolver_module`. Unlike other attributes in the config, which are simple maps/structs, the links attribute stores a vector. It seems that we need to either modify the `from_directive` function or implement a workaround to handle this unique case.

melsonic:
> looks like there are some tests broken @melsonic. if you need a hand or want to collaborate on it, let me know, as #859 (which I'm assigned to) will rely on `@link` to load environment variables.

let's talk on discord :+1:

src/config/resolver.rs
melsonic: Since we don't need to detect the source type for these files. Because Config and GraphQL files could have recursive links, but these files don't. To, remove unnecessary processing I have created a different function for these two types

ologbonowiwi: from my more preferred to least preferred, I'd say that we have a few alternatives:

1. load source every time and don't check for it at all, as it'll just not find a link to recursive resolve;
2. move this condition to lines 48 and 63, early returning in case it matches;
3. move the loading (of both file/network protocol) to specific functions that can be re-used in both cases;

in case we go with `2.`, feel free to use `get_raw_content` as much as we don't duplicate the code for loading stuff (so this method will need to rely on the new methods).

IMO: if the processing is not high costly, we can certainly load everything and early return in case is `Protobuf` or `Data`.

src/config/config.rs
```diff
@@ -210,6 +214,18 @@ fn merge_unions(
   self_unions
 }
 
+pub async fn merge_links(mut self_links: Vec<Link>, other_links: Vec<Link>) -> anyhow::Result<Vec<Link>> {
+  // let runtime = tokio::runtime::Runtime::new()?;
+  for link in other_links {
+    // let recurse_links = runtime.block_on(async { Link::resolve_recurse(link).await })?;
+    let recurse_links = Link::resolve_recurse(link).await?;
```
ologbonowiwi, line 221: IMO, we should have one step to merge the configs and other step to actually load it.

here we should make only the sync operation of merging the data structure. the loading of it should be on other step.

tusharmath:
closing in favour of #911

Issue #750, opened 2023-12-14
🚀 Tailcall Mini Hackathon: Bringing WASM to the Party with $2000 Bounty! 🎉


tusharmath:
**Existing Functionality**
- [x] Entity Cache (#966)
- [x] HTTP Support (#892)
- [x] Implement `FileIO` via assets or buckets for cloudflare (#926)
- [x] Http Batching (#997)
- [x] HTTP Caching (Powered by cloudflare already)
- [x] GRPC support (#1041)

**Architecture**
- [x] Support for multi-tenant deployment (#991)
- [x] Add support for static io (#941)
- [x] Add support for loading config from url path (#980)
- [x] Create an IO Abstraction that supports File and Network (#914)
- [x] Add panic hook (#996)
- [x] Support Relative Paths in config reader (Out of scope)
- [x] Remove hard-coded tailcall config URL
- [ ] Remove feature flags for cargo.toml
- [ ] Update package structure — `cli`, `cloudflare`, `core`
- [ ] Drop usage of `default` feature flag (CLI work is pending)
- [x] Remove GRPC specific client APIs that take Protobuf as Input.
- [x] Merge `HttpClientOptions` with `Upstream`
- [x] Add `EnvIO` 
- [x] Review usage of `set_req_version`
- [x] Add unit tests/test for cloudflare response (#1016 )
- [x] Add HTML response for ValidationErrors
- [ ] Interface with async_graphql Req/Res instead of going via hyper
- [x] Move `Env` to test
- [ ] Drop `spawn_local` from `cloudflare::Http`
---

## Later
- [ ] Use `impl HttpIO` instead of `dyn HttpIO`
---

🌟 **Calling All Code Wizards: Tailcall Mini Hackathon Awaits! 🧙‍♂️💻**

We're on a quest to bring WASM to Tailcall, and we're rolling out the red carpet for developers far and wide! With a whopping $2000 bounty for the winner, this is your chance to dive into Rust, WebAssembly, and Cloudflare, and make a lasting impact. 🏆💼

**Your Challenge:**
- Bring WASM support to Tailcall for Cloudflare deployment.
- Ensure no performance dip and full feature functionality.
- Be the hero Tailcall needs!

**Why You Can't Miss This:**
- **$2000 Bounty:** A grand prize for the winner to sweeten the deal.
- **Community and Collaboration:** Join forces with fellow developers in our Discord community.
- **Gain Visibility:** By starring our repository, you help boost its prominence and your own profile.
[...]

PR #892, opened 2024-01-04
feat: add wasm compatibility to tailcall/core

ssddOnTop:
Orignal issue: #750


ssddOnTop:
List of things which are **not supported** in this PR:

## GRPC
**Cause**

Currently we only support compiling proto files with fs. [code referance](https://github.com/tailcallhq/tailcall/blob/2dcacb6e8fc48b840dbd4e2087e052f316df23ca/src/grpc/protobuf.rs#L80)

**Possible Solution**

Add support for loading proto files over http. 

## Caching

**Cause** 

reqwest::Response is not thread safe and it lacks support for `try_from` trait for standard http::Response. However this feature is expected to be released [soon](https://github.com/seanmonstar/reqwest/pull/1920#issuecomment-1868505500)

src/http/client/client_cli.rs
```diff
@@ -4,27 +4,18 @@ use http_cache_reqwest::{Cache, CacheMode, HttpCache, HttpCacheOptions, MokaMana
...
 use reqwest_middleware::{ClientBuilder, ClientWithMiddleware};
 
-use super::Response;
-use crate::config::{self, Upstream};
-
-#[async_trait::async_trait]
-pub trait HttpClient: Sync + Send {
-  async fn execute(&self, req: reqwest::Request) -> anyhow::Result<Response>;
-  async fn execute_raw(&self, req: reqwest::Request) -> anyhow::Result<reqwest::Response>;
-}
```
tusharmath, line 14: Make Response polymorphic on Body

```rust
pub trait HttpClient: Sync + Send {
  async fn execute(&self, req: reqwest::Request) -> anyhow::Result<Response<Value>>;
  async fn execute_raw(&self, req: reqwest::Request) -> anyhow::Result<reqwest::Response<Vec<u8>>>;
}
```

tusharmath:
**IMPORTANT**
Add a wasm build only for cloudflare before merging this PR.

Cargo.toml
```diff
@@ -8,63 +8,60 @@ name = "tailcall"
...
 num_cpus = "1.16.0"
-hyper-rustls = "0.24.2"
-rustls = "0.21.9"
-rustls-pemfile = "1.0.4"
 fnv = "1.0.7"
-futures-channel = { version = "0.3.30" }
+futures-channel = { version = "0.3.29" }
 futures-timer = { version = "3.0.2" }
-futures-util = { version = "0.3.30" }
+futures-util = { version = "0.3.29" }
```
tusharmath, line 60: don't downgrade dependencies.

PR #896, opened 2024-01-04
feat: add encoding option in @http

### LINKED ISSUE #887: Support for `application/x-www-form-urlencoded` in `@http`
Currently, the default action of the `@http` directive is to treat the request body as a JSON object. While this works with most APIs, some APIs do expect `application/x-www-form-urlencoded`. 

P.S. specifying `Content-Type: application/x-www-form-urlencoded` in the `@http` headers doesn't work

It would be really nice to have a feature where the request body gets treated based upon the content-type header. The default option is to go for `application/json`. But if specified otherwise, we should encode in `application/x-www-form-urlencoded`

rajdip-b:
**Summary:**  
Added a field called `encoding` in the `@http` directive with two current values:
- APPLICATION_JSON (default)
- APPLICATION_X_WWW_FORM_URLENCODED

If not used, or `APPLICATION_JSON` is used, then the `Content-Type` stays `application/json`.
Else, if `APPLICATION_X_WWW_FORM_URLENCODED` is used, `Content-Type` becomes `application/x-www-form-urlencoded`.

Along with the change in the `Content-Type`, the request body also gets changed accordingly.

**Issue Reference(s):**  
Fixes #887 
/claim #887

**Build & Testing:**

**Checklist:**

- [ ] I have updated the [documentation] accordingly. (Will be making a guide on this so just have updated the `.tailcallrc.graphql`) file


examples/.tailcallrc.graphql
```diff
@@ -203,6 +203,11 @@ directive @http(
   The `groupBy` parameter groups multiple data requests into a single call. For more details please refer out [n + 1 guide](https://tailcall.run/docs/guides/n+1#solving-using-batching).
   """
   groupBy: [String!]
+
+  """
+  The `encoding` parameter specifies the encoding of the request body. It can be `APPLICATION_JSON` or `APPLICATION_X_WWW_FORM_URLENCODED`. @default `APPLICATION_JSON`.
+  """
+  encoding: Encoding = APPLICATION_JSON
```
tusharmath, line 210:
```suggestion
  encoding: Encoding
```

src/http/request_template.rs
```diff
@@ -304,6 +323,20 @@ mod tests {
...
+    let req = tmpl.to_request(&ctx).unwrap();
+    assert_eq!(req.headers().get("Content-Type").unwrap(), "application/json");
+  }
+  #[test]
+  fn test_encoding_application_x_www_form_urlencoded() {
+    let tmpl = RequestTemplate::new("http://localhost:3000").unwrap().encoding(crate::config::Encoding::APPLICATION_X_WWW_FORM_URLENCODED);
+    let ctx = Context::default();
+    let req = tmpl.to_request(&ctx).unwrap();
+    assert_eq!(req.headers().get("Content-Type").unwrap(), "application/x-www-form-urlencoded");
+  }
```
tusharmath, line 338: Add tests to assert the body of the request for every encoding type.

src/http/request_template.rs
```diff
@@ -116,10 +133,16 @@ impl RequestTemplate {
     }
 
     let headers = req.headers_mut();
+    // We want to set the header value based on encoding
+    // We want to set the header value based on encoding
```
tusharmath, line 137:
```suggestion
```

src/http/request_template.rs
```diff
@@ -103,7 +105,22 @@ impl RequestTemplate {
...
+        Encoding::ApplicationJson => {
+          req.body_mut().replace(body.render(ctx).into());
+        }
+        Encoding::ApplicationXWwwFormUrlencoded => {
+          let raw_data: String = body.render(ctx);
+          // The raw_data can be both JSON or plain string. If it is
+          // JSON, we need to parse it, or, we can directly use it.
+          if let Ok(deserialized_data) = serde_json::from_str::<serde_json::Value>(&raw_data) {
+            let form_data = serde_urlencoded::to_string(deserialized_data).unwrap();
+            req.body_mut().replace(form_data.into());
```
tusharmath, line 118: `ctx` actually has the complete body already, and we are converting that value to string and then string to value and then value to URL encoded. This is going to be very slow and inefficient.

src/http/request_template.rs
```diff
@@ -103,7 +105,22 @@ impl RequestTemplate {
...
+      match &self.encoding {
+        Encoding::ApplicationJson => {
+          req.body_mut().replace(body.render(ctx).into());
+        }
+        Encoding::ApplicationXWwwFormUrlencoded => {
+          let raw_data: String = body.render(ctx);
+          // The raw_data can be both JSON or plain string. If it is 
+          // JSON, we need to parse it, or, we can directly use it.
+          if let Ok(deserialized_data) = serde_json::from_str::<serde_json::Value>(&raw_data) {
+            let form_data = serde_urlencoded::to_string(deserialized_data).unwrap();
```
tusharmath, line 117: Unsafe usage of `unwrap` Should propagate this error up.

src/http/request_template.rs
```diff
@@ -349,6 +394,69 @@ mod tests {
...
+    }));
+    let body = tmpl
+      .to_request(&ctx)
+      .unwrap()
+      .body()
+      .unwrap()
+      .as_bytes()
+      .unwrap()
+      .to_owned();
+    assert_eq!(body, "baz".as_bytes());
```
tusharmath, line 415: Too much boilerplate involving unwrapping and chaining of methods. Would be better if we create helper methods to make this more maintainable.

rajdip-b: that might involve changing code outside the scope of this issue. can we park this for a dependent issue? so that when this gets merged, i can start working on the refactor?

src/http/request_template.rs
rajdip-b: image
would this work? seems fair

tusharmath (approved):

src/http/request_template.rs
tusharmath: I have made this and a few other changes in this PR. Have a look.

PR #915, opened 2024-01-09
fix: try_fold multiple calls for the same handler

meskill:
**Summary:**  
Fixes the problem with TryFold type that calls some of the handlers multiple times. It mostly affects conversion from config to blueprint and leads to some of handlers to set resolvers by operators to get called up to 4 times due to nested try_folds.

Right now there is no big problem in the code because most of the updates returns new instance to use instead of mutating it. But in case of mutating in such handlers it could lead to problems and unnecessary wrappers.

The problem revealed itself in the https://github.com/tailcallhq/tailcall/pull/833 and `cache` directive. So if I debug this I see Cache expression inside Cache expression because of the bug.

image

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/try_fold.rs
+++ b/src/try_fold.rs
@@ -4,7 +4,7 @@ use crate::valid::Valid;
 ///
 /// `TryFolding` describes a composable folding operation that can potentially fail.
 /// It can optionally consume an input to transform the provided value.
-type TryFoldFn<'a, I, O, E> = Box<dyn Fn(&I, O) -> Valid<O, E> + 'a>;
+type TryFoldFn<'a, I, O, E> = Box<dyn FnOnce(&I, O) -> Valid<O, E> + 'a>;
 
 pub struct TryFold<'a, I: 'a, O: 'a, E: 'a>(TryFoldFn<'a, I, O, E>);
 
@@ -18,7 +18,7 @@ impl<'a, I, O: Clone + 'a, E> TryFold<'a, I, O, E> {
   /// # Returns
   /// Returns a `Valid` value, which can be either a success with the folded value
   /// or an error.
-  pub fn try_fold(&self, input: &I, state: O) -> Valid<O, E> {
+  pub fn try_fold(self, input: &I, state: O) -> Valid<O, E> {
     (self.0)(input, state)
   }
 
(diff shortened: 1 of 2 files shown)


src/try_fold.rs
```diff
@@ -160,30 +170,25 @@ impl<'a, I, O: Clone, E> FromIterator<TryFold<'a, I, O, E>> for TryFold<'a, I, O
...
+    let t2 = TryFold::<i32, i32, ()>::new(|a: &i32, b: i32| {
+      *call_number.borrow_mut() += 1;
+      Valid::succeed(a * b)
+    });
     let t = t1.and(t2);
 
     let actual = t.try_fold(&2, 3).to_result().unwrap();
     let expected = 10;
 
+    assert_eq!(*call_number.borrow(), 1);
```
meskill, line 191: t2 is called 2 times rn

tusharmath: Can we add a separate test for asserting FnOnce behaviour? The existing tests should pass with the new implementation.

src/try_fold.rs
```diff
@@ -34,9 +34,12 @@ impl<'a, I, O: Clone + 'a, E> TryFold<'a, I, O, E> {
...
     TryFold(Box::new(move |input, state| {
-      self
-        .try_fold(input, state.clone())
-        .fold(|state| other.try_fold(input, state), other.try_fold(input, state))
+      let r = self.try_fold(input, state.clone()).to_result();
+
+      match r {
+        Ok(state) => other.try_fold(input, state),
+        Err(error) => Valid::<O, E>::from_validation_err(error).and(other.try_fold(input, state)),
+      }
```
tusharmath, line 42: As a general rule of thumb, we should always use the most fundamental operator that's available on the domain. In this case, Valid has the `fold` operator and we should leverage it, instead of using the result. This has two advantages - 
1. Your implementation remains unaffected when the `Result` type is dropped from Valid.
2. The code is more concise, and maintains correctness without duplicating lot of logic.
3. Optimizing `fold` will optimize every operator that's built on it.

src/try_fold.rs
meskill: actually, removed that change, since `test_order` should already verify that the handlers are called once

tusharmath (approved):

PR #927, opened 2024-01-12
feat: apollo federation compatible subgraph

amitksingh1490:


tusharmath:
Will come back to this later.

PR #938, opened 2024-01-14
perf: optimize `set_body` function

### LINKED ISSUE #934: performance: improve performance of `RequestTemplate`
The `set_body` method can be optimized for the case of `ApplicationXWwwFormUrlencoded`.
Currently, it renders the mustache template as a string, then converts it to serde_json and then again converts it to a form_url_encoded value.

Ideally, there it shouldn't take such redundant steps to make this transformation.

**Technical Requirements**
- Write idiomatic rust code while making an optimization
- Write criterion benchmarks to compare the before and after performance.

SOUMOJIT-CHAKRABORTY:
**Summary:**  
This pull request optimizes the set_body function in the RequestTemplate module, eliminating redundancy and improving code readability.

**Issue Reference(s):**  
Fixes #934

**Build & Testing:**

- [x ] I ran `cargo test` successfully.

**Checklist:**

- [ x] I have added relevant unit & integration tests.
- [x ] I have performed a self-review of my code.
- [x ] PR follows the naming convention of `<type>(<optional scope>): <title>`

### DIFF
--- a/src/http/request_template.rs
+++ b/src/http/request_template.rs
@@ -2,9 +2,8 @@ use std::borrow::Cow;
 
 use derive_setters::Setters;
 use hyper::HeaderMap;
-use reqwest::header::HeaderValue;
+use reqwest::header::{HeaderValue, CONTENT_TYPE};
 use url::Url;
-
 use crate::endpoint::Endpoint;
 use crate::has_headers::HasHeaders;
 use crate::helpers::headers::MustacheHeaders;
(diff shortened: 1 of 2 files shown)


tusharmath:
@SOUMOJIT-CHAKRABORTY Take a pull from the main and then run the benchmarks comparing results.

SOUMOJIT-CHAKRABORTY:
> @SOUMOJIT-CHAKRABORTY Take a pull from the main and then run the benchmarks comparing results.

src/http/request_template.rs
```diff
@@ -100,12 +99,32 @@ impl RequestTemplate {
...
+}
+
+  /// checks for the CONTENT_TYPE header
+  pub fn is_application_x_www_form_urlencoded(headers: &HeaderMap) -> bool {
+    if let Some(content_type) = headers.get(CONTENT_TYPE) {
+        if let Ok(content_type_str) = content_type.to_str() {
+            return content_type_str.to_lowercase().starts_with("application/x-www-form-urlencoded");
+        }
+    }
+    false
```
ologbonowiwi, line 127: We could have the headers parsed into a Rust enum so we don't have something specific to `form-urlencoded`. Instead, we receive an enum and handle explicitly each case.

Two approaches can be taken:
- Have one variation to each implemented enum, plus a non-implemented variation
  - e.g., if matches `application/x-www-form-urlencoded` is urlencoded, else if matches `application/json`, otherwise it'll be the non-implemented variation.
- Consider `application/json` the default and add more variations to each new implemented
  - e.g., if matches `application/x-www-form-urlencoded` is urlencoded, otherwise it'll always be json

SOUMOJIT-CHAKRABORTY: @ologbonowiwi Thanks for the help.🙂

PR #948, opened 2024-01-15
feat: add initial set of operators in expr

shashitnak:
**Summary:**  
Adding `concatAll` and `intersectionAll` operations to `expr`

**Build & Testing:**

**Checklist:**


src/lambda/expression.rs
```diff
@@ -78,6 +85,15 @@ pub enum EvaluationError {
...
   APIValidationError(Vec<String>),
+
+  #[error("ConcatException: {0:?}")]
+  ConcatException(String),
+
+  #[error("IntersectionException: {0:?}")]
+  IntersectionException(String),
+
+  #[error("OperationFailed: {0:?}")]
+  OperationFailed(String),
```
tusharmath, line 96: Generalize it to `ExprEvaluationError` don't need specialized errors for each type of expr operator.

src/lambda/logic.rs
```diff
@@ -0,0 +1,141 @@
...
+            while let Some(result) = futures.next().await {
+              let result: Result<ConstValue> = result;
+              if is_truthy(result?) {
+                output = true;
+                break;
+              }
+            }
+            output
+          }
+          Concurrency::Sequential => {
```
tusharmath, line 49: implement a method in `Concurrency` 

```rust
impl<A> Concurrency {
  fn foreach(&self, iter: Iterator, fn: () -> impl Future<A>) -> Vec<A>;
  fn foreach_unordered(&self, iter: Iterator, fn: () -> impl Future<A>) -> Vec<A>;
}
```

src/lambda/expression.rs
```diff
@@ -31,11 +32,17 @@ pub enum Expression {
   EqualTo(Box<Expression>, Box<Expression>),
   Unsafe(Unsafe),
```
tusharmath, line 33:
```suggestion
  IO(IO),
```

Rename `Unsafe` to `IO`

src/lambda/eval.rs
```diff
@@ -0,0 +1,28 @@
...
+    Output: 'a,
+  {
+    Box::pin(self.async_eval(ctx, conc))
+  }
+
+  fn async_eval<'a, Ctx: ResolverContextLike<'a> + Sync + Send>(
+    &'a self,
+    ctx: &'a EvaluationContext<'a, Ctx>,
+    conc: &'a Concurrency,
+  ) -> impl Future<Output = Result<Output>> + Send;
```
tusharmath, line 27: We should keep only one implementation or make the return type an associated type for Eval.

src/lambda/expression.rs
tusharmath: Move Unsafe to `io.rs`

tusharmath (approved):

PR #993, opened 2024-01-19
fix: fixing entity caching bug

### LINKED ISSUE #835: Bug in entity cache implementation
**Problem description**
In the current cache implementation, if `@cache` is added to fields of a type and not to the query where it is being resolved then the values doesn't get cached.
```graphql
type Query {
}

type Post {
  id: Int! @cache(maxAge: 3000)
  userId: Int! @cache(maxAge: 4000)
  title: String!
  body: String!
}
```
The above config will not result in any form of caching.
```graphql
type Query {
  posts: [Post] @http(path: "/posts") @cache(maxAge: 3000)
}

type Post {
  id: Int!
  userId: Int!
  title: String!
  body: String!
}
```
This config, however, will cache the result of `posts` query.

**Solution**
One of the ways of fixing this issue is to do the following:
- Add a field in `ServerContext` to hold the values provided by `@cache` directive for whichever field it is added to.
- While creating the resolver function in `src/blueprint/into_schema.rs` use the `field.of_type.name` to get the type of output and use it to determine what caching rules are applied to the type and the fields of that type and use them to decide what values to fetch and what already exist in the cache.

shashitnak:
**Summary:**  
Fixing entity caching bug

**Issue Reference(s):**  
Fixes #835 

**Build & Testing:**

**Checklist:**


tests/http/entity-caching.yml
```diff
@@ -0,0 +1,41 @@
...
+        - id: 3
+          foo:
+            id: 4
+        - id: 5
+          foo:
+            id: 6
+        - id: 7
+          foo:
+            id: 8
+assert:
```
meskill, line 20: is this test somehow tests the bug?

src/blueprint/into_schema.rs
```diff
@@ -63,6 +63,58 @@ fn get_cache_key<'a, H: Hasher + Clone>(
...
+) -> anyhow::Result<ConstValue> {
+  if let Some(Cache { hasher, .. }) = ctx.req_ctx.type_cache_config.get(type_name) {
+    let hasher = hasher.clone();
+
+    if let Some(key) = get_cache_key(ctx, hasher) {
+      return ctx
+        .req_ctx
+        .cache
+        .get(&key)
+        .ok_or(anyhow::anyhow!("NOT_PRESENT_IN_CACHE"));
```
meskill, line 99: looks like unnecessary conversion to Result type instead of using Option

src/blueprint/into_schema.rs
```diff
@@ -63,6 +63,58 @@ fn get_cache_key<'a, H: Hasher + Clone>(
...
+  //   let fields_present_in_response = fld_caches
+  //     .iter()
+  //     .filter_map(|(k, c)| {
+  //       output.get_key(k).map(|r| (r, c))
+  //     });
+  //   for (field, Cache{max_age: ttl, hasher}) in fields_present_in_response {
+  //     let key = hasher.finish();
+  //     ctx.req_ctx.cache.insert(key, field.clone(), *ttl);
+  //   }
+  // }
```
meskill, line 115: do you need those commented code for anything?

src/blueprint/into_schema.rs
```diff
@@ -102,11 +160,23 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
+                      const_value
+                    } else {
+                      expr.eval(&ctx, &Concurrent::Sequential).await?
+                    }
+                  }
                 };
 
+                if !read_from_entity_cache {
+                  write_entity_cache(&ctx, &of_type, &const_value);
+                }
```
meskill, line 176: sounds like we gonna to write to cache without checking that cache is enabled. Is it ok?

src/blueprint/into_schema.rs
```diff
@@ -102,11 +160,23 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
-                  _ => expr.eval(&ctx, &Concurrent::Sequential).await?,
+                  _ => {
+                    if let Ok(const_value) = read_entity_cache(&ctx, &of_type).await {
+                      read_from_entity_cache = true;
+                      log::info!("Reading to cache.");
+                      const_value
+                    } else {
+                      expr.eval(&ctx, &Concurrent::Sequential).await?
+                    }
+                  }
```
meskill, line 171: I thought in `src/config/from_document.rs` we copy cache settings from object type down to fields, so settings are spread to every field. It doesn't work that way how I think? If I'm right then there couldn't be a situation when we have an object level cache and not field level

tests/http/entity-caching.yml
shashitnak: not really, this is to test that the new changes don't mess up anything and to increase code coverage. I have manually tested it and I am not sure if it's possible to write an integration test like this to check that

src/blueprint/into_schema.rs
shashitnak: the function `write_entity_cache` is doing that checking

src/blueprint/into_schema.rs
shashitnak: The way the implementation works currently is that the cache directives at field will work only for the fields which have resolvers. If we want a type to be cached irrespective of the query fetching it, then we will need to keep it on the object level and it is used globally by any query or field which will output that type

meskill: now I get it.

but anyway if we still copy cache settings to field level is `_ =>` condition will apply? It will probably be still something there

src/blueprint/into_schema.rs
shashitnak: removed them. it wasn't required

src/blueprint/into_schema.rs
shashitnak: Good point. I think now since we have an additional way of caching at entity level, it is better to not make the fields inherit the cache setting provided at the type. Directives provided at field will use the old way of caching and the ones at the type will use the new implementation. Does this sound like the right way of doing it? @tusharmath @amitksingh1490

meskill: Let's think again about the bug and the fix because current code looks strange for me.

For single field you just now have 2 caches with unclear relation. Consider the case when caches are empty:
1. ctx.req_ctx.cache_get for field level cache returns None
2. read_entity_cache for type level cache returns None
3. then we execute else branch that writes field level cache
4. since read_from_entity_cache is false we write also to type level cache

At the end we have same value in two caches and we'll use only first one anyway.

Consider now when you want to remove copying `@cache` directive from type to field. In that case for such fields we won't create field level cache and will use type level cache. Seems ok, but it's actually the same as before just with different cache entity.

I see the example in the parent issue like this
```
type Query {
}

type Post {
  id: Int! @cache(maxAge: 3000)
  userId: Int! @cache(maxAge: 4000)
  title: String!
  body: String!
}
```

but it's not clear for me what are we going to cache here? constant values that anyway will be resolved by actual http parent request? what are the benefits of caching values that will be resolved by something else?

shashitnak: The limitation of the cache directive at field level, right now, is that we cannot cache the values which do not have some resolver, because the code responsible for caching the output of the field doesn't get executed for them. In the above example, only if cache is applied for the `posts` query or the `user` field inside `Post` type will it be able to cache the output. This can be fixed if while storing the cache we split the values into separate fields and store them separately. This would require us to store the cache settings globally just like I'm doing for this global cache.
Also, the reason we are having to store same value at two separate caches is that, the first cache will only work for that particular query or field but the second one will work for any query or field which output's that type. And also the second cache will not be populated if cache isn't applied to the output type of that query. The way I am thinking about this is that, if the user wants to cache a type, irrespective of the query which outputs it, then they will only provide cache setting for that type, and if they only want to cache output of a query, irrespective of the type which it outputs then they will provide it at that query

meskill: I've actually misread the code in terms what is type_name is used to check the cache. It quite confusing that while parsing config we copy cache settings from type to all the fields of that type. And here we check the type of the field itself, not the wrapping type. Am I right?

Also, I think example from the issue is misleading since the fields without resolver are actually resolved with another branch above i.e.

```
match &field.resolver {
  None => {
    FieldFuture::from_value(ctx.path_value(&[field_name]).map(|a| a.to_owned()))
  }
```

and that related to values that is already resolved by parent value with its resolver and there is no benefits of caching those.

tusharmath:
fixing in favour of #1078

PR #998, opened 2024-01-20
chore(cloudflare): package cleanup

meskill:
**Summary:**  
Activate optimized profile for wasm build

Size dropped from `9,78 MB` to `6,78 MB` for `cloudflare.wasm`

**Build & Testing:**

**Checklist:**


ssddOnTop (approved):

ssddOnTop (commented):
the total upload size still remains same after the changes

meskill:
> the total upload size still remains same after the changes

is there another tool that makes additional compilation for wasm?

if so then we probably do not need special profiles for cloudflare at all in our config. For now cargo just spams warnings about wrong usage having it inside `cloudflare/Cargo.toml`

ssddOnTop:
We can not do anything externally, wrangler is designed to compile and publish every time, even if we use some external tools, it will just recompile before publishing.

Anyways its nice to save 3 mb of local storage with this pr

meskill:
> We can not do anything externally, wrangler is designed to compile and publish every time, even if we use some external tools, it will just recompile before publishing.
> 
> Anyways its nice to save 3 mb of local storage with this pr

my question was about why the upload size stayed the same after change.

But I think I already got it. It's because wrangler is have step like `[INFO]: Optimizing wasm binaries with wasm-opt` when deploying and therefore those settings are not so important and affects build output only when building purely with cargo

cloudflare/package.json
```diff
@@ -4,8 +4,8 @@
   "private": true,
   "scripts": {
     "deploy": "npx wrangler deploy --minify",
-    "publish": "npx wrangler publish --minify",
```
meskill, line 7: publish is deprecated https://developers.cloudflare.com/workers/wrangler/commands/#publish

ssddOnTop (approved):

tusharmath (approved):

cloudflare/package.json
tusharmath: We should add `npm run deploy` so that there is only one command we work with.

cloudflare/package.json
```diff
@@ -4,8 +4,8 @@
...
-    "dev": "npx wrangler dev --port 19194 --remote"
+    "test-debug": "npx wrangler dev --port 19194 --remote --env debug",
+    "test-release": "npx wrangler dev --port 19194 --remote"
```
tusharmath, line 8:
```suggestion
    "dev-debug": "npx wrangler dev --port 19194 --env debug",
    "dev": "npx wrangler dev --port 19194"
```

This is not a test command so let's not use `test`. The additional params such as `--remote` can be passed anyways with `npm run` so can drop those too.

cloudflare/package.json
meskill: it's right above it

tusharmath (approved):

tusharmath (approved):

PR #1015, opened 2024-01-23
feat: add `/showcase/graphql` endpoint

### LINKED ISSUE #1012: feature: add a `/showcase` url
The `/showcase` url on tailcall takes a `config` query parameter.

```
curl https://tailcall.run/showcase/graphql?config=http://abc.com/tc.graphql
```
The `config` parameter takes a link to the GraphQL file and runs a GraphQL server in that context. 

**Technical Requirements**
- Similar code is written `cloudflare/handle` remove it from there.
- Ensure unit tests are added.
- All error messages while validating should be returned as a JSON response.
- Add a feature flag in `@server(showcase: Boolean)` to configure the feature enablement.

mogery:
**Summary:**  
This PR adds a `/showcase/graphql` endpoint that behaves like how it's described in #1012. Some caveats:
 - For security reasons, the parsed config will not have access to environment variables.
 - The `Upstream` config will be ignored.
 - The cache will be shared with the main instance.

This PR also rewrites `cloudflare/src/handler.rs` to use the new showcase logic. That implementation has none of the above caveats, and should continue to work like it has before, except that it may have even better error reporting.

**Issue Reference(s):**  
Fixes #1012 
/claim #1012 

**Build & Testing:**

**Checklist:**

  - tailcallhq/tailcallhq.github.io#101


src/http/request_handler.rs
```diff
@@ -113,6 +115,23 @@ pub async fn handle_request<T: DeserializeOwned + GraphQLRequestLike, Http: Http
   state: Arc<AppContext<Http, Env>>,
 ) -> Result<Response<Body>> {
   match *req.method() {
+    hyper::Method::POST
+      if req.uri().path().ends_with("/showcase/graphql") && state.blueprint.server.enable_showcase =>
```
tusharmath, line 119: check `enable_showcase` first. That's going to be faster.

src/http/showcase.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::async_graphql_hyper::{GraphQLRequestLike, GraphQLResponse};
+use crate::blueprint::Blueprint;
+use crate::cli::{init_chrono_cache, init_http};
+use crate::config::reader::ConfigReader;
+use crate::config::Upstream;
+use crate::{EnvIO, FileIO, HttpIO};
+
+struct DummyFileIO;
+
+impl FileIO for DummyFileIO {
```
tusharmath, line 19: for a custom file IO. We can load configuration over HTTP anyways via `config::reader`, we should leverage that.

mogery: `ConfigReader` requires a `File` handle and will attempt to read from the file system if it cannot parse the URL. This is needed in `cloudflare/src/handle.rs`, but it could be dangerous in a showcase environment. I implemented this to allow disabling the `File` operations without uprooting `ConfigReader`.

I can make the change in `ConfigReader` to take an `Option<File>` instead of this if you'd like, but I'd need to change a lot of unrelated codepaths.

mogery:
WASM build fails should be fixed. Unfortunately I'm not sure why the linter failing. The action doesn't log any issues and rustfmt is broken for me locally so I can't check for it easily. Might spin up a codespace and try running it there.

mogery:
Ugh, nevermind, I'm running into compiler problems...

src/http/showcase.rs
mogery: > I can make the change in `ConfigReader` to take an `Option<File>` instead of this if you'd like, but I'd need to change a lot of unrelated codepaths.

Actually, this is not even possible. We need to provide `showcase_get_app_ctx` with a type implementing `FileIO` even when we're not using it, because using `Option<dyn FileIO>` causes a type-checking cycle and a build fail. We could give it `NativeFileIO` as a type, but that breaks WASM build, because it will depend on `cli`. Thus `DummyFileIO` is needed.

mogery:
@tusharmath Everything should be resolved now wrt CI checks. Also opened up a codespace to fix lint successfully, my local rustfmt seems to be messed up.

tests/http_spec.rs
```diff
@@ -311,7 +311,11 @@ impl HttpIO for MockHttpClient {
       response.body = Bytes::from_iter(body);
       Ok(response)
     } else {
-      let body = serde_json::to_vec(&mock_response.0.body)?;
+      let body = if let Value::String(x) = mock_response.0.body {
+        string_to_bytes(&x)
+      } else {
+        serde_json::to_vec(&mock_response.0.body)?
+      };
```
tusharmath, line 318: Can you explain why we need this change?

cloudflare/src/handle.rs
```diff
@@ -24,70 +23,64 @@ lazy_static! {
 ///
 pub async fn fetch(req: worker::Request, env: worker::Env, _: worker::Context) -> anyhow::Result<worker::Response> {
   log::info!("{} {:?}", req.method().to_string(), req.url().map(|u| u.to_string()));
-  let env = Rc::new(env);
-  let hyper_req = to_request(req).await?;
-  if hyper_req.method() == hyper::Method::GET {
-    let response = graphiql(&hyper_req)?;
+  let req = to_request(req).await?;
+  if req.method() == hyper::Method::GET {
```
tusharmath, line 27: I don't think we need to handle this here. The behavior should be controlled by the core module.

src/http/request_handler.rs
```diff
@@ -113,6 +114,26 @@ pub async fn handle_request<T: DeserializeOwned + GraphQLRequestLike, Http: Http
...
+      )
+      .await?
+      {
+        Ok(server_ctx) => server_ctx,
+        Err(res) => return Ok(res),
+      };
+
+      graphql_request::<T, _, _>(req, &server_ctx).await
+    }
     hyper::Method::POST if req.uri().path().ends_with("/graphql") => {
```
tusharmath, line 137: This `POST /graphql` is in our hot path and should be the first route we should check.

src/http/request_handler.rs
```diff
@@ -113,6 +114,26 @@ pub async fn handle_request<T: DeserializeOwned + GraphQLRequestLike, Http: Http
   state: Arc<AppContext<Http, Env>>,
```
tusharmath, line 114: we should pass the app context from here instead.

tests/http_spec.rs
mogery: it lets us mock an endpoint that returns a GraphQL schema in `tests/http/showcase.yml`

src/http/request_handler.rs
mogery: The endpoint is not `POST /graphql`, it's `POST */graphql`, since we're using `ends_with`. Checking it first would require us to check that the endpoint ends with `/graphql`, **and that the endpoint doesn't end with `/showcase/graphql`**, otherwise we can't check for the showcase. This comparison has to be run anyways, so changing the order of the match wouldn't make any difference.

src/http/request_handler.rs
mogery: wdym?

src/http/request_handler.rs
mogery: We could change the showcase endpoint to just be `/showcase` so that it doesn't collide with the `/graphql` check

mogery:
Rebased on main and changed the internal API a bit. Not having AppContext generics made it quite a bit prettier!

mogery:
Running into a JS-related `http_spec` fail on this one. Let's put it on hold 'til I can debug tomorrow.

mogery:
Alright, this is should be passing everywhere now :)

cloudflare/src/handle.rs
mogery: Had to re-add this. When GraphiQL is requested, a config parameter is not specified, and `get_app_ctx` fails without it, so we don't actually get to the point where `handle_request` could route to `graphiql`.

mogery:
This is ready now @tusharmath

cloudflare/src/lib.rs
```diff
@@ -17,12 +17,14 @@ pub fn init_env(env: Rc<worker::Env>) -> Arc<dyn EnvIO> {
     Arc::new(env::CloudflareEnv::init(env))
 }
 
-pub fn init_file(env: Rc<worker::Env>, bucket_id: String) -> anyhow::Result<Arc<dyn FileIO>> {
-    #[allow(clippy::arc_with_non_send_sync)]
+pub fn init_file(
+    env: Rc<worker::Env>,
+    bucket_id: String,
+) -> anyhow::Result<Arc<dyn FileIO + Send + Sync>> {
```
tusharmath, line 23: FileIO and HttpIO is already send sync

src/cli/mod.rs
```diff
@@ -30,7 +30,7 @@ pub fn init_env() -> Arc<dyn EnvIO> {
 }
 
 // Provides access to file system in native rust environment
-pub fn init_file() -> Arc<dyn FileIO> {
+pub fn init_file() -> Arc<dyn FileIO + Send + Sync> {
```
tusharmath, line 33: FilIO is already Send + Sync

src/http/showcase.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub async fn create_app_ctx<T: DeserializeOwned + GraphQLRequestLike>(
+    req: &Request<Body>,
+    runtime: TargetRuntime,
+    enable_fs: bool,
+) -> Result<Result<AppContext, Response<Body>>> {
+    let url = Url::parse(&req.uri().to_string())?;
+    let mut query = url.query_pairs();
+
+    let config_url = if let Some(pair) = query.find(|x| x.0 == "config") {
```
tusharmath, line 21: Can use `serde_qs` here.

src/cli/mod.rs
mogery: Fixed

tusharmath (approved):

PR #1026, opened 2024-01-24
fix(cache): cache set failure

### LINKED ISSUE #1021: Bug in NativeChronoCache set method
### Prerequisites

### Describe the bug

The `NativeChronoCache::set` method throws an error when `None` is returned from `TtlCache::insert`. `TtlCache::insert` returns `None` when the inserted key isn't already present in the cache and so it doesn't indicate any failure.

### Steps to reproduce

1. Create an instance of `NativeChronoCache`.
2. try inserting a key which isn't already present in the cache.

### Expected behavior

The output of `NativeChronoCache::set` should be an `Option` type which should return the output of `TtlCache::insert` and the decision should be left on the caller on how they want to handle the output.

### Actual behavior

An error is thrown saying `unable to insert value`.

shashitnak:
**Summary:**  
Fixing the bug in set method of NativeChronoCache

**Issue Reference(s):**  
Fixes #1021 

**Build & Testing:**

**Checklist:**


src/cli/cache.rs
```diff
@@ -32,12 +32,14 @@ impl<K: Hash + Eq + Send + Sync, V: Clone + Send + Sync> Cache for NativeChronoC
...
-      .write()
-      .unwrap()
-      .insert(key, value, ttl)
-      .ok_or(anyhow!("unable to insert value"))
+    Ok(
+      self
+        .data
+        .write()
+        .unwrap()
+        .insert(key, value.clone(), ttl)
```
meskill, line 40: well, it's an additional cloning for value (that is already happens inside `into_schema.rs`) mostly just to satisfy type?

I wonder why we even return `V` in this method to drop it immediately inside `request_context`? Maybe it'll be better to change interface and drop `V`?

shashitnak: maybe I can convert it from Result to Option type. since `TtlCache` returns an Option type so returning it wouldn't have any overhead. I just think it should atleast return the optional type because there might be some use case in the future where that value is required

src/lib.rs
```diff
@@ -51,8 +51,8 @@ pub trait FileIO {
 pub trait Cache: Send + Sync {
   type Key: Hash + Eq;
   type Value;
-  async fn set<'a>(&'a self, key: Self::Key, value: Self::Value, ttl: NonZeroU64) -> anyhow::Result<Self::Value>;
-  async fn get<'a>(&'a self, key: &'a Self::Key) -> anyhow::Result<Self::Value>;
+  async fn set<'a>(&'a self, key: Self::Key, value: Self::Value, ttl: NonZeroU64) -> Option<Self::Value>;
+  async fn get<'a>(&'a self, key: &'a Self::Key) -> Option<Self::Value>;
```
tusharmath, line 55: I think it's better if we specify the error types as an associated type. Error can happen during insertion and retrieval. For ex: the remote service, in case of cloudflare's KV store, might be inaccessible or offline.

src/cli/cache.rs
```diff
@@ -25,28 +24,18 @@ impl<K: Hash + Eq, V: Clone> NativeChronoCache<K, V> {
     NativeChronoCache { data: Arc::new(RwLock::new(TtlCache::new(CACHE_CAPACITY))) }
   }
 }
+
```
tusharmath, line 27: `src/cli/cache.rs` can be moved into `src/cache.rs` because it's not a CLI specific implementation.

src/cli/cache.rs
```diff
@@ -25,28 +24,18 @@ impl<K: Hash + Eq, V: Clone> NativeChronoCache<K, V> {
...
   }
 
-  async fn get<'a>(&'a self, key: &'a K) -> Result<V> {
-    self
-      .data
-      .read()
-      .unwrap()
-      .get(key)
-      .cloned()
-      .ok_or(anyhow!("key not found"))
```
tusharmath, line 50: There are many changes being done here, can we add a test case.

src/lib.rs
shashitnak: making Error an associated type will work but I think creating an Error type which will behave like an actual error for the get method and not an error in the set method doesn't seem right. Instead I am thinking I can just make the `V` type an `Option` and just simply return the `insert` and `get` results from the `TtlCache` with `Ok` everytime

shashitnak: > making Error an associated type will work but I think creating an Error type which will behave like an actual error for the get method and not an error in the set method doesn't seem right. Instead I am thinking I can just make the `V` type an `Option` and just simply return the `insert` and `get` results from the `TtlCache` with `Ok` everytime

Actually this wouldn't work. I'll make an error type instead

shashitnak: I changed the set method to output a `Result<()>`. Other ways just seemed to make things unncessarily complicated so I thought since we aren't using the old value from the cache when we do an insert, we can return unit type for now

src/cli/cache.rs
shashitnak: moved it to `src/native_chrono_cache.rs` because there's already a file named `src/cache.rs` which contains a `Cache` type

src/cli/cache.rs
shashitnak: added unit test for `NativeChronoCache`

src/lib.rs
tusharmath: A `get` operation can fail or succeed with no value. I think a better design here would be `Result<Option<Self::Value>>`. Not having the value vs not being able to retrieve it are two separate things and that it should be propagated to the caller.

src/cli/cache.rs
tusharmath: this is a in-memory implementation. We can use it as a default implementation for the trait. Better we move it to `cache.rs`. There is nothing "native" about it, we can use this implementation anywhere.

src/lib.rs
shashitnak: So I will have to make this change in the `Cache` trait because the `Self::Value` type has to be the type that actually goes inside the `TtlCache`. If we're doing this then I think we can change the `set` to `Result<Option<Self::Value>>` as well

src/lib.rs
```diff
@@ -58,8 +58,8 @@ pub trait Cache: Send + Sync {
         key: Self::Key,
         value: Self::Value,
         ttl: NonZeroU64,
-    ) -> anyhow::Result<Self::Value>;
-    async fn get<'a>(&'a self, key: &'a Self::Key) -> anyhow::Result<Self::Value>;
+    ) -> anyhow::Result<Option<Self::Value>>;
```
meskill, line 61: I think it should be
```suggestion
    ) -> anyhow::Result<()>;
```

otherwise you're introducing non-compatible behaviour for memory and cloudflare implementations since ttlCache return Some or None depending on the presence of the value in cache already and cloudflare always returns Some.
It'll be simpler for now just return nothing

tusharmath (approved):

PR #1029, opened 2024-01-25
chore(cf): store app ctx to map

ssddOnTop:
Original issue: #750


tusharmath:
I don't think we need this anymore. Let's use the showcase capability instead.

PR #1032, opened 2024-01-25
fix: use request batching inside an expression

meskill:
**Summary:**  
Fixes the panic bug when using batching request inside an @expr block

**Build & Testing:**

**Checklist:**


tests/http/config/batching-nested-expr.graphql
```diff
@@ -0,0 +1,42 @@
...
+  id: Int
+  title: String
+  body: String
+  userId: Int!
+  user: User
+    @expr(
+      body: {
+        if: {
+          cond: {const: {data: true}}
+          then: {http: {path: "/users", query: [{key: "id", value: "{{value.userId}}"}], groupBy: ["id"]}}
```
meskill, line 19: this was panicking at https://github.com/tailcallhq/tailcall/blob/7a82e6bc7d764180f6b2f8807b6242bd52beabf6/src/lambda/io.rs#L223

src/lambda/expression.rs
```diff
@@ -106,3 +107,101 @@ impl Eval for Expression {
     })
   }
 }
+
+impl Expression {
+  pub(crate) fn io_iter(&mut self) -> IOExpressionIterator {
```
meskill, line 112: special iterator that will walk through all levels of expression and return only IO operation. 
Made it only for io and not general just to solve the problem inside app_context since later we may consider some other options to work with that

tusharmath: Can you add some Rust Docs on why this exists and where should it be used. Ideally we would like to make sure its not used inside of expression evaluation.

tests/http/config/batching-nested-expr.graphql
tusharmath: thanks!

src/lambda/expression.rs
meskill: I didn't like that it was specific for io resolvers and I've rewrote this to another method that could be considered general for working with expressions (traversing over expression tree) and just use that method in app_context for specific case

meskill:
The bug was resolved in https://github.com/tailcallhq/tailcall/pull/1078
The test case is recreated in https://github.com/tailcallhq/tailcall/pull/1128

Issue #1038, opened 2024-01-26
feat: publish tailcall WASM build on NPM


tusharmath:
Publish tailcall on npm under 

- `@tailcallhq/wasm-browser`
- `@tailcallhq/wasm-node`

**Public API**

```js
import {GraphQLExecutor} from '@tailcallhq/tailcall'


// YML, JSON or GraphQL
const jsonConfig = " ... <tailcall config> ... "; 


// Use the builder to create the configuration
const builder = new TailcallBuilder().withJSONConfig(jsonConfig).withYMLConfig(ymlConfig);

// Build an instance of Tailcall, resolving configurations
const tailcall = await builder.build();


// Execute queries using the builder.
const query = "query {users {name}}"
await executor.execute({query})
```

**Technical Requirements**
- I should be able to install the package via npm
- I should be able to use the package in the browser directly
- I should be able to use the package via node.js

aabidsofi19:
can i get assigned to this , would love to work on this issue .

ayewo:
✋

ssddOnTop:
Can you explain more about the package `@tailcallhq/wasm-browser`? It's confusing, according to my understanding, there should be just 1 package published i.e. `@tailcallhq/wasm-node`which can be used in browser using nodejs. Or is it like we provide a js file (maybe via cdn) which can be used in browsers like 

```html
<!doctype html>
<html lang="en-US">
  <head>
    <meta charset="utf-8" />
    <title>hello-wasm example</title>
  </head>
  <body>
    <script type="module">
      import init, { GraphQLExecutor } from "http://lint/to/tailcall.js";
      init().then(() => {
          const config = " ... <tailcall config> ... "; // YML, JSON or GraphQL
          const executor = new GraphQLExecutor(config);
          const query = "query {users {name}}"
          
          await executor.execute({query})
      });
    </script>
  </body>
</html>
```

gabrieldemian:
is anyone still working on this? if not, I would like to get assigned

ologbonowiwi:
This `GraphQLExecutor` being exported, do you imagine it being a Rust source code compiled to WASM or being a JS layer that interfaces with the binary @tusharmath?

tusharmath:
Yes.

webbdays:
@coderabbitai

ssddOnTop:
waiting for completion of #1153 or should I start without the builder?

webbdays:
@tusharmath 
[...]

PR #1039, opened 2024-01-26
feat: WASM build on NPM

ssddOnTop:
**Summary:**  
Added wasm build that can be used in browser

### Node
Steps to publish:`cd node && npm run publish`
Steps to run example: `cd node/example && npm run serve`

### Browser
Steps to publish:`cd browser && npm run publish`
Steps to run example: `cd browser` and you can simply use `index.html`

refer: https://developer.mozilla.org/en-US/docs/WebAssembly/Rust_to_Wasm

**Issue Reference(s):**  
Fixes: #1038 
/claim 1038

**Build & Testing:**

**Checklist:**


node/src/cache.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    async fn get<'a>(&'a self, key: &'a Self::Key) -> anyhow::Result<Self::Value> {
+        Ok(self
+            .cache
+            .get(key.clone())
+            .ok_or(anyhow!("No such key found"))?
+            .value()
+            .clone())
+    }
+}
```
tusharmath, line 39: can use `src/cache.rs`

browser/index.html
```diff
@@ -0,0 +1,56 @@
...
+        "\n" +
+        "type Post {\n" +
+        "  id: Int!\n" +
+        "  userId: Int!\n" +
+        "  title: String!\n" +
+        "  body: String!\n" +
+        '  user: User @http(path: "/users/{{value.userId}}")\n' +
+        "}\n"
+      const source = ".graphql"
+      const executor = new GraphQLExecutor(schema, source)
```
tusharmath, line 45:
```suggestion
      const executor = new GraphQLExecutor({graphql: schema})
```

For other extensions

```js
const executor = new GraphQLExecutor({json: schema})
const executor = new GraphQLExecutor({yml: schema})
```

node/src/env.rs
```diff
@@ -0,0 +1,13 @@
...
+
+pub struct WasmEnv {}
+impl WasmEnv {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+impl EnvIO for WasmEnv {
+    fn get(&self, key: &str) -> Option<String> {
+        unimplemented!("Not needed for npm pkg")
```
tusharmath, line 11: Not ok. There should be a way to support this in the browser.

browser/index.html
tusharmath: Pass env here.

```js
new GraphQLExecutor({env: {foo: "bar"})
```

node/src/env.rs
ologbonowiwi: On node we do have process.env, ideally we should read from it

PR #1044, opened 2024-01-26
feat(aws): add AWS Lambda function

### LINKED ISSUE #874: Support serverless via AWS Lambda
AWS has support to run Rust binaries on serverless.

**Technical Requirements**
- Reuse code as much as possible.
- Create a separate binary for AWS serverless
- Update CI jobs to create AWS Lambda binaries.

**Additional Links**
https://github.com/aws-samples/serverless-rust-demo/

mogery:
**Summary:**  
This PR adds the ability to run tailcall on AWS Lambda.

**Issue Reference(s):**  
Fixes #874
/claim #874

**To-do:**
- [x] Implement configuration

**Build & Testing:**

**Checklist:**


mogery:
## Implementation details:
 - **Deployment to Lambda is done via Terraform.**
   - The GraphQL definition is bundled with the image at deploy-time.
 - There's no built-in KV store for AWS like there is in Cloudflare, so I built a simple Cache provider based on HashMap.
 - The current implementation of `LambdaFileIO` reads from the filesystem, instead of, say, an S3 bucket, since it's needed to retrieve the GraphQL definition from the image bundle.

## Questions:

### Should the Terraform deployment code live in this repo?

The DX of a Lambda deployment is pretty bad right now. A dev has to:
1. Install Rust, Terraform, Python and cargo-lambda
1. Add AWS credentials to environment variables
1. Clone this repo
1. Build the `lambda` crate with cargo-lambda
1. Write a GraphQL definition into `lambda/deploy/config.graphql`
1. Run `terraform init && terraform apply`

We could move this to an external repo, and I can make the CI build Lambda images and put it in Releases, and the Terraform config can be adapted to download it. With that, the DX would be the following:
1. Install Terraform
1. Add AWS credentials to environment variables
1. Clone that repo
1. Write a GraphQL definition into `lambda/deploy/config.graphql`
1. Run `terraform init && terraform apply`

This is, in my opinion, much better. It also lets devs fork the repo and manage their own deployment schema on GitHub.

Also, where has the documentation moved to? The link in the PR template is dead.

.github/workflows/ci.yml
```diff
@@ -27,6 +27,34 @@ jobs:
...
+          toolchain: stable
+          override: true
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install cargo-lambda
+        run: pip install cargo-lambda
```
meskill, line 53: why not from crates.io? https://crates.io/crates/cargo-lambda

lambda/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "lambda"
```
meskill, line 2: can we have more specific name, related to aws?

.github/workflows/ci.yml
mogery: From [cargo-lambda.info](https://www.cargo-lambda.info/guide/installation.html)
> WARNING
cargo-install compiles the binary in your system, which usually takes more than 10 minutes. This method doesn't install [Zig](https://ziglang.org/) either, which is a requirement if you want to cross compile packages from macOS or Windows to Lambda Linux environments.

meskill: ok...

mogery: It's not ideal. I'll look into alternatives.

meskill: let's not spend much time on this. If it works, it's ok and it actually simpler then installing zig or anything.

lambda/Cargo.toml
mogery: renamed to `aws-lambda`

mogery:
Adapted to the new `TargetRuntime`.

Cargo.toml
```diff
@@ -124,7 +124,7 @@ default = ["cli", "js"]
 
 
 [workspace]
-members = [".", "autogen", "cloudflare", "testconv"]
+members = [".", "autogen", "cloudflare", "lambda", "testconv"]
```
tusharmath, line 127:
```suggestion
members = [".", "autogen", "cloudflare", "aws-lambda", "testconv"]
```

lambda/deploy/lambda.tf
```diff
@@ -0,0 +1,131 @@
...
+    principal     = "apigateway.amazonaws.com"
+
+    # The /*/* portion grants access from any method on any resource
+    # within the API Gateway "REST API".
+    source_arn = "${aws_api_gateway_rest_api.tailcall.execution_arn}/*/*"
+}
+
+output "base_url" {
+    value = "${aws_api_gateway_deployment.tailcall.invoke_url}"
+}
```
tusharmath, line 131: Should this file be part of this repo?

lambda/deploy/example.graphql
```diff
@@ -0,0 +1,12 @@
...
+}
+
+type User {
+  id: Int
+  name: String
+}
+
+type Query {
+  user: User @http(path: "/users/1", baseURL: "http://jsonplaceholder.typicode.com")
+}
```
tusharmath, line 12: Do we need this file here?

lambda/deploy/lambda.tf
mogery: It's only there so we can test the Terraform stuff before `tailcall-on-aws` is made official. I can get rid of it.

lambda/deploy/example.graphql
mogery: It's only there so we can test the Terraform stuff before `tailcall-on-aws` is made official. I can get rid of it.

.github/workflows/ci.yml
```diff
@@ -33,6 +33,34 @@ jobs:
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./aws-lambda
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
```
meskill, line 48: please replace it with `actions-rust-lang/setup-rust-toolchain@v1` similar to other jobs

.github/workflows/ci.yml
```diff
@@ -282,11 +310,57 @@ jobs:
...
+    env:
+      GITHUB_TOKEN: ${{secrets.GITHUBTOKEN}}
+      APP_VERSION: ${{ needs.draft_release.outputs.create_release_name }}
 
+    steps:
+      - name: Checkout Current Branch (Fast)
+        uses: actions/checkout@v4
+
+      - name: Install Correct Toolchain
+        uses: dtolnay/rust-toolchain@stable
```
meskill, line 330: here is also need to replace action

.github/workflows/ci.yml
```diff
@@ -33,6 +33,34 @@ jobs:
     outputs:
       check_if_build: ${{ steps.check.outputs.check_if_build }}
 
+  build_lambda:
```
meskill, line 36: what's the purpose of this job? seems like we are building that package again on release anyway.

It's still could be useful for pr testing purposes, right? In that way we can call it `test_aws_build` and call it only for pr. And without `--release` flag so we don't spend much time on build that is thrown away

aws-lambda/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+        .clone()
+        .expect("AppContext not initialized yet, please wait");
+    let resp = handle_request::<GraphQLRequest>(to_request(event), app_ctx).await?;
+    Ok(to_response(resp)?)
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Error> {
+    tracing_subscriber::fmt()
+        .with_max_level(tracing::Level::TRACE)
```
meskill, line 33: by default it should be `WARN` 
and will be good to be able to specify it with env variable as it works with cli right now

meskill (changes requested):

aws-lambda/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+use tailcall::blueprint::Blueprint;
+use tailcall::config::reader::ConfigReader;
+use tailcall::http::{handle_request, AppContext};
+use tokio::sync::RwLock;
+
+mod http;
+mod runtime;
+
+lazy_static! {
+    static ref APP_CTX: RwLock<Option<Arc<AppContext>>> = RwLock::new(None);
```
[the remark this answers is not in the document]
mogery: Yes.

aws-lambda/src/main.rs
mogery: fixed the default. env variable is a bit complex because i need to add it to terraform instead. that's an issue for `tailcall-on-aws`

.github/workflows/ci.yml
```diff
@@ -282,11 +310,57 @@ jobs:
...
+      - name: Rename Binary with Target Name
+        run: |
+          pwd
+          cp target/lambda/aws-lambda/bootstrap target/lambda/aws-lambda/tailcall-aws-lambda-bootstrap
+
+      - name: Upload AWS Lambda Bootstrap Binary
+        uses: xresloader/upload-to-github-release@v1
+        with:
+          release_id: ${{ needs.draft_release.outputs.create_release_id }}
+          file: target/lambda/aws-lambda/bootstrap target/lambda/aws-lambda/tailcall-aws-lambda-bootstrap
```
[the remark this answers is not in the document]
mogery: Fixed in 991b3d5d36f733fc2d3ff9f18f3caba83c7b7d6f

.github/workflows/ci.yml
mogery: Fixed in 0256bfcbb3acffd6ed9f8ebf4e65a59fa91af274

mogery: Yes, it's for testing purposes.

.github/workflows/ci.yml
mogery: Fixed in 7747b20f9975ffa1dd077e08a559bfa27b0fb288

lambda/deploy/example.graphql
mogery: Resolved in eef1c91cba5171b1dc2a943a5eef5203a3464030

lambda/deploy/lambda.tf
mogery: Resolved in eef1c91cba5171b1dc2a943a5eef5203a3464030

Cargo.toml
mogery: Resolved in fedb0e1541ffdf23504c7e0c98ba800b92be3d16

aws-lambda/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+    let config = ConfigReader::init(runtime.clone())
+        .read("./config.graphql")
+        .await?;
+    let blueprint = Blueprint::try_from(&config)?;
+
+    let app_ctx = Arc::new(AppContext::new(blueprint, runtime));
+
+    APP_CTX.write().await.replace(app_ctx);
+
+    run(service_fn(function_handler)).await
```
meskill, line 50: we can just probably do smth like this 

```rust
    let app_ctx = Arc::new(AppContext::new(blueprint, runtime));

    run(service_fn(|event| async {
        let resp = handle_request::<GraphQLRequest>(to_request(event), app_ctx.clone()).await?;
        Ok::<Response<hyper::Body>, Error>(to_response(resp)?)
    }))
    ```
    
    and not create any rwlock

Comment (mogery) [reply aws-lambda/src/main.rs]: Good catch! Fixed in 8be53a85c9008168362a0a94167deb93f0684c5b

File: aws-lambda/src/http.rs [hunk 13]
```diff
@@ -0,0 +1,56 @@
...
+    Ok(response)
+}
+
+pub fn to_request(req: lambda_http::Request) -> hyper::Request<hyper::Body> {
+    let builder = hyper::Request::builder()
+        .method(req.method())
+        .uri(req.uri())
+        .body(hyper::Body::from(req.body().as_ref().to_vec()));
+
+    builder.unwrap()
```
meskill, line 51: can we be sure there wouldn't be any error?

consider to propagate error

aws-lambda/src/http.rs
```diff
@@ -0,0 +1,56 @@
...
+    response: hyper::Response<hyper::Body>,
+) -> anyhow::Result<lambda_http::Response<hyper::Body>> {
+    Ok(response)
+}
+
+pub fn to_request(req: lambda_http::Request) -> hyper::Request<hyper::Body> {
+    let builder = hyper::Request::builder()
+        .method(req.method())
+        .uri(req.uri())
+        .body(hyper::Body::from(req.body().as_ref().to_vec()));
```
meskill, line 49: `to_vec` will allocate new vec, do we need it, it could be probably enough to use just the ref

aws-lambda/src/http.rs
```diff
@@ -0,0 +1,56 @@
...
+        Default::default()
+    }
+}
+
+#[async_trait::async_trait]
+impl HttpIO for LambdaHttp {
+    async fn execute(&self, request: reqwest::Request) -> Result<Response<Bytes>> {
+        let client = self.client.clone();
+        let method = request.method().clone();
+        let url = request.url().clone();
```
meskill, line 31: do we need all this cloning? 
this in new allocations and cpu time to do cloning

aws-lambda/src/http.rs
```diff
@@ -0,0 +1,56 @@
...
+        let response = client.execute(request).await?.error_for_status()?;
+        let res = Response::from_reqwest(response).await?;
+        tracing::info!("{} {} {}", method, url, res.status.as_u16());
```
meskill, line 34: since you're propagating errors and use `error_for_status` here in status you'll see only success statuses that might be not what you've intended

aws-lambda/src/http.rs
```diff
@@ -0,0 +1,56 @@
...
+        Ok(res)
+    }
+}
+
+pub fn to_response(
+    response: hyper::Response<hyper::Body>,
+) -> anyhow::Result<lambda_http::Response<hyper::Body>> {
+    Ok(response)
+}
```
meskill, line 43: its usage unwraps and wraps its to the result again `Ok::<Response<hyper::Body>, Error>(to_response(resp)?)` so one of the part could be simplified. This function is probably redundant

meskill (changes requested):

aws-lambda/src/http.rs
mogery: We need it, cannot use `as_ref` as the `req` will be dropped after the function ends

mogery: Omitted `as_ref` altogether and using straight `to_vec` now.

aws-lambda/src/http.rs
mogery: Fixed by pre-generating a string with `format!` for logging before `request` is consumed.

aws-lambda/src/http.rs
mogery: Fixed

mogery:
Screenshot 2024-02-08 at 21 00 45
...wtf are these codecov errors

aws-lambda/src/http.rs
meskill: I think removing `error_for_status` could be a problem because we will consider that request as successful and will try to parse the body and at the end will respond with nulls instead of erroring

meskill:
> Screenshot 2024-02-08 at 21 00 45 ...wtf are these codecov errors

yeah, strange, but don't bother about it, it won't affect pr merge

aws-lambda/src/http.rs
mogery: Oh, yeah. Same code is in `cloudflare`. Re-added.

meskill (approved):

PR #1054, opened 2024-01-27
chore: add snapshot testing via `md` files

mogery:
**Summary:**  
This PR adds `tests/execution_spec.rs`, which uses snapshots via `insta` and `.md` specs to perform testing, like outlined in #983. It also adds `testconv`, an automatic test converter.

**To-do:**
- [x] Convert from `http_spec`
- [x] Convert from `graphql_spec`

**Issue Reference(s):**  
Fixes #983
/claim #983

**Build & Testing:**

- [x] I ran `cargo test` successfully.
- [x] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [x] I have added relevant unit & integration tests.
- [x] I have updated the [documentation] accordingly.
- [x] I have performed a self-review of my code.
- [x] PR follows the naming convention of `<type>(<optional scope>): <title>`

[documentation]: https://github.com/tailcallhq/tailcallhq.github.io/tree/develop/docs

PR #1058, opened 2024-01-28
fix(document): serialize @cache directives

mogery:
**Summary:**  
`@cache` directives are not serialized when calling `config.to_sdl()`. This PR adds logic to `into_document.rs` to fix this.

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/config/into_document.rs
+++ b/src/config/into_document.rs
@@ -182,19 +182,17 @@ fn config_document(config: &Config) -> ServiceDocument {
             extend: false,
             description: None,
             name: pos(Name::new(type_name.clone())),
-            directives: [
-                type_def
-                    .added_fields
-                    .iter()
-                    .map(|added_field: &super::AddField| pos(added_field.to_directive()))
-                    .collect::<Vec<_>>(),
-                if let Some(cache) = &type_def.cache {
-                    vec![pos(cache.to_directive())]
-                } else {
-                    vec![]
-                },
-            ]
-            .concat(),
+            directives: type_def
+                .added_fields
+                .iter()
+                .map(|added_field: &super::AddField| pos(added_field.to_directive()))
+                .chain(
+                    type_def
+                        .cache
+                        .as_ref()
+                        .map(|cache| pos(cache.to_directive())),
+                )
+                .collect::<Vec<_>>(),
             kind,
         })));
     }
--- a/src/config/into_document.rs
+++ b/src/config/into_document.rs
@@ -182,11 +182,18 @@ fn config_document(config: &Config) -> ServiceDocument {
             extend: false,
             description: None,
             name: pos(Name::new(type_name.clone())),
-            directives: type_def
-                .added_fields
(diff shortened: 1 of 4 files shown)


mogery:
This is all ready now. Should be a quick merge.

src/config/into_document.rs
```diff
@@ -182,11 +182,19 @@ fn config_document(config: &Config) -> ServiceDocument {
...
+                    .iter()
+                    .map(|added_field: &super::AddField| pos(added_field.to_directive()))
+                    .collect::<Vec<_>>(),
+                if let Some(cache) = &type_def.cache {
+                    vec![pos(cache.to_directive())]
+                } else {
+                    vec![]
+                },
+            ]
+            .concat(),
```
meskill, line 197: Could be

```suggestion
            directives: type_def
                .added_fields
                .iter()
                .map(|added_field: &super::AddField| pos(added_field.to_directive()))
                .chain(
                    type_def
                        .cache
                        .as_ref()
                        .map(|cache| pos(cache.to_directive())),
                )
                .collect::<Vec<_>>(),
```

But I wonder if that changes are really required. I don't think we have a way to specify cache for addedField with current SDL and there is no test case for this?

mogery: It came up in #1054, which does test for identity for http_spec too, where there's a test for this.

mogery: Also this is not the cache specified for the addedField. This is the cache specified for the type.

meskill: > Also this is not the cache specified for the addedField. This is the cache specified for the type.

yes, you're right, it's the type itself. Can we extend the test for cache on type as well?

mogery: There is no `graphql_spec` test for either (type or field) currently. `execution_spec` will check both.

mogery: My bad. Brain isn't working great today :D On it.

mogery: Huh. Caching was added in an... odd way, so specifying a `@cache` on the type doesn't apply it to the type itself, but to the fields directly. This means that the identity check fails:

```diff
  schema @server @upstream(baseURL: "http://jsonplacheholder.typicode.com") {
    query: Query
  }
 
  type Query {
    user: User @http(path: "/foo") @cache(maxAge: 300)
  }
 
< type User {
<   id: Int @cache(maxAge: 900)
<   name: String @cache(maxAge: 900)
> type User @cache(maxAge: 900) {
>   name: String
>   id: Int
  }
 ```

Should I re-write `@cache` logic to apply it to the type? Or should I write logic in `into_document` to assume type caching if every field has the same directive?

mogery:
Pushed a new commit that fixes the identity fail of type-level cache directives, and adds a test for it. Should be good to go 👍🏻

meskill (approved):

meskill:
Great work, thanks for the fix!

PR #1064, opened 2024-01-29
feat: create `@call` operator

### LINKED ISSUE #846: Create a `@call` operator

ologbonowiwi:
**Issue Reference(s):**  
Fixes #846 
/claim #846

**Build & Testing:**

**Checklist:**


tusharmath:
@ologbonowiwi I want to come back to this. Can you resolve conflicts once.

ologbonowiwi:
@tusharmath conflicts solved on this one

src/config/expr.rs
```diff
@@ -39,6 +39,10 @@ pub enum ExprBody {
     #[serde(rename = "graphQL")]
     GraphQL(GraphQL),
 
+    /// Reuses a resolver pre-defined on type `Query`
```
ologbonowiwi, line 42:
```suggestion
    /// Reuses a resolver pre-defined on `Query` or `Mutation`
```

examples/jsonplaceholder.graphql
```diff
@@ -24,5 +24,5 @@ type Post {
   userId: Int!
   title: String!
   body: String!
-  user: User @http(path: "/users/{{value.userId}}")
+  user: User @call(query: "user", args: {id: "{{value.userId}}"})
```
tusharmath, line 27: Let's add another example config for using `call`

src/blueprint/mustache.rs
```diff
@@ -101,14 +101,14 @@ impl<'a> MustachePartsValidator<'a> {
 }
 
 impl FieldDefinition {
-    pub fn validate_field(&self, type_of: &config::Type, config: &Config) -> Valid<(), String> {
+    pub fn validate_field(self, type_of: &config::Type, config: &Config) -> Valid<Self, String> {
```
tusharmath, line 104: Why are we making this change?

ologbonowiwi: Or open another PR to it

examples/jsonplaceholder.graphql
tusharmath: Also as side note, in this case `user` doesn't actually accept `args`

tests/execution/call-graphql-datasource.md
```diff
@@ -0,0 +1,87 @@
...
+schema
+  @server(port: 8000, graphiql: true, hostname: "0.0.0.0")
+  @upstream(baseURL: "http://jsonplaceholder.typicode.com", httpCache: true) {
+  query: Query
+}
+
+type Query {
+  posts: [Post] @http(path: "/posts")
+  user(id: Int!): User
+    @graphQL(baseURL: "http://upstream/graphql", name: "user", args: [{key: "id", value: "{{args.id}}"}])
```
tusharmath, line 15:
```suggestion
  user(id: Int!): User @http(path: "/users/{{args.id}}")
```
Let's use a simpler example here.

tests/execution/call-mutation.md
```diff
@@ -0,0 +1,85 @@
...
+
+type Query {
+  firstUser: User @http(method: "GET", path: "/users/1")
+}
+
+type User {
+  id: Int
+  name: String
+  post: Post
+    @call(mutation: "insertPost", args: {input: "{\"body\":\"user-body\",\"title\":\"user-title\",\"userId\":1}"})
```
tusharmath, line 36: Writing the input as string encoded JSON isn't the most convinient. I think we should be able to pass the value using the graphQL syntax as follows:
```graphql
@call(mutation: "insertPost", args: {input: {body: "user-body", title: "user-title", userId: 1}})
```

tests/execution/call-graphql-datasource.md
ologbonowiwi: But in this case, we'd not be testing call with `@graphql`

src/blueprint/operators/call.rs
```diff
@@ -0,0 +1,168 @@
...
+                return Valid::succeed(b_field);
+            };
+
+            compile_call(field, config, call, operation_type)
+                .map(|resolver| b_field.resolver(Some(resolver)))
+        },
+    )
+}
+
+pub fn compile_call(
```
tusharmath, line 31: I think this implementation can be massively simplified. We need to invent a new expression inside `Context`

```rust
pub enum Context {
    Value,
    Path(Vec<String>),
    
    // A new operator
    CallWith {
        args: HashMap<String, Vec<String>>,
        expression: Box<Expression>
    } 
}
```

Whenever we see a `call` config, we should use it's `args` field to modify the current args in context and then call the expression with the new context. That way we won't need to make changes here, when we add a new IO type.

src/blueprint/mustache.rs
tusharmath: It takes a reference and validates. That's a good design for a validator. Let's revert this please.

src/blueprint/operators/const_field.rs
```diff
@@ -35,23 +35,28 @@ pub fn compile_const(inputs: CompileConst) -> Valid<Expression, String> {
...
-            Ok(gql) => {
-                let validation = if validate {
-                    validate_data_with_schema(config_module, field, gql)
-                } else {
-                    Valid::succeed(())
-                };
-                validation.map(|_| Literal(data.to_owned()))
+    Valid::from(
+        DynamicValue::try_from(&value.clone()).map_err(|e| ValidationError::new(e.to_string())),
+    )
```
ologbonowiwi, line 40: The change was necessary as this code was repeated across multiple places.

src/blueprint/operators/call.rs
ologbonowiwi: Mentioned on Discord but updating it here also: tried to implement this but did not succeed. I do have a WIP on [feat/#846/create-call-operator-via-resolver-with-arg](https://github.com/ologbonowiwi/tailcall/tree/feat/%23846/create-call-operator-via-resolver-with-arg), though.

src/blueprint/mustache.rs
ologbonowiwi: Reverted.

tests/execution/test-call-operator-errors.md
```diff
@@ -0,0 +1,32 @@
...
+  id: Int!
+}
+
+type Post {
+  userId: Int!
+  withoutResolver: User @call(query: "userWithoutResolver", args: {id: "{{value.userId}}"})
+  withoutOperator: User @call(args: {id: "{{value.userId}}"})
+  urlMismatchHttp: User @call(query: "user", args: {})
+  argumentMismatchGraphQL: User @call(query: "userWithGraphQLResolver", args: {})
+  headersMismatchGraphQL: User @call(query: "userWithGraphQLResolver", args: {})
```
tusharmath, line 30:
```suggestion
  headersMismatchGraphQL: User @call(query: "userWithGraphQLResolver")
```
Args should be optional.

ologbonowiwi: > Args should be optional.

Changed. The `args` didn't had a default value, but I implemented it on [b199a05](https://github.com/tailcallhq/tailcall/pull/1064/commits/b199a05a64b910cf96a45a15edf3c4a2f6eab519)

> Error when no args are supplied

This is already being tested. The snapshot for this error is on [tests/snapshots/execution_spec__test-call-operator-errors.md_errors.snap](https://github.com/ologbonowiwi/tailcall/blob/feat/%23846/create-call-operator/tests/snapshots/execution_spec__test-call-operator-errors.md_errors.snap)

ologbonowiwi: > Changed. The args didn't had a default value, but I implemented it on [b199a05](https://github.com/tailcallhq/tailcall/pull/1064/commits/b199a05a64b910cf96a45a15edf3c4a2f6eab519)

Re-reading now, I'm not sure if this was the expected solution. Let me know if so. It made sense to make it optional as sometimes no arg is needed, and if it's the case, the user shouldn't be required to declare an "empty" `args` field.

src/blueprint/operators/call.rs
tusharmath: Adding a todo here — https://github.com/tailcallhq/tailcall/issues/1285

tusharmath:
Thanks @ologbonowiwi!

Looking forward to it's documentation on the website :)
/tip 100$

tusharmath (approved):

PR #1078, opened 2024-01-31
fix: caching values based on resolvers

shashitnak:
**Summary:** 
Caching the values based on resolvers

**Build & Testing:**

**Checklist:**

- [ ] Drop `cache` from `blueprint::FieldDefinition`. Leverage `Expression` to store cache meta information.


src/graphql/request_template.rs
```diff
@@ -118,6 +127,15 @@ impl RequestTemplate {
     }
 }
 
+impl<Ctx: PathGraphql + HasHeaders + GraphQLOperationContext> CacheKey<Ctx> for RequestTemplate {
+    fn cache_key(&self, ctx: &Ctx) -> anyhow::Result<u64> {
```
tusharmath, line 131: - Hash calculations should not throw errors.

src/http/request_template.rs
```diff
@@ -215,6 +218,19 @@ impl TryFrom<Endpoint> for RequestTemplate {
...
 }
 
+impl<Ctx: PathString + HasHeaders> CacheKey<Ctx> for RequestTemplate {
+    fn cache_key(&self, ctx: &Ctx) -> anyhow::Result<u64> {
+        let mut hasher = DefaultHasher::new();
+
+        self.method.hash(&mut hasher);
+
+        let url = self.create_url(ctx)?;
+        url.hash(&mut hasher);
```
tusharmath, line 228: Add additional headers that are available on the RequestTemplate.
Add unit tests for hash.

src/grpc/request_template.rs
```diff
@@ -32,6 +36,13 @@ pub struct RenderedRequestTemplate {
     pub operation: ProtobufOperation,
 }
 
+impl Hash for RenderedRequestTemplate {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.url.hash(state);
+        self.body.hash(state);
+    }
+}
```
tusharmath, line 44: Add unit tests for hash. Write a test where we guarantee if the url is diff the hash is diff, where the body is diff the hash is diff etc.

src/cache_key.rs
```diff
@@ -0,0 +1,3 @@
+pub trait CacheKey<Ctx> {
+    fn cache_key(&self, ctx: &Ctx) -> anyhow::Result<u64>;
+}
```
tusharmath, line 3: can move to `lambda::cache.rs`

src/blueprint/blueprint.rs
```diff
@@ -130,7 +129,6 @@ pub struct InputFieldDefinition {
 #[derive(Clone, Debug)]
 pub struct Cache {
     pub max_age: NonZeroU64,
-    pub hasher: DefaultHasher,
 }
```
tusharmath, line 132: drop `cache` from blueprint.

src/blueprint/blueprint.rs
```diff
@@ -130,7 +129,6 @@ pub struct InputFieldDefinition {
 #[derive(Clone, Debug)]
 pub struct Cache {
```
tusharmath, line 130: please drop cache from blueprint.

src/app_context.rs
```diff
@@ -90,6 +90,12 @@ impl AppContext {
                             }
                         }
                     }
+
+                    if let Some(Cache { max_age }) = field.cache.clone() {
+                        let expr = field.resolver.take();
+                        field.resolver =
+                            expr.map(|expr| Expression::Cached(Cached::new(max_age, expr)));
+                    }
```
tusharmath, line 98: can we move this logic into from_config piece?

src/lambda/cached.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct NonIOCache {
+    data: Arc<RwLock<Option<ConstValue>>>,
+    expr: Box<Expression>,
+}
+
+impl Cached {
+    pub fn new(max_age: NonZeroU64, expr: Expression) -> Self {
```
tusharmath, line 35:
```suggestion
    pub fn wrap(max_age: NonZeroU64, expr: Expression) -> Expression {
```

Move the IO check outside of Cached::wrap. You are anyways check for one condition which is — Does the field have a cache operator, add this second check about it the field has an IO operator. Essentially we are performing a "Cachability" test.

src/lambda/cached.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Clone, Debug)]
+pub enum Cached {
+    IOCache(IOCache),
+    NonIOCache(NonIOCache),
+}
+
+#[derive(Clone, Debug)]
+pub struct IOCache {
+    max_age: NonZeroU64,
+    expr: Box<Expression>,
```
tusharmath, line 25:
```suggestion
    expr: Box<Expression>,
    cache_key: Fn(Ctx) -> u64
```

Can we inject the cache key function directly into the IOCache? That way we won't need to pattern match and look further into what is the actual expression and get it's key.

src/lambda/cached.rs
tusharmath: Alternative way could be to move all the logic into wrap.
- Check if the resolver can be cached
- Check if there is an IO operation
- Identify the cache key
- Return an Expression that has access to the cache key generator function, the original expression and the TTL.

tusharmath (approved):

PR #1125, opened 2024-02-06
test(`@grpc`): add tests for method wrong format

ologbonowiwi:
- test: add tests for grpc wrong format


tests/execution/test-grpc-invalid-method-format.md
```diff
@@ -0,0 +1,26 @@
...
+
+#### server:
+
+```graphql
+schema {
+  query: Query
+}
+
+type Query {
+  news: NewsData @grpc(method: "abc.NewsService", baseURL: "http://localhost:4000")
```
tusharmath, line 13: Can we add an integration test with `baseURL` set in `@upstream`

tusharmath (approved):

PR #1154, opened 2024-02-08
chore: retry on codecov failing error

ologbonowiwi:


.github/workflows/ci.yml
```diff
@@ -80,7 +80,6 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
           files: lcov.info
-          fail_ci_if_error: true
```
meskill, line 83: we could miss some important information without coverage and will end up with rerunning the job

we can try rerun it automatically like it's done [here](https://github.com/tagatac/bagoup/pull/57/files)

.github/workflows/ci.yml
```diff
@@ -76,11 +76,15 @@ jobs:
         run: cargo llvm-cov --all-features --lcov --output-path lcov.info
 
       - name: Upload Coverage to Codecov
-        uses: codecov/codecov-action@v4
+        uses: Wandalen/wretry.action@v1
         with:
-          token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
-          files: lcov.info
-          fail_ci_if_error: true
+          action: codecov/codecov-action@v3
```
meskill, line 81: return back to `v4` =)

.github/workflows/ci.yml
```diff
@@ -76,11 +76,15 @@ jobs:
...
+          attempt_limit: 10
```
meskill, line 82: let's have like 3 attempts

meskill (dismissed):

.github/workflows/ci.yml
ologbonowiwi: changed, but 3 attempts likely will not solve the issue 100%, as usually codecov stays down for a couple of seconds/minutes.

tusharmath (approved):

PR #1171, opened 2024-02-09
fix(aws): construct proper URL for core + set logging level with .env file

mogery:
**Summary:**  
This (actually, finally) fixes the Lambda builds. There was an unforeseen change that had to do with how the core handles paths (equality instead of `ends_with` when checking for `/graphql` in path) with how API Gateway constructs URLs (cannot call the base without an API key, must use a stage and call `https://{host}/{stage}/graphql`). This PR adds extra logic to the request converter to construct core-compatible URLs.

**Build & Testing:**

**Checklist:**


mogery:
Here's some proof it actually works now (you can run it too):
```bash
$ curl -X POST -H "Content-Type: application/json" https://zznym6zvdc.execute-api.eu-central-1.amazonaws.com/live/graphql --data '{"query": "query { user { id } }"}'
{"data":{"user":{"id":1}}}
```

aws-lambda/src/main.rs
```diff
@@ -14,7 +14,8 @@ mod runtime;
 #[tokio::main]
 async fn main() -> Result<(), Error> {
     tracing_subscriber::fmt()
-        .with_max_level(tracing::Level::WARN)
+        // log everything here since logs can be filtered by level in CloudWatch.
+        .with_max_level(tracing::Level::TRACE)
```
tusharmath, line 18: Log Levels are set by env variables. Why do we need this?

mogery: Lambda logs are not set by env variables. It logs to AWS CloudFront. Every log should be ingested into CloudFront, where it can later be filtered, that is why I've configured the tracing subscriber to log everything.

tusharmath: In either case, I think this should be configured by the user and we shouldn't hard code it. May be via our terraform scripts.

mogery: I can add an env variable. This or next PR?

mogery: did it in this PR.

aws-lambda/src/main.rs
```diff
@@ -13,8 +15,17 @@ mod runtime;
 
 #[tokio::main]
 async fn main() -> Result<(), Error> {
+    let _ = dotenv();
+
+    let trace: tracing::Level = std::env::var("LOG_LEVEL")
```
amitksingh1490, line 20:
```suggestion
    let trace: tracing::Level = std::env::var("TC_LOG_LEVEL")
```
https://tailcall.run/docs/guides/logging/

aws-lambda/src/main.rs
```diff
@@ -13,8 +15,17 @@ mod runtime;
...
+        .ok()
+        .as_ref()
+        .and_then(|x| tracing::Level::from_str(x).ok())
+        // log everything by default since logs can be filtered by level in CloudWatch.
+        .unwrap_or(tracing::Level::TRACE);
```
amitksingh1490, line 25:
```suggestion
        .unwrap_or(tracing::Level::INFO);
```
to be consistent with our documentation

mogery: Read the comment above.

mogery: And [this](https://github.com/tailcallhq/tailcall/pull/1171#discussion_r1484304309).

tusharmath (approved):

PR #1172, opened 2024-02-09
fix: avoid `<package>.<service>` on URLs and search proto file by operation on extensions

ologbonowiwi:
When we updated the `@grpc` directive to use `@link` instead of loading from a `.proto` file, the service field was wrongfully changed, and it was including the package_name, which broke the generated URL for a given gRPC mock.

At some point, we also changed the tests (gRPC URLs do not include the package name by default), and these PRs rollbacks these bugs and restores the original behavior. 

edit: a proper URL for a given grpc file is `<host>:<port>/<service>/<method>`. The current released version is using `<host>:<port>/<package/id>.<service>/<method>`. That's what this PR aims to fix


tusharmath:
What are we doing here? Can you provide a brief description in the pull request.

src/blueprint/operators/grpc.rs
```diff
@@ -121,6 +121,7 @@ pub struct CompileGrpc<'a> {
 
 struct GrpcMethod {
     pub id: String,
+    pub package_and_service: String,
```
tusharmath, line 124: Why not just call it package? And keep as an array of strings?

tests/execution/grpc-batch.md
```diff
@@ -81,14 +81,14 @@ type News {
...
-    url: http://localhost:50051/news.NewsService/GetMultipleNews
+    url: http://localhost:50051/NewsService/GetMultipleNews
     body: \0\0\0\0\n\x02\x08\x02\n\x02\x08\x03
   response:
     status: 200
     body: \0\0\0\0t\n#\x08\x02\x12\x06Note 2\x1a\tContent 2\"\x0cPost image 2\n#\x08\x03\x12\x06Note 3\x1a\tContent 3\"\x0cPost image 3
 - request:
     method: POST
-    url: http://localhost:50051/news.NewsService/GetMultipleNews
+    url: http://localhost:50051/NewsService/GetMultipleNews
```
tusharmath, line 91: Should not change existing tests.

tusharmath (changes requested):

tests/execution/grpc-batch.md
ologbonowiwi: this test was wrongly changed before. the way it should work is /<service>/<method>

ologbonowiwi:
PR description updated @tusharmath

tests/execution/grpc-batch.md
ologbonowiwi: I catch this when looking into the @call operator tests. It was working and needed this change after the merge (because a bug was introduced earlier on)

src/blueprint/operators/grpc.rs
ologbonowiwi: We could call `package`, but it also holds the name of service

tests/execution/grpc-url-from-upstream.md
```diff
@@ -75,7 +75,7 @@ type News {
 ```yml
 - request:
     method: POST
-    url: http://localhost:50051/news.NewsService/GetAllNews
+    url: http://localhost:50051/NewsService/GetAllNews
```
tusharmath, line 78: Here is the actual expected URL. 

Screenshot 2024-02-16 at 7 26 06 PM

ologbonowiwi: But where do we have `a.b.c.d.` on the config?

examples/grpc.graphql
```diff
@@ -2,7 +2,7 @@
 schema
   @server(port: 8000, graphiql: true)
   @upstream(baseURL: "http://localhost:50051", httpCache: true, batch: {delay: 10})
-  @link(id: "news", src: "../src/grpc/tests/news.proto", type: Protobuf) {
+  @link(src: "../src/grpc/tests/news.proto", type: Protobuf) {
```
ologbonowiwi, line 5: removing the id as it's not used anymore

tests/execution/grpc-batch.md
meskill: package name could be the part of url actually if it's defined in the proto and it's defined in proto file.

take a look into

meskill (changes requested):

tests/execution/grpc-batch.md
ologbonowiwi: do you have any example of it that I can add to the tests?

ologbonowiwi:
Closing in favour of #1198

PR #1182, opened 2024-02-13
feat: add upstream request filter

tusharmath:
~The mainline implementation did not support making multiple HTTP calls in parallel. This implementation allows the dispatching of multiple commands from the client (JS) side.~

This is a slight improvement over the existing API. It supports writing simple js expressions 
that can be invoked to modify the outgoing request's - headers, method and URL. After many experiments, we have realized that data transformations on the request & response body our extremely expensive. This is not because JS is slow, it's because we need to move a huge object from Rust to JS memory, which requires a massive level of serde operations.


examples/scripts/tsconfig.json
ologbonowiwi, line 1: We could simplify this file by removing the comments

ologbonowiwi (changes requested):

src/cli/javascript/js_request.rs
```diff
@@ -50,12 +50,9 @@ impl TryFrom<reqwest::Request> for JsRequest {
...
             })
             .collect::<BTreeMap<String, String>>();
-        let body = req.body().map(|body| {
-            let bytes = body.as_bytes().unwrap_or_default();
-            Bytes::from_iter(bytes.to_vec())
-        });
 
-        Ok(JsRequest { url, method, headers, body })
+        // NOTE: We don't pass body to worker for performance reasons
+        Ok(JsRequest { url, method, headers, body: None })
```
ologbonowiwi, line 55: We actually need `body` then? From my findings, the only place we actually use it is on tests.

tusharmath: Request from rust will not contain body. Request from JS can contain body.

examples/scripts/tsconfig.json
meskill, line 1: do we need this file?

src/cli/javascript/runtime.rs
```diff
@@ -44,37 +51,41 @@ impl Runtime {
...
-        let local_value = v8::Local::new(scope, value);
-        let closure: v8::Local<v8::Function> = local_value.try_into()?;
-        let input = serde_v8::to_v8(scope, message)?;
-        log::debug!("js input: {:?}", input);
-        let null_ctx = v8::null(scope);
+        let js_runtime = &mut runtime.js_runtime;
+        let scope = &mut js_runtime.handle_scope();
+        let global = v8::Local::<v8::Object>::try_from(v8::Local::new(scope, &runtime.global))?;
+        let args = serde_v8::to_v8(scope, event)?;
+        let fn_server_emit = v8::String::new(scope, name.as_str()).unwrap();
```
meskill, line 79:
```suggestion
        let fn_server_emit = v8::String::new(scope, name.as_str())?;
```

examples/scripts/echo.js
```diff
@@ -1,17 +1,11 @@
...
-const asyncId = () => {
-  i++
-  return i
-}
-function onEvent(event) {
-  if (event.message.request) {
-    event.id = asyncId()
+function onRequest(request) {
+  return {
+    response: {
```
amitksingh1490, line 3: this should return a request

PR #1198, opened 2024-02-14
fix(grpc): allow nested package names

meskill:
**Summary:**  
Fix usage of package definition inside proto file. Separate it from link id definition

**Build & Testing:**

**Checklist:**

### DIFF
--- a/examples/grpc.graphql
+++ b/examples/grpc.graphql
@@ -8,14 +8,9 @@ schema
 
 type Query {
   news: NewsData! @grpc(method: "news.news.NewsService.GetAllNews")
-  newsById(news: NewsInput!): News!
-    @grpc(method: "news.news.NewsService.GetNews", body: "{{args.news}}")
+  newsById(news: NewsInput!): News! @grpc(method: "news.news.NewsService.GetNews", body: "{{args.news}}")
   newsByIdBatch(news: NewsInput!): News!
-    @grpc(
-      method: "news.news.NewsService.GetMultipleNews"
-      body: "{{args.news}}"
-      groupBy: ["news", "id"]
-    )
+    @grpc(method: "news.news.NewsService.GetMultipleNews", body: "{{args.news}}", groupBy: ["news", "id"])
 }
 
 type News {
--- a/examples/grpc.graphql
+++ b/examples/grpc.graphql
@@ -7,15 +7,13 @@ schema
 }
 
 type Query {
-  news: NewsData! @grpc(service: "news.NewsService", method: "GetAllNews", protoId: "news")
+  news: NewsData! @grpc(method: "news.news.NewsService.GetAllNews")
   newsById(news: NewsInput!): News!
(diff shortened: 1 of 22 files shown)


examples/grpc.graphql
```diff
@@ -7,17 +7,10 @@ schema
 }
 
 type Query {
-  news: NewsData! @grpc(service: "news.NewsService", method: "GetAllNews", protoId: "news")
-  newsById(news: NewsInput!): News!
-    @grpc(service: "news.NewsService", method: "GetNews", body: "{{args.news}}", protoId: "news")
+  news: NewsData! @grpc(method: "news.news.NewsService.GetAllNews")
+  newsById(news: NewsInput!): News! @grpc(method: "news.news.NewsService.GetNews", body: "{{args.news}}")
```
amitksingh1490, line 11: keep the example simple do this change for tests only,

src/blueprint/operators/grpc.rs
```diff
@@ -131,16 +131,22 @@ impl TryFrom<String> for GrpcMethod {
...
-        let service = format!("{}.{}", id, method[1]);
-        let name = method[2].to_string();
+        // TODO: some of implementation (like tonic) will use file name as default package name
+        // if the package is not specified in proto explicitly, that affect the request url.
+        // but `prost_reflect` crate won't find the service if we try to specify the package name if there is
+        // no package definition in proto file, so we may end up with case that we forced to specify
+        // package name to be able to get response, but in the same time we can't do that
+        // because prost won't find that definition for same proto file.
+        let service = method[1..method.len() - 1].join(".");
+        let name = method.last().unwrap().to_string();
```
amitksingh1490, line 149: make this safe. Handle errors

examples/grpc.graphql
meskill: the example is broken and should be updated

src/blueprint/operators/grpc.rs
meskill: we have check for `method.len() < 3` earlier, this is not necessary

ssddOnTop: I think we should get change the purpose of link id. It should give user ability to define a name in case there is no file name or package provided. We should by default use package id everywhere.

something like https://github.com/tailcallhq/tailcall/blob/b543c748ea793696da0552e8627b48d05008b2c3/src/config/reader.rs#L119 

check #1227

meskill: > but this will force users to define methods like `news.news.NewsService.GetAllNews` (from modified grpc.graphql example) which is incorrect if package id is `news`. Method should be defined as `news.NewsService.GetAllNews`

it'll work here since the first news is the name of the link, and only `news.NewsService.GetAllNews` will be used as service name. 
But that looks awkward actually, let's discuss your suggestion, I left some comments there

meskill:
In favor of https://github.com/tailcallhq/tailcall/pull/1227/files

PR #1199, opened 2024-02-15
fix(link): bug in relative path

ssddOnTop:


meskill (approved):

src/config/reader.rs
```diff
@@ -72,7 +72,7 @@ impl ConfigReader {
     async fn ext_links(
         &self,
         mut config_module: ConfigModule,
-        path: Option<String>,
+        path: Option<PathBuf>,
```
meskill, line 75: as I understand we changed this behaviour from passing file path to passing directory, right?

at least other naming for argument could be helpful

src/config/reader.rs
```diff
@@ -205,11 +197,10 @@ impl ConfigReader {
             let schema = &file.content;
 
             // Create initial config module
+            let mut parent_dir = PathBuf::from(&file.path); // path to config
+            parent_dir.pop(); // parent dir
```
meskill, line 201: that's strange that we need to repeat that conversion to dir path again and again

meskill (changes requested):

src/config/reader.rs
ssddOnTop: I think you are referring to parent_dir.pop() being done twice. 1st one is to remove `*.config` from original path, and 2nd time it's done in `LinkType::Config` because loading after loading config via link operator, we need to pop `*.config` once again

meskill: yeap, and also that repeats in tests. 

Having the same code all over the place feels like degradation and easier to mistake code.

src/config/reader.rs
```diff
@@ -117,7 +107,7 @@ impl ConfigReader {
                     if !config.links.is_empty() {
                         config_module = config_module.merge_right(
                             &self
-                                .ext_links(ConfigModule::from(config), Some(source.path))
+                                .ext_links(ConfigModule::from(config), parent_dir)
```
meskill, line 110: is this part right? it was path to source of the link itself with probably its parent_dir, but now we reuse the same parent_dir. It's probably related to case for links inside already linked config

src/config/reader.rs
```diff
@@ -205,11 +195,10 @@ impl ConfigReader {
             let schema = &file.content;
 
             // Create initial config module
+            let config_path = PathBuf::from(&file.path); // path to config
+            let parent_dir = config_path.parent();
```
meskill, line 199:
```suggestion
            let parent_dir = Path::new(&file.path).parent();
```
will it work?

ssddOnTop: `temporary value dropped while borrowed`

ssddOnTop: maybe `Path::new(&file.path).parent()` should work

meskill (approved):

PR #1215, opened 2024-02-18
feat(grpc): add reflection support 

### LINKED ISSUE #868: add support for using reflection API in grpc
Add support for using reflection API instead of providing proto path in `@grpc` operator.

**Links**
https://github.com/grpc/grpc/blob/master/doc/server-reflection.md

ssddOnTop:
**Summary:**  

**Issue Reference(s):**  
Fixes #868
/claim #868

## Pending
- [ ] Test in execution spec (related issue: #1405)

## Summary by CodeRabbit

- **New Features**
	- Introduced gRPC reflection support, enabling interaction with gRPC reflection servers.
	- Added the ability to query server reflection information, enhancing the dynamic discovery of services.
	- Expanded the application's configuration capabilities to support new gRPC reflection types.
- **Enhancements**
	- Improved gRPC operation handling with updated serialization methods.
	- Enhanced configuration reading and processing for diverse link types.
- **Tests**
	- Implemented comprehensive testing for gRPC reflection functionality, including mock server interactions and schema assertions.
- **Documentation**
	- Provided examples and test documentation for utilizing gRPC reflection in querying news data.


src/config/grpc_fetch.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    let mut methods = vec![];
+    let mut url: url::Url = url.parse()?;
+    url.set_path("grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo");
+
+    let mut req = reqwest::Request::new(Method::POST, url);
+    *req.body_mut() = Some(reqwest::Body::from(b"\0\0\0\0\x02:\0".to_vec())); // magic :)
+
+    let resp = target_runtime.http.execute(req).await?;
+    let body = resp.body.as_bytes();
```
meskill, line 38: instead of using raw reqwest and create body manually consider using `TargetRuntime` and `grpc::RequestTemplate`.

For RequestTemplate use the `protobuf_set` you've already created and its `convert_input` method to construct bytes for body from the json string. 
How it works you can find in file `src/grpc/protobuf.rs` and its tests.
e.g. your `b"\0\0\0\0\x02:\0"` is the same as `protobuf_set.convert_input("{listServices: ""}")`

src/config/grpc_fetch.rs
```diff
@@ -0,0 +1,176 @@
...
+    target_runtime: &TargetRuntime,
+    service: &str,
+) -> Result<FileDescriptorProto> {
+    let protobuf_set = get_protobuf_set()?;
+
+    let mut url: url::Url = url.parse()?;
+    url.set_path("grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo");
+
+    let mut req = reqwest::Request::new(Method::POST, url);
+    *req.body_mut() = Some(reqwest::Body::from(format!("\0\0\0\0\x12\"\x10{service}")));
```
meskill, line 87: that won't work in other cases since bytes for protobuf encodes length of string etc.
But if you migrate to `RequestTemplate` and its `to_request` method it will work fine and you will just pass smth like `format!("\{fileContainingSymbol: \"{service}\"\})`

meskill (changes requested):

src/config/grpc_fetch.rs
```diff
@@ -0,0 +1,331 @@
...
+                "grpc.reflection.v1alpha.ServerReflection".to_string()
+            ]
+            .to_vec(),
+            resp
+        );
+
+        http_reflection_list_all.assert();
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
ssddOnTop: generate additional test cases covering edge cases and error conditions

tests/execution_spec.rs
```diff
@@ -1023,6 +1024,22 @@ async fn assert_spec(spec: ExecutionSpec) {
     log::info!("[{}] {} ... ok", spec.name, spec.path.display());
 }
 
+fn start_mock_server() -> Server {
+    Server::new_with_opts(ServerOpts { host: "127.0.0.1", port: 50051, assert_on_drop: true })
+}
```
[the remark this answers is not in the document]
ssddOnTop: this is intentional. marking this as resolved

src/config/reader.rs
```diff
@@ -145,18 +147,132 @@
...
+
+                        let mut resolved_descriptor =
+                            self.resolve_descriptors(file_descriptor_proto).await?;
+                        file_descriptor_set.file.append(&mut resolved_descriptor);
+
+                        config_module
+                            .extensions
+                            .grpc_file_descriptors
+                            .push(Content { id: Some(id), content: file_descriptor_set });
+                    }
```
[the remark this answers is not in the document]
ssddOnTop: suggest some code

src/config/reader/reader.rs
```diff
@@ -145,18 +147,126 @@ impl ConfigReader {
...
+                LinkType::ReflectionWithFileName => {
+                    let mut file_descriptor_set = FileDescriptorSet::default();
+
+                    let file_descriptor_proto = get_by_proto_name(
+                        config_link.src.as_str(),
+                        &self.runtime,
+                        &config_link
+                            .id
+                            .clone()
+                            .context("Expected link id with method name but found none")?,
```
meskill, line 183: this error is showed to user in case of errors but in such form it says more about the code and semantics not about the config and expected actions.

Consider updating errors message accross pr with clear messages from user perspective and possible actions to fix. Also trace for definition could be helpful.
i.e. "id for link is not specified. When using grpc ... it is mandatory"

src/config/link.rs
```diff
@@ -12,6 +12,9 @@ pub enum LinkType {
     Script,
     Cert,
     Key,
+    ReflectionWithFileName,
+    ReflectionWithService,
+    ReflectionAllFiles,
```
meskill, line 17: do we really need 3 different type for reflection? 

Feels like ReflectionAllFiles could solve all of the problems. And if you have concerns about redundant data loaded we can load only services that are used across `@grpc` directive

src/config/link.rs
```diff
@@ -12,6 +12,9 @@ pub enum LinkType {
     Script,
     Cert,
     Key,
+    ReflectionWithFileName,
```
meskill, line 15:
```suggestion
    GrpcReflection,
```
just `Reflection` is too general term

src/config/reader/grpc_fetch.rs
```diff
@@ -0,0 +1,404 @@
...
+        operation_type: Default::default(),
+    };
+
+    let ctx = TargetRuntimeContext {
+        runtime: target_runtime,
+        vars: &Default::default(),
+        headers: Default::default(),
+    };
+
+    let req = req_template.render(&ctx)?.to_request()?;
```
meskill, line 93: that is very similar code across 3 function here and that could be moved to separate function

src/runtime.rs
```diff
@@ -16,6 +21,34 @@ pub struct TargetRuntime {
...
+        }
+
+        path.split_first()
+            .and_then(|(head, tail)| match head.as_ref() {
+                "vars" => self.vars.get(tail[0].as_ref()).map(|v| v.into()),
+                "env" => self.runtime.env.get(tail[0].as_ref()).map(|v| v.into()),
+                _ => None,
+            })
+    }
+}
```
meskill, line 44: in main changes with similar structure have been landed here `https://github.com/tailcallhq/tailcall/blob/main/src/config/reader_context.rs` so you can reuse that

tests/execution_spec.rs
```diff
@@ -1023,6 +1024,22 @@ async fn assert_spec(spec: ExecutionSpec) {
     log::info!("[{}] {} ... ok", spec.name, spec.path.display());
 }
 
+fn start_mock_server() -> Server {
```
meskill, line 1027: why do we need additional server for tests and why mocks are not sufficient for this?

meskill (changes requested):

tests/execution_spec.rs
ssddOnTop: I don't think it's good idea to put 1.4kb of data in every reflection related test.. maybe we can design some new proto file for tests but the descriptor file will always be huge

src/config/link.rs
ssddOnTop: grpc directive could be defined anywhere.. it would be difficult to cover all cases specially if we make any changes in future..

This will also increase time to compile the config.. we'll have to iterate for n^2 twice (once for conversion to blueprint and other to search all grpc directives)

src/config/reader/reader.rs
ssddOnTop: let's see if we continue with `Reflection*` or switch to single `GrpcReflection`. I'll update the errors later

src/config/link.rs
meskill: could you explain why we need 3 different types for this and how the issue you've mentioned is solved?

ssddOnTop: We are directly asking for pkg id or file name as link id.. so we just have to make request for specific files defined by the user instead of iterating all over config to find grpc method.. and thats’s the purpose of link right?

here is correction for my comment above: the complexity will be 2^n as grpc is field directive. In that case, I dont think we’ll need to change code in future but again but schema compilation will be slower. 

But at this point, I dont think iterating over fields will take A LOT more time, either of the approaches are fine. Lmk what do you think

tests/execution_spec.rs
meskill: current approach is case specific. if we add feature to specify mocks as separate files will it solve the issue?

ssddOnTop: can you give an example?

PR #1221, opened 2024-02-20
chore:  make `resolve_descriptors` accept `FileDescriptorProto`

ssddOnTop:


src/config/reader.rs
```diff
@@ -247,7 +246,7 @@ impl ConfigReader {
             }
         }
 
-        descriptors.insert(proto_path, parent_proto);
+        descriptors.insert(String::new(), parent_proto);
```
meskill, line 249: that seems odd to put into hashmap the entry with empty key. If it's intentional how this should work?

src/config/reader.rs
```diff
@@ -247,7 +246,7 @@ impl ConfigReader {
             }
```
meskill, line 246: there is still `self.read_proto(import)` inside `while let`. How will that be handled for the reflection case?

src/config/reader.rs
```diff
@@ -117,7 +117,7 @@ impl ConfigReader {
                 }
                 LinkType::Protobuf => {
                     let descriptors = self
-                        .resolve_descriptors(HashMap::new(), source.path)
+                        .resolve_descriptors(HashMap::new(), self.read_proto(&source.path).await?)
```
meskill, line 120: not related to changes, but why do we pass `HashMap` to the call instead of creating it inside the `resolve_descriptors` itself?

src/config/reader.rs
ssddOnTop: Not sure about that as of now. Maybe I’ll raise another pr if it can be divided into smaller part

src/config/reader.rs
ssddOnTop: earlier the implementation was different, someone refactored it and now the function is returning a hashmap, we don't need to take hashmap as arg. I'll raise separate pr for later

src/config/reader.rs
ssddOnTop: we never need the key outside the function, the key just used to avoid remove any cycle dependencies or adding same file several times

meskill: in that case I suggest to keep such data itself that function only and accept/return only the data that is required i.e. return the list of descriptors instead of map to prevent any misuses

ssddOnTop: I removed hashmap from arg and made the function return vector

meskill (approved):

PR #1227, opened 2024-02-21
feat: make package mandatory in proto files

### LINKED ISSUE #1226: bug: make proto package id seperable by `.`
Currently method is expected to be in form of: `<package/proto_id>.<service>.<method>`. Currently it works only if the package id is in the form of `package foo;` but if the package id is in form of, say `package foo.bar;` then it will throw an error that `Invalid method format: {}. Expected format is <package/proto_id>.<service>.<method>`.

## Challenges
1. Currently we depend upon `id` field in `link` operator. But ideally we should use package id by default.
2. Currently only `LinkType::Protobuf` is defined in which we rely upon the `id` field but with implementation of grpc server reflection, we have ability to load all proto files or load proto file based on `package id`. So in those cases we would not have any id or we might just have `package id`. 
3. We need to define when to use package id, file name or link id respectively.

## Example
let there be some valid proto file in which package id is separated by `.`
```proto
syntax = "proto3";

package package news.foo;
```

Then gql query with `@grpc` operator should look like
```graphql
type Query {
  news: NewsData! @grpc(method: "news.foo.Service.Method")
}
```  

Tailcall is unable to parse with exception `Invalid method format: news.foo.Service.Method. Expected format is <package/proto_id>.<service>.<method> [at Query.news.@grpc]` because it expects the package name to be single string without `.`, which is a faulty implementation.

ssddOnTop:
**Issue Reference(s):**  
Fixes #1226


src/config/reader.rs
```diff
@@ -116,20 +116,32 @@ impl ConfigReader {
...
                 }
                 LinkType::Protobuf => {
-                    let mut descriptors = self
-                        .resolve_descriptors(self.read_proto(&source.path).await?)
-                        .await?;
+                    let parent_descriptor = self.read_proto(&source.path).await?;
+                    let id = parent_descriptor
+                        .package // we should ideally take the package name as default id
+                        .clone()
+                        .unwrap_or(
```
meskill, line 123: in case if package name is not defined in proto file to fetch the service with protox we need to pass just pure service_name without any package. But to link `@grpc` with specific `@link` we need some kind of id.
we need to add more tests for this with different cases of package naming here

src/config/reader.rs
```diff
@@ -116,20 +116,32 @@ impl ConfigReader {
                     }
                 }
                 LinkType::Protobuf => {
-                    let mut descriptors = self
-                        .resolve_descriptors(self.read_proto(&source.path).await?)
-                        .await?;
+                    let parent_descriptor = self.read_proto(&source.path).await?;
+                    let id = parent_descriptor
+                        .package // we should ideally take the package name as default id
```
meskill, line 121: there was some attempt to specify fully qualified method name without linking to file explicitly and just search for the required definition across all files. 

Here https://github.com/tailcallhq/tailcall/pull/1172/files#diff-292fd7f2de6455aa7a8afbf0b7a1d0f4a7538f4f1c92ba3be3e29b914d749867
wdyt about this approach?

src/config/reader.rs
```diff
@@ -116,20 +116,32 @@ impl ConfigReader {
...
+                        .unwrap_or(
+                            config_link
+                                .id
+                                .to_owned() // if there is no package id then we use link id
+                                .unwrap_or(
+                                    parent_descriptor
+                                        .name // if none of them is available,
+                                        // then we use file name.
+                                        .clone()
+                                        .unwrap_or_default(),
```
meskill, line 132: I don't understand how it will work from user perspective.
if we have package inside proto file it seems ok, but in other cases I need really think link ids or filenames and they overrides each other. 
Also, the `name` is the path to file, not the file name. And we may end up with empty string here that is even more unclear how it should work.

meskill (changes requested):

src/config/reader.rs
ssddOnTop: `name` must be proto name, we have a faulty implementation of `protox_parse` as I told on discord.
https://github.com/tailcallhq/tailcall/blob/f4306dfcd629d0b2e477dda2ce4096ba18083965/src/config/reader.rs#L263

it just puts the path as name, but the path might be a url or full path, we need some way to avoid that. I think we can have 3 cases, if the proto file have package, then we just use it everywhere, else we check if the path ends with `*.proto`, else we throw an exception about naming or package id

src/config/reader.rs
ssddOnTop: if we fix the `protox_parse::parse` issue (in comments below) then we can grantee some kind of id. Because in case of server reflection, we have 3 ways, but in all the ways it's expected that the proto file must have package name or file name.

but in our case we have freedom to load proto files over url.. we need to think of some wasy to give proper file name instead of path

ssddOnTop: but I think that LinkType::Protobuf will always have `id` field. So we won't need to care about anything. But user might get confused because sometimes link id would be used or maybe package id or file name

tusharmath:
PR needs some unit tests.

src/config/reader.rs
meskill: when trying to fix the original issue with package name I faced the issue described in the [comment](https://github.com/tailcallhq/tailcall/pull/1198/files#diff-292fd7f2de6455aa7a8afbf0b7a1d0f4a7538f4f1c92ba3be3e29b914d749867R142).

I think the most easiest way for now is to just require package in proto and throw meaningful validation error otherwise. That will clear and obvious way to work with without any mentioned gotchas for backend implementation. And depending on future use we'll think if that require changes cuz proto files without package could be not very used for cases except simple testing

src/config/reader.rs
meskill: I agree

meskill (approved):

src/blueprint/operators/grpc.rs
```diff
@@ -249,7 +256,7 @@ mod tests {
         assert!(result.is_err());
         assert_eq!(
             result.err().unwrap(),
-            ValidationError::new("Invalid method format: package_name.ServiceName. Expected format is <package/proto_id>.<service>.<method>".to_string())
+            ValidationError::new("Invalid method format: package_name.ServiceName. Expected format is <package>.<service>.<method>".to_string())
```
tusharmath, line 259:
```suggestion
            ValidationError::new("Invalid method format: package_name.ServiceName. Expected format is <package>.<service>/<method>".to_string())
```

ssddOnTop: marking it as resolved as it was resolved on discord.

src/config/reader.rs
```diff
@@ -116,20 +117,26 @@ impl ConfigReader {
...
-                        .await?;
+                    let parent_descriptor = self.read_proto(&source.path).await?;
+
+                    let id = Valid::from_option(
+                        parent_descriptor.package.clone(),
+                        format!(
+                            "Package name is not defined for proto file: {:?} with link id: {:?}",
+                            parent_descriptor.name, config_link.id
+                        ),
+                    )
```
tusharmath, line 128: We should instead set `trace` in validation error.

tusharmath: Need a test where we throw up this error.

src/config/reader.rs
```diff
@@ -292,6 +299,17 @@ mod test_proto_config {
...
     }
 
+    #[tokio::test]
+    async fn test_proto_no_pkg() -> Result<()> {
+        let runtime = crate::runtime::test::init(None);
+        let reader = ConfigReader::init(runtime);
+        let mut proto_no_pkg = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        proto_no_pkg.push("tests/unit_tests/proto_no_pkg.graphql");
+        let config_module = reader.read(proto_no_pkg.to_str().unwrap()).await;
+        assert!(config_module.is_err());
```
tusharmath, line 309: assert the message, description, trace etc.

tests/unit_tests/proto_no_pkg.graphql
```diff
@@ -0,0 +1,17 @@
...
+type Query {
+  news: News! @grpc(method: "news.NewsService.GetAllNews")
+}
+
+type News {
+  id: Int
+  title: String
+  body: String
+  postImage: String
+}
```
tusharmath, line 17: the root tests directory is for integration tests.

ssddOnTop: where should I put these fils?

src/config/reader.rs
```diff
@@ -292,6 +299,17 @@ mod test_proto_config {
             .unwrap();
     }
 
+    #[tokio::test]
+    async fn test_proto_no_pkg() -> Result<()> {
+        let runtime = crate::runtime::test::init(None);
+        let reader = ConfigReader::init(runtime);
+        let mut proto_no_pkg = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        proto_no_pkg.push("tests/unit_tests/proto_no_pkg.graphql");
```
tusharmath, line 307: Let's load it from `src/grpc/tests/proto` directory instead.

ssddOnTop: will have to change existing tests then

src/blueprint/operators/grpc.rs
```diff
@@ -129,20 +128,22 @@ impl TryFrom<String> for GrpcMethod {
...
+        let method: Vec<&str> = value.rsplitn(3, '.').collect();
 
-        if method.len() != 3 {
-            return Err(ValidationError::new(format!(
-                "Invalid method format: {}. Expected format is <package/proto_id>.<service>.<method>",
+        match &method[..] {
+            &[name, service, id] => {
+                let method = GrpcMethod {
+                    id: id.to_owned(),
+                    service: format!("{id}.{service}"),
```
tusharmath, line 137: Service should only have `service`.

src/config/reader.rs
ssddOnTop: there is a test which reads all available proto files to check if the logic works for nested files and stuff

src/blueprint/operators/grpc.rs
ssddOnTop: I did not edit any logic here.. it matches main branch.

ssddOnTop: will have to change the implementation for that

src/config/link.rs
```diff
@@ -47,3 +47,17 @@ pub struct Link {
...
+            format!(
+                "@link(id: {}, src: {}, type: {:?})",
+                id, self.src, self.type_of
+            )
+        } else {
+            format!("@link(src: {}, type: {:?})", self.src, self.type_of)
+        };
+        f.write_str(&format)
+    }
+}
```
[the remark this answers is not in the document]
ssddOnTop: Not really but it shows exact link that the user have defined so it would be easier for user to find the problematic link

PR #1238, opened 2024-02-23
chore: fix `cargo test <module name> panics`

### LINKED ISSUE #1236: chore: cargo test <module name> panics
Ran the following command:
```bash
cargo test cli::javascript
```

**Actual**
`execution_spec.rs` is also test

```
    Finished test [unoptimized + debuginfo] target(s) in 0.20s
     Running unittests src/lib.rs (target/debug/deps/tailcall-e896f29f34ba5636)

running 6 tests
test cli::javascript::js_response::test::test_js_response_with_defaults ... ok
test cli::javascript::js_response::test::test_unusual_headers ... ok
test cli::javascript::js_response::test::test_from_js_response ... ok
test cli::javascript::js_request::tests::test_reqwest_request_to_js_request ... ok
test cli::javascript::js_request::tests::test_js_request_to_reqwest_request ... ok
test cli::javascript::js_response::test::test_to_js_response ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 249 filtered out; finished in 0.00s

     Running unittests src/main.rs (target/debug/deps/tailcall-917f44037362571b)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/execution_spec.rs (target/debug/deps/execution_spec-0cb63587319e8dd8)
thread 'main' panicked at tests/execution_spec.rs:974:33:
Failed to parse explicit test path "cli::javascript"
error: test failed, to rerun pass `--test execution_spec`
```
**Expected**
There should be no panics.

ssddOnTop:
## Example use
**Before:** earlier we were able to run insa test for single md file with `cargo test tests/execution/foo.md`
**Now**: now we have proper structure, all insa tests lies in `execution_spec` so we can run test for single file by:
 
`cargo test --test execution_spec test -- i tests/execution/foo.md`  you can replace `i` with `insta`
or
~~`cargo test -- i tests/execution/foo.md`~~ this won't work if we switch from `tokio::main` to `tokio::test` in  execution_spec
etc

also, this gives ability to pass several files at the same time, for eg: `cargo test -- i tests/execution/foo.md foo1.md`

**Issue Reference(s):**  
Fixes #1236
/claim 1236


amitksingh1490:
cargo test cli::javascript should only run cli::javascript however its running execution_spec as well

amitksingh1490:
> Thats’s another issue I think. Its related to insta test.. its using main function which acts as normal main function instead of test. “I think” its possible to convert that into test but then we wont be able to pass args to to test single md file (idk thats needed anyways)

I don't think thats the different issue, It was panicking because it was running execution_spec even if it doesn't supposed to.
I think running single test is already supported by execution_spec already.
https://github.com/tailcallhq/tailcall/blob/main/tests/execution_spec.md#annotation

ssddOnTop:
after the commit above, if we want to run test for specific file then we'll have to pass arg to the `test` function in executio_spec

example:
`cargo test --test execution_spec test -- -i tests/execution/foo.md`

to avoid passing name of `test` function, we need to move content of execution spec under `#[cfg(test)]` but that will lead to change in all snapshots

amitksingh1490:
`cargo test --test execution_spec test -- -i tests/execution/test-const-with-mustache.md`

this command fails with
```
error: Unrecognized option: 'i'
error: test failed, to rerun pass `--test execution_spec
```
however this works
`cargo test --test execution_spec test -- i tests/execution/test-const-with-mustache.md`

ssddOnTop:
> `cargo test --test execution_spec test -- -i tests/execution/test-const-with-mustache.md`
> 
> this command fails with
> 
> ```
> error: Unrecognized option: 'i'
> error: test failed, to rerun pass `--test execution_spec
> ```
> 
> however this works `cargo test --test execution_spec test -- i tests/execution/test-const-with-mustache.md`

yes. I should remove `--` and `-` because it tries to parse it as arg for cargo test

tests/execution_spec.rs
```diff
@@ -959,29 +959,42 @@ async fn assert_spec(spec: ExecutionSpec) {
...
-
-        let contents = fs::read_to_string(&path)?;
-        let spec: ExecutionSpec = ExecutionSpec::from_source(&path, contents)
-            .await
-            .map_err(|err| err.context(path.to_str().unwrap().to_string()))?;
 
-        vec![spec]
-    } else {
+    let args: Vec<String> = std::env::args().collect();
+    println!("args: {:?}", args);
```
amitksingh1490, line 972:
```suggestion
```

ssddOnTop:
> @tusharmath what do you think. I feel `cargo test --test execution_spec -- i tests/execution/foo.md` has better devex than ##### only

I think you misinterpreted something. `#####` is used in md file to perform respective function. But here the argument is used to path of md files which needs to be executed.. this pr has no relation with that.

I checked execution_spec.md, I don't think there was anything that needs to be updated for this pr. However I found that `Porting` section have some outdated info, I'll make a commit to update that

amitksingh1490:
incorrect filename for the snapshot
tests/snapshots/execution_spec__execution_spec__add-field-index-list.md_assert_0.snap.
Thats why all the snapshots are comming as new one

ssddOnTop:
> incorrect filename for the snapshot tests/snapshots/execution_spec__execution_spec__add-field-index-list.md_assert_0.snap. Thats why all the snapshots are comming as new one

I accidentally committed them 😓

tests/execution_spec.md
```diff
@@ -238,7 +238,7 @@ This will regenerate all snapshots without interrupting the test every time ther
 
 ## Porting from `graphql_spec`/`http_spec`
 
-Porting is automatically done by `testconv`. You can run `testconv` by doing `cargo run -p testconv`, but it is automatically performed by `./lint.sh --mode=fix`. This is a description of what it does under the hood.
+It's converted as a normal test. We do not need to anything additional for testcov.
```
amitksingh1490, line 241:
```suggestion
```

tests/execution_spec.md
```diff
@@ -238,7 +238,7 @@ This will regenerate all snapshots without interrupting the test every time ther
 
 ## Porting from `graphql_spec`/`http_spec`
```
amitksingh1490, line 239:
```suggestion
```

Issue #702, opened 2023-11-29
Implement http_spec tests with actual running mock server


meskill:
## The problem

Current http_spec test implementation has some limitations:
- to implement mocks it uses custom HttpClient implementation and the test partially reimplements logic how tailcall server is started. Those implementation could diverge with original implementation and brings burden of supporting both versions of code for actual and test run.
- assertions with config file like yml for specifying tests doesn't allow to test complex cases like caching, wildcard routes, testing passed headers. Also, it's harder to implement testing for custom protocols, e.g. grpc
- the tests don't make actual http requests to the mock upstream that could hide possible issues with app that lie in http layer
 
 ## Solution
 
 Implement mocks for http_spec as separate server (for example, with `mockito` crate)
 
 That should allow to bypass the limitations above:
 - we will test actual http requests to the mock server
 - we won't need to reimplement server handling for tests and can just start tailcall with different configurations
 - will be able to test more complex cases when we want to test batch requests, caching mechanic
 
 ### Acceptance terms
 
 When testing http behaviour:
- we can run mock server
- we can define mocks for server that could match the request ignoring order of parts of the requests (for example ignore query params order to test batching is worked despite the order)
- we can start tailcall server based on test config and that will automatically pointed to the mock server as upstream servers
- we can assert mocks in terms that they were called, what response we've got from tailcall, how many times the mock was called

meskill:
> Can you provide a few examples of what we aren't able to test right now but still fall within the scope of http spec.

I've updated the description with more details and with expected solution

PR #1242, opened 2024-02-23
fix: losing cache wrapper for expr from app_context

meskill:
**Summary:**  
Cache was not working because it was cleaned when updating resolvers from app_context

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/lambda/cache.rs
+++ b/src/lambda/cache.rs
@@ -1,11 +1,12 @@
 use core::future::Future;
 use std::num::NonZeroU64;
+use std::ops::Deref;
 use std::pin::Pin;
 
 use anyhow::Result;
 use async_graphql_value::ConstValue;
 
-use super::{Concurrent, Eval, EvaluationContext, Expression, ResolverContextLike, IO};
+use super::{Concurrent, Eval, EvaluationContext, Expression, ResolverContextLike};
 
 pub trait CacheKey<Ctx> {
     fn cache_key(&self, ctx: &Ctx) -> u64;
@@ -14,7 +15,7 @@ pub trait CacheKey<Ctx> {
 #[derive(Clone, Debug)]
 pub struct Cache {
     pub max_age: NonZeroU64,
-    pub expr: IO,
+    pub expr: Box<Expression>,
 }
 
 impl Cache {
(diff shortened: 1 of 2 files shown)


src/lambda/cache.rs
```diff
@@ -14,7 +15,7 @@ pub trait CacheKey<Ctx> {
 #[derive(Clone, Debug)]
 pub struct Cache {
     pub max_age: NonZeroU64,
-    pub expr: IO,
+    pub expr: Box<Expression>,
```
tusharmath, line 18: We wanted to make sure that cache can only be applied on IOs. Why have we changed this guarantee?

tusharmath:
How come we don't have any failing tests?

src/lambda/cache.rs
meskill: because in `src/lambda/modify.rs` we can pass functions that works only on expression and to call it recursively we need to pass the expression not io to support typings.

tusharmath: If I understand you, are you saying we need to recursively wrap the expression with Cache expression?

I think that might not be required. We could insert the cache operator at the first IO express, and cache just that. Any IO that's happening internally shouldn't need any caching because the top IO expression just resolve from cache.

It's possible that I misunderstand your point here. Feel free to expand more with an example if you think I am wrong.

meskill: no, it's not about that, it's more about type and traits we have.

1. In app_ctx we want to traverse all the field resolvers and update io operations.
2. The code https://github.com/tailcallhq/tailcall/blob/51d8b7aff838f0f4c362d4ee9e39492ae1f51fdb/src/app_context.rs#L37-L39 use `Expression::modify` that accepts callback `Expression -> Expression`
3. In order to update only specific fields we have that modify logic that recursively traverse expression and calls the callback https://github.com/tailcallhq/tailcall/blob/a12599f0fe916835bb04e46c28339f956d06b90b/src/lambda/modify.rs#L41-L43 Note here we return new IO expr but we actually should return Expression::Cache
4. But the `modifier` can actually return any expression (according to types) not only IO and we need somehow to unwrap returned IO from Expression.

Because the modifier returns wrapper `Expression` but cache stores only IO I think the mistake was done in the first place i.e. types are misleading. 
We could solve by having something specific to handle io operations or have in-place updates how I was doing in the https://github.com/tailcallhq/tailcall/pull/1032/files

meskill:
> How come we don't have any failing tests?

Because with `execution_spec` we don't have functionality to verify if actual requests were made or the cache was used. I found out the problem only when I tried to test opentelemetry metrics output and noticing that they are not generated at all for the cache.

That was one of the points in https://github.com/tailcallhq/tailcall/issues/702

PR #1255, opened 2024-02-27
feat(scalar): add email gql scalar

### LINKED ISSUE #1214: Support custom GraphQL scalars
Inspired by https://www.npmjs.com/package/graphql-scalars we could make the server a lot more type-safe by adding custom scalars. This should include type definitions that are auto-generated and validations for each type.
Some example scalars are

- EmailAddress
- PhoneNumber
- Date
- URL

ssddOnTop:
**Summary:**  
Adds Email Scalar

**Issue Reference(s):**  
 #1214

## Summary by CodeRabbit

- **New Features**
	- Introduced a new custom scalar type `Email` for enhanced data validation in GraphQL schemas.
	- Added validation logic for the `Email` scalar type to ensure adherence to standard email formats.
- **Enhancements**
	- Improved scalar handling in GraphQL schema generation by adding custom scalar support.
	- Implemented a new validation function for scalar types to ensure data integrity.
- **Tests**
	- Added test cases for validating the behavior of the new `Email` scalar type with both valid and invalid inputs.


tests/snapshots/execution_spec__add-field-index-list.md_client.snap
```diff
@@ -2,6 +2,8 @@
 source: tests/execution_spec.rs
 expression: client
 ---
+scalar Email
```
amitksingh1490, line 5: this should come only if its used in config.

src/scalars/email.rs
```diff
@@ -0,0 +1,99 @@
...
+mod test {
+    use anyhow::Result;
+    use async_graphql::dynamic;
+    use async_graphql::dynamic::Type::Object;
+    use async_graphql::dynamic::{Field, FieldFuture, InputValue, TypeRef};
+    use async_graphql_value::value;
+
+    use crate::scalars::Email;
+
+    fn get_schema(scalar: dynamic::Scalar, resp: String) -> Result<dynamic::Schema> {
```
amitksingh1490, line 31: add test using execution spec

src/scalars/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+pub struct Scalars {
+    pub email: Email,
+}
+
+pub fn get_scalar(name: &str) -> fn(&ConstValue) -> bool {
+    match name {
+        "Email" | "email" => Email::validate,
+        &_ => |_| true,
+    }
+}
```
tusharmath, line 20: why do we need this?

ssddOnTop: We will add scalars for date phone number etc 

the scalar will be in form of string, so we can just match scalar and get corresponding validation method

src/scalars/email.rs
```diff
@@ -0,0 +1,99 @@
+use async_graphql::validators::email;
+use async_graphql_value::ConstValue;
+
+use crate::json::JsonLike;
+
+#[derive(schemars::JsonSchema)]
+/// A custom scalar to validate the format of email address
```
tusharmath, line 7: Update the description to:
> A field whose value conforms to the standard internet email address format as specified in HTML Spec: https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address.'

generated/.tailcallrc.graphql
```diff
@@ -638,6 +638,7 @@ input Schema {
   Arr: Schema
   Opt: Schema
 }
+scalar Email
```
tusharmath, line 641: Documentation should be generated for `Email`

src/scalars/email.rs
```diff
@@ -0,0 +1,99 @@
...
+        scalar = scalar.validator(Email::validate);
+
+        let schema = get_schema(scalar, "alo@validresp.com".to_string())?;
+
+        let resp = schema.execute("{ value(val: \"alo@invalidvalid\") }").await;
+
+        assert!(resp.data.into_json().unwrap().is_null());
+
+        Ok(())
+    }
```
tusharmath, line 98: I think just testing for `validate` function should be enough in these tests.

tests/snapshots/execution_spec__test-scalar-email-invalid.md_assert_0.snap
```diff
@@ -0,0 +1,27 @@
...
+{
+  "status": 200,
+  "headers": {
+    "content-type": "application/json"
+  },
+  "body": {
+    "data": null,
+    "errors": [
+      {
+        "message": "internal: invalid value for scalar \"Email\", expected \"FieldValue::Value\"",
```
tusharmath, line 14: The first part of the error still makes sense. The latter -  `expected \"FieldValue::Value\"` doesn't seem to make much sense to me.

src/scalars/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+
+mod email;
+
+pub const CUSTOM_SCALARS: &[&str] = &["Email"];
+
+#[derive(schemars::JsonSchema)]
+/// A wrapper to store all custom scalar types
+pub struct Scalars {
+    pub email: Email,
+}
```
tusharmath, line 13: If we are using `Scalars` primarily for packaging all the scalars into one type, then we should perhaps use an enum.

src/scalars/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+
+pub fn get_scalar(name: &str) -> fn(&ConstValue) -> bool {
+    match name {
+        "Email" | "email" => Email::validate,
```
tusharmath, line 17: We can create a trait `Scalar` with a `validate` method on it, and implement it for `Email` type. We could also then update Scalars to hold a `HasMap<String, dyn Scalar>`.

```rust
trait Scalar {
  fn validate(c: &ConstValue) -> bool;
}
```

src/scalars/email.rs
ssddOnTop: async_graphql is providing the email function, which is already tested

I don't think we need tests, but these tests are good for people to get idea about future implementations

tests/snapshots/execution_spec__test-scalar-email-invalid.md_assert_0.snap
ssddOnTop: That error is thrown by async_graphql itself.. I don't think we have control over it

src/scalars/mod.rs
ssddOnTop: this function is called while compiling schema. I don't think we need map here

src/scalars/email.rs
tusharmath: I find it quite confusing to test scalar like this.

src/scalars/mod.rs
tusharmath: **PLEASE DON'T RESOLVE COMMENTS THAT HAVE NOT BE FIXED**

tests/snapshots/execution_spec__test-scalar-email-invalid.md_assert_0.snap
tusharmath: We need to have control over error messages.

src/scalars/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use async_graphql_value::ConstValue;
+use lazy_static::lazy_static;
+
+pub use crate::scalars::email::Email;
+
+mod email;
+
+lazy_static! {
+    pub static ref CUSTOM_SCALARS: HashMap<String, Arc<dyn Scalar + Send + Sync>> = {
```
tusharmath, line 12:
```suggestion
    static ref CUSTOM_SCALARS: HashMap<String, Arc<dyn Scalar + Send + Sync>> = {
```

ssddOnTop: I am using this in `is_scalar` function

```rust
pub fn is_scalar(type_name: &str) -> bool {
    ["String", "Int", "Float", "Boolean", "ID", "JSON"].contains(&type_name)
        || CUSTOM_SCALARS.keys().any(|v| (*v).eq(type_name))
}
```
it would be tedious to maintain several lists of scalars

tusharmath (approved):

PR #1450, opened 2024-03-15
feat(cache): add support for IO caching in RequestContext

### LINKED ISSUE #1445: feat: add support for IO Cache via RequestContext
Currently caching is only supported when we enable it via the `@cache` operator. By default, an IO within the context of one request, should only be made once. So we need to keep a Cache inside of the request context, and try to access it first before accessing the runtime cache. So the flow should look like the following:

1. Check **Request Context Cache**, if not found
2. Check Runtime Cache (if enabled via `@cache`), if not found
3. Perform the actual IO and update Request Cache and Runtime Cache.

**NOTE:** Request cache will be of unlimited size and is always enabled.

**Why is the cache size unlimited?**
The reason why we have unlimited size in request context cache is - That within a GraphQL Request, there are finite number of IOs that one needs to perform, 100s at most. So we don't need a limit per say, because the number of IOs cached will never go beyond a few 100s. Moreover the cache is destroyed, as soon as the response is sent back to the client.

**How is the cache key identified?**
We already use CacheKey for runtime cache. Leverage the same semantics and APIs to create the cache within the request context.

shashitnak:
**Summary:**  
Eliminates duplicate IO calls made in the same request by saving the results in `RequestContext`

**Issue Reference(s):**  
Fixes #1445 

**Build & Testing:**

**Checklist:**

## Summary by CodeRabbit

- **New Features**
	- Introduced a new caching mechanism across the application to improve performance.
	- Added an `AsyncCache` structure for asynchronous value retrieval and caching.

- **Enhancements**
	- Expanded `RequestContext` to include caching capabilities using a combination of `HashMap` and `broadcast` channels.
	- Enhanced `IO` operations to support caching, improving efficiency in data handling.

- **Refactor**
	- Reorganized module imports for better code structure and readability.

- **Documentation**
	- Updated public exports to include `CacheValue`, making it accessible for further development and integration.

### DIFF
--- a/benches/impl_path_string_for_evaluation_context.rs
+++ b/benches/impl_path_string_for_evaluation_context.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
-use std::collections::BTreeMap;
-use std::sync::{Arc, Mutex};
+use std::collections::{BTreeMap, HashMap};
+use std::sync::{Arc, Mutex, RwLock};
 use std::time::Duration;
 
 use async_graphql::context::SelectionField;
@@ -258,6 +258,7 @@ fn request_context() -> RequestContext {
         min_max_age: Arc::new(Mutex::new(None)),
         cache_public: Arc::new(Mutex::new(None)),
         runtime,
+        cache: Arc::new(RwLock::new(HashMap::new())),
     }
 }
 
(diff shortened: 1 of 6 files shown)


tusharmath:
closing in favour of https://github.com/tailcallhq/tailcall/pull/1468

PR #1462, opened 2024-03-16
test: running execution spec parallelly

mehulmathur16:
**Summary:**  
/claim #1461 

**Issue Reference(s):**  
Fixes #... _(Replace "..." with the issue number)_

**Build & Testing:**

- [ ] I ran `cargo test` successfully.
- [ ] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [ ] I have added relevant unit & integration tests.
- [ ] I have updated the [documentation] accordingly.
- [ ] I have performed a self-review of my code.
- [ ] PR follows the naming convention of `<type>(<optional scope>): <title>`

[documentation]: https://github.com/tailcallhq/tailcallhq.github.io/tree/develop/docs

## Summary by CodeRabbit

- **New Features**
	- Introduced parallel async task execution for asserting specifications.
	- Enhanced error handling for async tasks.
	- Aggregated and displayed errors for better debugging.

PR #1464, opened 2024-03-16
test: run execution spec parallel

### LINKED ISSUE #1461: execution spec bails out on the first failure
Currently execution_spec runs each test one by one and fails on the first error. The expectation is to run the tests in parallel and show all the failures at once.

Bhavyajain21:
/claim #1461 
/fixes #1461 

**Implementation includes :**

- optimal process of reading multiple files concurrently using Tokio's asynchronous task handling
- running the tests in parallel and show all the failures at once.

## Summary by CodeRabbit

- **Tests**
	- Improved test efficiency and error reporting by introducing parallel task execution.


Bhavyajain21:
Please review @tusharmath

tusharmath:
@Bhavyajain21 This is a duplicate of https://github.com/tailcallhq/tailcall/pull/1462
by @mehulmathur16 
One of you should close your PR and work on something else.

Bhavyajain21:
This includes optimized process of reading multiple files concurrently using Tokio's asynchronous task handling.

tusharmath:
What's the improvement in execution time when compared to main?

Bhavyajain21:
> What's the improvement in execution time when compared to main?

@tusharmath Asynchronous handling using Tokio allows for overlapping I/O operations, which can result in significant performance improvements when dealing with I/O-bound tasks, such as reading files from disk. By initiating multiple file read operations concurrently and efficiently handling the results as they become available, asynchronous tasks can make better use of system resources and reduce overall execution time.

tusharmath:
Can you share exact numbers.

Bhavyajain21:
@tusharmath To provide exact numbers on efficiency improvement, I plan to measure the execution time of the test suite both sequentially and with the parallel implementation. I'll use Instant::now() from std::time to capture the start and end times.

I'll update the PR with the exact numbers and performance improvement percentage once I've completed the measurements.

Bhavyajain21:
@tusharmath Here's the screenshot attached with exact numbers and performance improvement percentage after completing the measurements.

tusharmath:
That's pretty low.

Bhavyajain21:
> That's pretty low.

And this is the one after running it with my logic of code to optimize reading multiple files concurrently using Tokio's asynchronous task handling. It is increased to 10%

Screenshot 2024-03-17 at 5 18 47 PM

tusharmath:
Using multi threading we should have achived order of magnitude of increase in performance. It seems like we are still running the tests sequentially.

Bhavyajain21:
> Using multi threading we should have achived order of magnitude of increase in performance. It seems like we are still running the tests sequentially.

@tusharmath You're right, multithreading should ideally lead to a much bigger performance improvement than the current 11%

Though I did a search about it and got to know that there could be a few of these reasons for this:

Shared Resources: The tests might be waiting on shared resources like databases or files, creating bottlenecks.

Thread Overhead: Managing threads can introduce overhead that negates the benefits for short-running tests.

Test Design: The tests might not be properly divided for parallel execution.


Using the code in the other PR, it is coming 9% as well.

tusharmath:
This approach requires a lot more code to maintain. Opened up a new issue to track it https://github.com/tailcallhq/tailcall/issues/1500

/tip 20

Bhavyajain21:
Thanks for the tip! @tusharmath

PR #1460, opened 2024-03-16
fix: windows CI

ssddOnTop:
**Summary:**  
_Briefly describe the changes made in this PR._

**Issue Reference(s):**  
Fixes #1280
/claim 1280

**Build & Testing:**

- [ ] I ran `cargo test` successfully.
- [ ] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [ ] I have added relevant unit & integration tests.
- [ ] I have updated the [documentation] accordingly.
- [ ] I have performed a self-review of my code.
- [ ] PR follows the naming convention of `<type>(<optional scope>): <title>`

[documentation]: https://github.com/tailcallhq/tailcallhq.github.io/tree/develop/docs

## Summary by CodeRabbit

- **New Features**
	- Added support for distinguishing between HTTP URLs and file paths in configuration reading.
	- Introduced package declarations in various protocol buffer files to improve import management.

- **Bug Fixes**
	- Fixed content comparison in execution specification tests to properly handle different line endings.
	- Adjusted error handling in `unknown_file` test to check for the occurrence of an error without comparing the message.

- **Refactor**
	- Refactored error assertions and path resolutions in configuration reader tests.
	- Enhanced test setup by adjusting directory paths and comparison checks for package equality.

- **Tests**
	- Modified tests to support changes in error handling and package declarations.

- **Chores**
	- Updated CI workflow configurations to optimize build targets across Linux, macOS, and Windows platforms.

- **Documentation**
	- Noted the limitation regarding nested `@graphql` directives in the GraphQL data loader documentation.

PR #1472, opened 2024-03-16
fix/tests on windows

ologbonowiwi:
- **build: enable win32 build/testing**
- **test: ensure consistency between spec printing across platforms**


tusharmath:
this has been fixed in #1460

Issue #1475, opened 2024-03-16
ability to add multiple IOs in a config


tusharmath:
Currently `config::Field` can support only one resolver at a time. These resolvers can either be a GRPC, HTTP GraphQL or a Const. This proposal intends to change it by providing a vector of IO operations that can be piped into one another.

**Change in `config.rs`:**
```diff
  pub struct Field {            
      pub args: BTreeMap<String, Arg>,            
      pub cache: Option<Cache>,
      pub doc: Option<String>,            
      pub expr: Option<Expr>,        
      pub graphql: Option<GraphQL>,            
      pub list_type_required: bool,            
      pub list: bool,            
      pub modify: Option<Modify>,            
      pub omit: Option<Omit>,            
      pub required: bool,            
      pub script: Option<JS>,            
      pub type_of: String,            
-     pub call: Option<Call>,            
-     pub const_field: Option<Const>,            
-     pub grpc: Option<Grpc>,            
-     pub http: Option<Http>,            
+     pub resolver: Vec<Resolver>,
  }

+ pub enum Resolver {
+   Call(Call),
+   Const(Const),
+   Grpc(Grpc),
+   Http(Http),
+ }
```

**Change in Blueprint**
- The structure for blueprint remains the same with each resolver compiling into an Expression. However incase of multiple resolvers the expressions should be composed with each other using the `and_then` operation on Expression.

**Change in `from_document.rs`**
The following is a totally valid configuration:
```graphql
type Query {
  abc: ABC 
    @http(path: "/abc")  # assuming it returns {a: {b: {c: "abc"}}}
    @const(data: "{{value.a}}") 
    @const(data: "{{value.b}}") 
    @const(data: "{{value.c}}")
}
```
The `value` which is accessible via context contains the output of the previous resolver. This is already implemented by `and_then`

**Technical Requirements**
- Make sure there are enough integration tests.

tusharmath:
Will hold on to this for a bit. There is an alternative way to do the same thing.

ssddOnTop:
We are working on alternate method. Closing the issue

PR #1479, opened 2024-03-17
feat: add expr path

ssddOnTop:
## Summary by CodeRabbit

Related to: #1475 

## Summary by CodeRabbit

- **New Features**
	- Enhanced data handling capabilities by introducing a new `path` field for more complex expressions.
- **Refactor**
	- Expanded the logic for expression and object handling to support additional data structures and functionalities.

### DIFF
--- a/src/lambda/expression.rs
+++ b/src/lambda/expression.rs
@@ -2,7 +2,7 @@ use core::future::Future;
 use std::fmt::{Debug, Display};
 use std::pin::Pin;
 
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use async_graphql_value::ConstValue;
 use thiserror::Error;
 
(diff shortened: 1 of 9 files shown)


src/lambda/expression.rs
```diff
@@ -157,6 +160,16 @@ impl Eval for Expression {
...
                 Expression::List(list) => list.eval(ctx, conc).await,
                 Expression::Math(math) => math.eval(ctx, conc).await,
+                Expression::Object(object) => {
+                    match object {
+                        Object::Path(path) => Ok(ctx
+                            .value()
+                            .ok_or(anyhow!("Expected a value for path: {}", object))?
+                            .get_path(path)
+                            .cloned()
+                            .ok_or(anyhow!("No such value exists"))?),
```
tusharmath, line 170: Don't need to throw an error. We should just return null value if nothing is available.

src/lambda/object.rs
```diff
@@ -0,0 +1,13 @@
+use std::fmt::{Display, Formatter};
+#[derive(Clone, Debug)]
```
tusharmath, line 2:
```suggestion
#[derive(Clone, Debug, strum_macros::Display)]
```

tusharmath:
Need some tests for expr with path.

Issue #1153, opened 2024-02-08
Restructure Workspace


tusharmath:
High-Level Structure of workspace

```bash
/core           #publish on Crates
/core-runtime   #publish on Crates
/integration    #integration tests
/main-cli       #cli application
/main-aws-lambda
/main-cloudfare
/main-fastly
/main-web
/main-node

/main-azure    #personal project
```

- [ ] Create a `TailcallBuilder` in `src/builder.rs` (#1195)
- [ ] Migrate unit tests to `TailcallBuilder`
- [x] Separate Test IO (#1165)
- [ ] Make all `src/lib.rs` type crate private.
- [ ] Refactor everywhere to use `TailcallBuilder`
- [x] Drop `dyn HttpIO` and other IO from types and refer them to use TargetRuntime instead. (#1155)

PR #1573, opened 2024-03-27
refactor: move fmt to src

ssddOnTop:
moves cli/fmt to src
part of: #1153


tusharmath:
fmt is only used by CLI. Not sure why we should move it to src?

ssddOnTop:
I thought I am still assigned for the issue, just saw that kiryl picked it up so closing this pr

PR #1610, opened 2024-03-30
fix(generator): add map support

ssddOnTop:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new `GetPerson` functionality in the `PersonService`.

- **Improvements**
	- Enhanced processing logic to efficiently handle specific message types and field attributes.
	- Updated the logic to handle field types, setting `list` to false for type "JSON".

- **Refactor**
	- Adjusted type conversion logic for better data handling and representation.
	- Added a mapping for type "message" to "JSON" in the `convert_ty` function.


src/config_generator/proto/person.proto
```diff
@@ -14,4 +17,8 @@ message Person {
   required int32 id = 2;
   optional string email = 3;
   repeated PhoneNumber phone = 4;
+  map<string, string> map = 5;
```
tusharmath, line 20:
```suggestion
  map<string, string> stringMap = 5;
  map<string, Address> addressMap = 6;
```
And more types ...
Add something where there key and values are complex types.

src/config_generator/from_proto.rs
```diff
@@ -330,6 +336,8 @@ fn convert_ty(proto_ty: &str) -> String {
         "int32" | "int64" | "fixed32" | "fixed64" | "uint32" | "uint64" => "Int",
         "bool" => "Boolean",
         "string" | "bytes" => "String",
+        "message" => "JSON", /* JSON scalar is preloaded by tailcall, so there is no need to
+                               * explicitly define it in the config. */
```
tusharmath, line 340: Use `//` to write comments in code. We do that everywhere else.

ssddOnTop: it was converted to /* by cargo or prettier

tusharmath (approved):

PR #1634, opened 2024-04-02
fix: handle Aliases for enum

### LINKED ISSUE #1626: bug(grpc): Enums with aliases don't work in GRPC
**Proto File**
```proto
syntax = "proto3";

package greetings;

service Greeter {
  rpc SayHello (HelloRequest) returns (HelloReply) {}
}

message HelloRequest {
  string name = 1;
}

message HelloReply {
  string message = 1;
  repeated OBJECT_TYPE objects = 2;
}

enum OBJECT_TYPE {
  option allow_alias   = true;
  ONE = 0;
  TWO = 1;
  O_NE = 0;
  T_WO = 1;
}
```

**Configuration**
```graphql
schema
  @upstream(baseURL: "...", httpCache: true, batch: {delay: 10})
  @link(src: "./enums.proto", type: Protobuf) {
  query: Query
}

type Query {
  hello(name: JSON): HelloReply!
    @grpc(
      method: "greetings.Greeter.SayHello"
      body: "{{args.name}}"
    )
}

type HelloReply {
  message: String
  objects: [ObjectType]!
}

enum ObjectType {
  ONE
  TWO
}
```

**Actual**
```
ERROR Invalid Configuration
Caused by:
  • expected Enum type [at Query.hello.@grpc.HelloReply.objects]
```

**Expected**
Should start the server without errors and successfully query the GRPC endpoint.

webbdays:
**Summary:**  
Changes in json_schema module and other places to handle Aliases for enum

**Issue Reference(s):**  
may be Fixes #1626
/claim #1626

**Build & Testing:**

- [ 1 test failing ] I ran `cargo test` successfully.
Its due to ordering of enum variants/values. for that test order is not maintained as given in config file.
They are being sorted out. Hence test failing according to my initial review.

- [Yes] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [Yes] I have performed a self-review of my code.
- [Yes] PR follows the naming convention of `<type>(<optional scope>): <title>`

## Summary by CodeRabbit

- **New Features**
	- Enhanced the GraphQL schema to support `Enum` fields as `JSON` types, allowing for more complex and structured enum definitions.
- **Refactor**
	- Updated internal handling of enum variants to improve organization and efficiency, transitioning from `BTreeMap` and `BTreeSet` to `HashMap` and `HashSet`.


src/json/json_schema.rs
```diff
@@ -136,8 +136,22 @@ impl JsonSchema {
...
+                        return Valid::fail("expected proper Enum type".to_string()).trace(name);
+                    } else {
+                        let keys: HashSet<_> = a.keys().collect();
+                        for key in keys {
+                            let mut small = &b[key];
+                            let mut large = &a[key];
+                            if a[key].len() < b[key].len() {
+                                small = &a[key];
+                                large = &b[key];
+                            }
```
ssddOnTop, line 149: I don't think it's good idea to hardcode it this way.. we should check if alias is really allowed

webbdays: I think allow alias option is just for reading proto files with aliases.
If you read proto file with aliases when not allowing them, then we are getting validation error.
If alias is allowed then we can have variants with name number for the example proto provided in the issue.

webbdays: > I don't think it's good idea to concat array it this way.. we should check if alias is really allowed.

Where is concatenation?

webbdays: Actually, we should not do small and large this way, if we are certain that a is from schema file and b is from proto file.

ssddOnTop: > I think allow alias option is just for reading proto files with aliases. If you read proto file with aliases when not allowing them, then we are getting validation error. If alias is allowed then we can have variants with name number for the example proto provided in the issue. We use them this way to validate enum.

you are putting index as key which is incorrect. Yes you can keep value from proto as key but that’s not what it’s doing currently

webbdays: I am using value.number() as key for proto
And index of variants as key for enum from graphql schema. Which I am treating it in order defined by the user. If this is why you are talking about? Please clarify.
User can specify the number for each enum variant explicitly in graphql. Otherwise implicitly we do as above.
I think this clarifies.

webbdays: I am using value.number() as key for proto
And index of variants as key for enum from graphql schema. Which I am treating it in order defined by the user. If this is why you are talking about? Please clarify.
User can specify the number for each enum variant explicitly in graphql. Otherwise implicitly we do as above.
Let know if I am understanding it wrong.

src/blueprint/from_config.rs
```diff
@@ -89,7 +89,11 @@ where
     let schema = match type_ {
         Some(type_) => {
             if let Some(variants) = type_.variants.clone() {
-                JsonSchema::Enum(variants)
+                let mut map: HashMap<i32, HashSet<String>> = HashMap::new();
+                for (i, variant) in variants.into_iter().enumerate() {
+                    map.entry(i as i32).or_default().insert(variant);
+                }
+                JsonSchema::Enum(map)
```
ssddOnTop, line 96: do we need new map? Won't `JsonSchema::Enum(variants)` work fine?

src/json/json_schema.rs
ssddOnTop: I am confused on how are you planning to check if alias is turned on? If it's not explicitly defined, we should throw an error

src/blueprint/from_config.rs
webbdays: The change.in json_schema module wants the argument as hashmap.

Because I am using hashmap for validation of enum type in compare method of json_schema module there.

ssddOnTop: for that you might need to change type of `type_.variants` to HashMap as well because it's not guaranteed that the value number lies in the range of number of variants.  

example:

```
enum Foo {
 BAR = 0;
 BAR1 = 59;
}
```
Is perfectly allowed. But with current implementation I think it will never look for number 59.

src/json/json_schema.rs
webbdays: I am not checking it explicitly.
If there are aliases, to read the proto correctly without validation error one should define option there in enum to allow alias.

I am not sure where that validation is performed in code. But I have tested an example for it.

If no alias then it will as usual pass through.
In this case the value: hashset in hashmap is of size one.
In this case Is Subset validation is between hashsets of same same size.
Otherwise if there are alias it's unequal size.

We should loop over keys of small hashset ie. For graphql schema enum.

webbdays: In will correct it now.

ssddOnTop: you need to check for `allow_alias`

src/blueprint/from_config.rs
webbdays: Yes. Nice catch.
That would be the problem in that case. We need to find some way to capture those as well when reading.
not just variants values.

ssddOnTop: change `type_.variants` to the HashMap instead of BTreeSet

webbdays: ok.

webbdays: need to changes some code to solve
self.variants = Some(variants.union(other).cloned().collect());

webbdays: variants is now  a hasmap right.

webbdays: Like if there is conflicts like same keys when merging hashmap then preferring the right side one.?

tusharmath:
Let's add a test like described here https://github.com/tailcallhq/tailcall/issues/1626

Moving to draft until we have a failing test. Feel free to connect on discord, if something isn't clear.

src/json/json_schema.rs
ssddOnTop: The code must be strict. We need to throw an error if alias is false/undefined.

example: imagine there are 3 fields defined in an enum in proto file but the user forgot and defined just 3 fields in graphql config, in that case we must throw an error else the user will just get "2" as response instead of actual enum field name, which is quite incorrect.

webbdays: I think the alias property in proto files is handled by proto reader.

webbdays: let me try with an example.

ssddOnTop: > I think the alias property in proto files is handled by proto reader.

it's not handled anywhere, we are dropping all that information as soon as we read proto file 

check my comment: https://github.com/tailcallhq/tailcall/issues/1626#issuecomment-2032455118

we do not preserve any kind of options in our descriptor, which needs a fix.

webbdays: sorry not by proto reader, but someone.
not handling allow_alias property explicitly.
but its is checking like this:
ERROR Invalid Configuration
Caused by:
  • enum number '0' has already been used [at Query.hello.@grpc]

ssddOnTop: yes, we need to check if allow_alias is set to true. If it's true then we should not throw the error else we must.

webbdays: checking initially at proto reading stage is good.
But i didnt understand why this error wont come if set to true.
ERROR Invalid Configuration
Caused by:
• enum number '0' has already been used [at Query.hello.https://github.com/grpc]

PR #1649, opened 2024-04-03
refactor: enhance config types reading performance

### LINKED ISSUE #1642: refactor: Enhance ConfigModule
Config is designed to be optimized for writes and not for reads. This makes it expensive to query and find stuff in the config. Let's leverage ConfigModule for it.

**Technical Requirements**
- Implement optimized versions for the following functions in config_module such that they have typically a `O(1)` complexity to lookup.
  - [ ] `output_types`
  - [ ] `input_types`
  - [ ] `find_type`
  - [ ] `find_union`
  - [ ] `contains`
- Drop these methods from `Config`.

ologbonowiwi:
**Summary:**  
- [x] output_types – fixed by computing this at `From<Config> for ConfigModule` and holding the value into the struct
- [x] input_types – fixed by computing this at `From<Config> for ConfigModule` and holding the value into the struct
- [x] find_type - fixed by using `HashMap`, which has an `O(1)` time complexity for most cases
- [x] find_union - fixed by using `HashMap`, which has an `O(1)` time complexity for most cases
- [x] contains - fixed by using `HashMap`, which has an `O(1)` time complexity for most cases

I changed the `types` and `unions` properties from `BTreeMap` to a `HashMap,` which has the desired `O(1)` time complexity for most cases.
In places where we need some ordering (e.g., snapshot generation), we collect the types iterator into a `BTreeMap` before iterating over it.

**Issue Reference(s):**  
Fixes #1642
/claim #1642

/split @atamakahere-git 25

**Build & Testing:**

**Checklist:**

## Summary by CodeRabbit

- **Refactor**
	- Transitioned from `BTreeMap` to `HashMap` for certain internal configurations to optimize performance.
	- Enhanced type processing logic with the introduction of sorted type handling.
- **New Features**
	- Implemented computed values for `input_types` and `output_types` within configurations, enhancing the system's ability to process and differentiate between types more efficiently.
- **Tests**
	- Adjusted test cases to align with the new internal data structure changes.


src/blueprint/definitions.rs
```diff
@@ -522,35 +522,43 @@ pub fn to_definitions<'a>() -> TryFold<'a, ConfigModule, Vec<Definition>, String
...
-            let dbl_usage = input_types.contains(name) && output_types.contains(name);
-            if let Some(variants) = &type_.variants {
-                if !variants.is_empty() {
-                    to_enum_type_definition(name, type_, variants).trace(name)
+        Valid::from_iter(
+            config_module
+                .types
+                .iter()
+                .collect::<BTreeMap<_, _>>()
+                .iter(),
```
ologbonowiwi, line 530: This was necessary as `HashMap` is more optimized for reads but is unordered, while `BTreeMap` has ordering.

Instead of manually ordering the `HashMap` to be used, I'm collecting into the ordered map.

src/config/config.rs
```diff
@@ -158,15 +157,6 @@ impl Config {
...
 
-    pub fn types(mut self, types: Vec<(&str, Type)>) -> Self {
-        let mut graphql_types = BTreeMap::new();
-        for (name, type_) in types {
-            graphql_types.insert(name.to_string(), type_);
-        }
-        self.types = graphql_types;
-        self
-    }
-
```
ologbonowiwi, line 169: As receiving a `Vec` and creating a map is inefficient, I ripped this off and now we're using the version from `Setters` derive.

tusharmath: We should optimize the reads for performance and writes for ergonomics. Vec is quite easy to work with in Rust, hence the API looks the way it does.

src/blueprint/definitions.rs
```diff
@@ -520,10 +520,12 @@ pub fn to_field_definition(
 
 pub fn to_definitions<'a>() -> TryFold<'a, ConfigModule, Vec<Definition>, String> {
     TryFold::<ConfigModule, Vec<Definition>, String>::new(|config_module, _| {
-        let output_types = config_module.output_types();
-        let input_types = config_module.input_types();
-        Valid::from_iter(config_module.types.iter(), |(name, type_)| {
-            let dbl_usage = input_types.contains(name) && output_types.contains(name);
+        let output_types = &config_module.output_types;
+        let input_types = &config_module.input_types;
+        let sorted_types = config_module.types.iter().collect::<BTreeMap<_, _>>();
```
tusharmath, line 525: Not sure if we need this. Can you elaborate on why we need this change?

tusharmath (changes requested):
Overall looks fine to me. The changes to set the types using a hashmap is not ergonomic.

src/config/config.rs
```diff
@@ -45,13 +45,12 @@ pub struct Config {
...
     #[serde(default)]
-    #[setters(skip)]
-    pub types: BTreeMap<String, Type>,
+    pub types: HashMap<String, Type>,
 
     ///
     /// A map of all the union types in the schema.
     #[serde(default, skip_serializing_if = "is_default")]
-    pub unions: BTreeMap<String, Union>,
+    pub unions: HashMap<String, Union>,
```
tusharmath, line 53: HashMap doesn't guarantee order. That makes it hard to assert on tests.

src/config/config.rs
ologbonowiwi: The setter expecting a Vec it was used only for tests. Should I re-add it?

src/config/config.rs
ologbonowiwi: BTreeMap is slower, has average time of O(log n) while HashMap is O(1). Should we "give up" this performance boost and revert to BTreeMap?

Note that on parts that depend on ordering the HashMap is being collected into a BTreeMap, so we don't have issues with the tests in the current implementation. But I can for sure rollback this change if you prefer

atamakahere-git: A simple function can help comparing two hashmaps

```rs

map1.len() == map2.len() && map1.keys().all(|k| map2.contains_key(k))

```

If test cases are the only place to check for equality, it won't be a big of an issue probably.

src/config/config.rs
tusharmath: Yes please.

ologbonowiwi: Done @tusharmath

tusharmath (approved):

PR #1659, opened 2024-04-04
fix: make maxSize optional

### LINKED ISSUE #1655: bug(rc): fix `.tailcallrc.graphql`
The type definition is incorrect for `apollo` in `@telemetry` and `batch.maxSize` in `@server`.

The fields are supposed to be optional. Currently they are all required.

neo773:
**Issue Reference(s):**  
Fixes #1655

**Build & Testing:**

/claim #1655

## Summary by CodeRabbit

- **New Features**
	- Enhanced flexibility in batch processing by making the `maxSize` field optional and supporting `null` values.
- **Refactor**
	- Updated data loading mechanisms to handle optional batch sizes more gracefully across different protocols (GraphQL, gRPC, HTTP).
	- Adjusted handling of optional values in various structs to improve batch processing conditions and default value assignments.
	- Updated method calls to handle optional values using `unwrap_or_default()` for more robust batch size handling.


tusharmath (changes requested):
We need to also fix the apollo telemetry types.

neo773:
@tusharmath 
Pushed! I initially overlooked it as the type itself was optional made the fields that had default value optional.

src/config/apollo.rs
```diff
@@ -15,27 +15,27 @@ pub struct Apollo {
     ///
     /// Setting `userVersion` for Apollo.
     #[serde(default = "default_user_version")]
-    pub user_version: String,
+    pub user_version: Option<String>,
```
tusharmath, line 18: Serde default is already set. We can't have both serde default and `Option`, which defaults to `None`.

PR #1671, opened 2024-04-07
feat: Linter for schema

### LINKED ISSUE #1322: Validate schema using linters
Tailcall should ensure consistency in building the final GraphQL schema. This is the standard it should follow:

**Case styles**
- Field names should use `camelCase`. 
- Type names should use `PascalCase`.
- Enum names should use `PascalCase`.
- Enum values should use `ALL_CAPS`, because they are similar to constants.

There are two approaches to solve this problem:
1. When the setting is enabled, we can throw a validation error that the developer see and fix using the `@modify` operator.
2. We can automatically fix it for the user.

**Proposal**
```graphql
schema @server(lint: {field: true, type: true, enum: true, enumValue: true, autoFix: false}) {
  query: Query
}
```

**Technical Requirements**
Use the `ConfigOptimizer` API defined over here https://github.com/tailcallhq/tailcall/issues/1653

webbdays:
**Summary:**  
Linter

**Issue Reference(s):**  
may be Fixes #1322

**Build & Testing:**

- [Yes] I ran `cargo test` successfully.
- [fixed] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [yes I have performed a self-review of my code.
- [yes] PR follows the naming convention of `<type>(<optional scope>): <title>`

### DIFF
--- a/examples/lint-auto-fix.graphql
+++ b/examples/lint-auto-fix.graphql
@@ -1,0 +1,38 @@
+schema
+  @server(
+    lint: {type: Pascal, enum: Pascal, enumValue: ScreamingSnake, field: Camel, autoFix: true}
+    port: 8000
+    graphiql: true
+    hostname: "0.0.0.0"
+  )
+  @upstream(baseURL: "http://jsonplaceholder.typicode.com", httpCache: true, batch: {delay: 100}) {
+  query: Query
+}
+
+enum postType {
+  blog_post
+  ARTICLE
+}
+
+type Query {
+  posts: [Post] @http(path: "/posts")
+  users: [user] @http(path: "/users")
+  user(id: Int!): user @http(path: "/users/{{args.id}}")
+}
+
+type user {
+  id: Int!
+  name: String!
+  Username: String!
+  email: String!
+  phone: String
+  website: String
+}
+
+type Post {
+  id: Int!
+  userId: Int!
+  title: String!
+  body: String!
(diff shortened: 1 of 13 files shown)


tusharmath:
Moving it to draft, until we have tests to prove that this works. Current implementation doesn't seem to be working. Feel free to change it back once you have tests and it's ready for review.

webbdays:
ok.

webbdays:
I don't know why the tracing arguments are showing not covered. weird.
@tusharmath
Any idea of this.

webbdays:
Written tests working fine.

webbdays:
Still
Have few problems that need to be taken care.
Getting some validation errors when we want to lint field to Pascal, type to camel like these.
But working fine with
field : Camel, type: Pascal

Another thing is still need work to change fields in args. {{value.id}}.

still these need to be addressed.

webbdays:
Hello and hi @coderabbitai,
any idea why code coverage is not shown/covered for this tracing.

webbdays:
action taken 🤒

webbdays:
@meskill 
i have a problem in this.
Its very difficult to aauto lint.

meskill:
> @meskill i have a problem in this. Its very difficult to auto lint using config struct data. Can we opt for a cli solution which lint even before the schema is read by the config reader? i think its best.

could you elaborate more on your suggestion how to solve it in terms how it will work and how it'll solve issues you're facing?

Issue #1665, opened 2024-04-05
bug:  Error message on Authentication Faliure


amitksingh1490:
The error message for not Providing Header and Invalid(random string) Header is giving the same error like given below

```
{
  "data": null,
  "errors": [
    {
      "message": "Authentication Failure: Parameters not provided in the request.",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ]
    }
  ]
}
```

Expected behaviour: 
1. Header not provided
Authentication Failure: Missing Authorization Header
2. Invalid Value Provided
Authentication Failure: Invalid Authorization Header

webbdays:
Can i know, which feature or component this relates to in codebase.

webbdays:
is this:
https://github.com/tailcallhq/tailcall/blob/5770bb5e8b752e917ecc3fddcd03e38c598017d0/src/auth/jwt/jwt_verify.rs#L41
https://github.com/tailcallhq/tailcall/blob/5770bb5e8b752e917ecc3fddcd03e38c598017d0/src/auth/jwt/jwt_verify.rs#L70

amitksingh1490:
I think If we provide link both the verification Basic and jwt and provide a basic auth token, jwt is giving Error.Missing. Ideally it should give the error from basic auth.
Also errors can be fixed in 
errors.rs file inside Auth folder

amitksingh1490:
problem seems to be there in `basic.rs`
```rust
async fn verify(&self, req_ctx: &RequestContext) -> Verification {
    let header = req_ctx.allowed_headers.typed_get::<Authorization<Basic>>();
    println!("{:?}", header);
    let Some(header) = header else {
        return Verification::fail(Error::Missing);
    };

    if self.verifier.check(header.username(), header.password()) {
        Verification::succeed()
    } else {
        Verification::fail(Error::Invalid)
    }
}
```
Header is coming as None even If its invalid

webbdays:
@amitksingh1490 
seems you are testing the query on browser via graphiql UI.
in that case we wont be sending authorisation token along with request.
So you are getting headers None.

webbdays:
I have tested with
token
Getting header fine.
printed:
Header: Some(Authorization(Basic { decoded: "testuser1:$apr1$e3dp9qh2$fFIfHU9bilvVZBl8TxKzL/", colon_pos: 9 }))

webbdays:
we can pass http headers via ui as well.

webbdays:
[...]

PR #1672, opened 2024-04-07
fix: auth header error msg 

mehulmathur001:
**Summary:**  
/claim #1665

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/auth/jwt/jwt_verify.rs
+++ b/src/auth/jwt/jwt_verify.rs
@@ -38,10 +38,12 @@ impl JwtVerifier {
         }
     }
 
-    fn resolve_token(&self, request: &RequestContext) -> Option<String> {
-        let value = request.allowed_headers.typed_get::<Authorization<Bearer>>();
+    fn resolve_token(&self, request: &RequestContext) -> anyhow::Result<Option<String>> {
+        let value = request
+            .allowed_headers
+            .typed_try_get::<Authorization<Bearer>>()?;
 
-        value.map(|token| token.token().to_owned())
+        Ok(value.map(|token| token.token().to_owned()))
     }
 
     async fn validate_token(&self, token: &str) -> Verification {
@@ -65,7 +67,9 @@ impl JwtVerifier {
 impl Verify for JwtVerifier {
     async fn verify(&self, request: &RequestContext) -> Verification {
         let token = self.resolve_token(request);
-
+        let Ok(token) = token else {
+            return Verification::fail(Error::Invalid);
+        };
         let Some(token) = token else {
             return Verification::fail(Error::Missing);
         };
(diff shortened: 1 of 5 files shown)


mehulmathur001:
@tusharmath @amitksingh1490 can you please review this PR ?

amitksingh1490:
error messages are correctly fixed but the actual problem is in basic.rs
```async fn verify(&self, req_ctx: &RequestContext) -> Verification {
    // if the decode fails below line returns NONE. Which should not happen as the headers were passed but actually they were invalid
    let header = req_ctx.allowed_headers.typed_get::<Authorization<Basic>>();
    println!("{:?}", header);
    let Some(header) = header else {
        return Verification::fail(Error::Missing);
    };

    if self.verifier.check(header.username(), header.password()) {
        Verification::succeed()
    } else {
        Verification::fail(Error::Invalid)
    }
}
```

mehulmathur001:
@amitksingh1490 made some changes. Please review.

src/auth/basic.rs
```diff
@@ -18,14 +18,16 @@ impl Verify for BasicVerifier {
...
+            if self.verifier.check(header.username(), header.password()) {
+                Verification::succeed()
+            } else {
+                Verification::fail(Error::Invalid)
+            }
+        } else if req_ctx.allowed_headers.contains_key("authorization") {
+            return Verification::fail(Error::Invalid);
         } else {
-            Verification::fail(Error::Invalid)
+            return Verification::fail(Error::Missing);
```
ssddOnTop, line 30: Do not use `contains_key` to check for the header, it adds unnecessary cost to the code. the problem is with de-serialization, check #1673 (feel free to use the code)

ssddOnTop (approved):

src/auth/basic.rs
```diff
@@ -108,6 +115,18 @@ testuser3:{SHA}Y2fEjdGT1W6nsLqtJbGUVeUp9e4=
...
+    async fn verify_auth_failure() {
+        let provider = setup_provider();
+        let mut req_ctx = RequestContext::default();
+        req_ctx.allowed_headers.insert(
+            "Authorization",
+            HeaderValue::from_static("Basic dGVzdHVzZXIyOm15cGFzc3dvcmQ"),
+        );
+        let validation = provider.verify(&req_ctx).await;
+        assert_eq!(validation, Verification::fail(Error::Invalid));
+    }
```
tusharmath, line 128: Add an integration test for this case.

amitksingh1490:
`resolve_token` function in `jwt_verify.rs` should also have similar changes which are done for basic auth. Also add integration test for this scenario
```rust
fn resolve_token(&self, request: &RequestContext) -> Option<String> {
        let value = request.allowed_headers.typed_get::<Authorization<Bearer>>();

        value.map(|token| token.token().to_owned())
    }
```

amitksingh1490 (approved):

PR #1699, opened 2024-04-11
fix: Specifying @link in config skips upstream.allowedHeaders

### LINKED ISSUE #1683: bug(config): Specifying @link in config breaks upstream.allowedHeaders setting
### Prerequisites

### Describe the bug

Take this config

```graphql
schema
  @link(src: "./stripe-types.graphql", type: Config)
  @upstream(baseURL: "https://api.stripe.com/v1/", allowedHeaders: ["Authorization"]) {
  query: Query
}

type Query {
  # BALANCE
  balance: BalanceRoot
    @http(
      method: "GET"
      path: "/balance"
    )
}

```

and 

```graphql
# Balance Schema

type Balance {
  amount: Int
  currency: String
  source_types: SourceTypes
}

type SourceTypes {
  card: Int
}

type ConnectReserved {
  amount: Int
  currency: String
}

type BalanceRoot {
  object: String
  livemode: Boolean
  pending: [Balance]
  connect_reserved: [ConnectReserved]
  available: [Balance]
}
```

If we run this example and try to pass the Authorization header we'll get an auth error from the upstream. 

And if we combine these files into single config file everything will work just fine.

### Steps to reproduce

1. Use config from above
2. Start the tailcall
3. Try to pass authorization header

### Expected behavior

Authorization header should be passed to the upstream if it's defined in user's request

### Actual behavior

The Authorization header is ignored

### Additional context

The bug happens because of the implementation of `merge_right` for Config::Upstream here https://github.com/tailcallhq/tailcall/blob/b86b37bb4471ff07db24c63d84e7c59df6e3a49e/src/config/upstream.rs#L173-L180 - in this code in case `other.allowedHeaders` is None then `self.allowedHeaders` is ignored completely and the result would be None.

amitksingh1490:
fixes #1683

### DIFF
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -233,14 +233,11 @@ impl MergeRight for Server {
 
 #[cfg(test)]
 mod tests {
-    use crate::config::ScriptOptions;
     use super::*;
+    use crate::config::ScriptOptions;
 
     fn server_with_script_options(so: ScriptOptions) -> Server {
-        Server {
-            script: Some(so),
-            ..Default::default()
-        }
+        Server { script: Some(so), ..Default::default() }
     }
 
     #[test]
--- a/src/config/server.rs
+++ b/src/config/server.rs
@@ -195,6 +195,7 @@ impl Server {
         self.pipeline_flush.unwrap_or(true)
     }
 }
+
 impl MergeRight for Server {
     fn merge_right(mut self, other: Self) -> Self {
         self.apollo_tracing = self.apollo_tracing.merge_right(other.apollo_tracing);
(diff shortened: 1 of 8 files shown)


src/merge_right.rs
```diff
@@ -1,13 +1,61 @@
...
+
+impl Scalar for char {}
+
+impl Scalar for String {}
+
+impl Scalar for HttpVersion {}
+
+impl Scalar for ScriptOptions {}
+
+impl Scalar for Proxy {}
```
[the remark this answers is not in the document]
amitksingh1490: Why is this a problem

src/merge_right.rs
```diff
@@ -1,13 +1,61 @@
...
+
+impl Scalar for bool {}
+
+impl Scalar for char {}
+
+impl Scalar for String {}
+
+impl Scalar for HttpVersion {}
+
+impl Scalar for ScriptOptions {}
```
tusharmath, line 52: I don't think ScriptOptions is a a scalar.

amitksingh1490: done

src/merge_right.rs
```diff
@@ -1,13 +1,65 @@
...
+
+impl Scalar for bool {}
+
+impl Scalar for char {}
+
+impl Scalar for String {}
+
+impl Scalar for HttpVersion {}
+
+impl MergeRight for ScriptOptions {
```
tusharmath, line 52: Move the implementation closer to where the type is located.

src/merge_right.rs
```diff
@@ -1,13 +1,65 @@
...
+    fn merge_right(self, other: Self) -> Self {
+        ScriptOptions { timeout: self.timeout.merge_right(other.timeout) }
+    }
+}
+
+impl Scalar for Proxy {}
```
tusharmath, line 58: Proxy isn't a Scalar.

src/merge_right.rs
```diff
@@ -1,13 +1,65 @@
...
+
+impl<A: Scalar> MergeRight for A {
```
tusharmath, line 60:
```suggestion
impl MergeRight for Scalar {
```

src/merge_right.rs
```diff
@@ -1,13 +1,65 @@
...
+
+impl Scalar for f32 {}
+
+impl Scalar for bool {}
+
+impl Scalar for char {}
+
+impl Scalar for String {}
+
+impl Scalar for HttpVersion {}
```
tusharmath, line 50: Create a separate file under src called `scalar`. Where all the scalars are implemented.

tusharmath: We can even call it `Primitive` since graphQL scalars already exist.

amitksingh1490: done file called primitives

src/config/server.rs
```diff
@@ -195,6 +195,7 @@ impl Server {
         self.pipeline_flush.unwrap_or(true)
     }
 }
+
```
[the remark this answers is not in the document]
amitksingh1490: untouched code path, we should do it in separate PR

tests/execution/test-merge-right-with-link-config.md
```diff
@@ -0,0 +1,45 @@
+---
+check_identity: false
+---
+
+# test-merge-right-with-link-config
```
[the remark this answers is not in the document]
amitksingh1490: I want the file name to be same as test name

src/config/server.rs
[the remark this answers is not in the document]
amitksingh1490: sure open an issue

tusharmath (approved):

PR #1727, opened 2024-04-14
refactor: Refactor execution_spec

### LINKED ISSUE #1500: refactor execution spec
Execution spec is getting out of control. Let's refactor! 

**Requirements**
- Create a mod for execution_spec
- Mod should have a `parse.rs` which parses the md files and creates ExecutionSpec
- `spec.rs` that actually creates and executes the spec.
- create a `runtime.rs` that contains all the mock runtime.
- Use https://crates.io/crates/datatest to generate tests via spec.rs.
- Rename `@assert` to `@test` everywhere in the spec.

webbdays:
**Summary:**  
Refactor execution_spec : spec.rs, runtime.rs, parse.rs was created in executionspec module.
we can use executionspec module inside execution_spec.rs to run the tests for each test file written in "tests/execution/".
changed @assert to @test
moved and renamed snaps accordingly.
removed old snapshots location.
i have performed --delete-unreferenced-snapshots .
removed success log in test_spec function.

**Issue Reference(s):**  
Fixes #1500
/claim #1500 

**Build & Testing:**

- [No] I ran `cargo test` successfully.
All tests passed.
But there was an error with open telemetry for some tests when execution_spec tests are run.
"OpenTelemetry metrics error occurred. Metrics error: reader is not registered"

**Checklist:**

- [yes] PR follows the naming convention of `<type>(<optional scope>): <title>`

## Summary by CodeRabbit

- **Refactor**
	- Restructured test execution code to improve organization and behavior.
- **New Features**
	- Introduced new structures and functionalities for parsing and processing execution specifications.
	- Enhanced handling of API requests and responses through new mocking mechanisms.
	- Added file I/O operations support for testing environments.
- **Tests**
	- Modified CI workflow to refine code coverage generation process.


webbdays:
Rename @assert to @test everywhere in the spec - still it is pending thing. will do it.

tusharmath:
Please run - 

- [ ] `cargo insta test --delete-unreferenced-snapshots`

tests/executionspec/spec.rs
```diff
@@ -0,0 +1,400 @@
...
+    let mut specs = vec![];
+    let contents = fs::read_to_string(path)?;
+    let spec: ExecutionSpec = ExecutionSpec::from_source(path, contents)
+        .await
+        .map_err(|err| err.context(path.to_str().unwrap().to_string()))?;
+    specs.push(spec);
+    specs = ExecutionSpec::filter_specs(specs);
+
+    for spec in specs.into_iter() {
+        test_spec(spec, &opentelemetry).await;
```
tusharmath, line 278: Why are we using `test_spec` if we are using datatest now?

webbdays: that is the existing assert_spec function.
renamed to test_spec.
using datatest-stable not datatest, both are different.
datatest-stable just runs the  test for each file right.
test_spec is the main thing that runs the test(perform snapshot assertions) for the given spec.

webbdays: specs variable at any time contain only 0 or 1 spec based on skip, only annotations.

webbdays:
removed old commits, commited new changes.

tests/executionspec/spec.rs
webbdays: datatest-stable is doing anything else, just run a test for each file.

webbdays:
@tusharmath

webbdays:
Hi @tusharmath,
I have cleaned all previous commits and commited a new one.

[ Done] cargo insta test --delete-unreferenced-snapshots

[No conflicts now] Resolve conflicts

[Removed test_spec success log] Drop the logs since you are using datatest

[ Pending, i am not aware of this. please guide] Fix reader is not registered error on console

[Looking into it, it seems the insta by default saving snapshots with prefix as module path in which insta is run.] All snapshots are being renamed, if it can be avoided we should try.

And i dont understand why code coverage is failing.

Thank you.

webbdays:
@tusharmath
please review.
opentelemetry error thing, i cannot understand the root cause for that.

webbdays:
for snapshots filenames i have remove module path prefix.
should i add execution_spec prefix manually?

tusharmath:
@meskill can you have a look at this issue.

meskill:
> @tusharmath please review. opentelemetry error thing, i cannot understand the root cause for that. help.

I took a look into opentelemetry issue.

That is because of the refactor since tests are now ran in parallel and that is causing issues with opentelemetry because we're using global state for telemetry providers and setting it to every test.

My suggestion would be to drop opentelemetry from execution_spec for now for following reasons:
- as mentioned it doesn't work with parallel execution and the data leaks between tests that leads to errors and inability to assert basically anything
- we don't actually use it any tests right now because they were unstable even for previous implementation.

tests/executionspec/spec.rs
meskill: and seems like we should remove `ExecutionSpec::filter_specs` and storing specs as vec anyway since it's not applicable anymore when using datatest.

I've tested that it works fine with standart cargo options https://doc.rust-lang.org/rustc/tests/index.html#filters

tests/executionspec/spec.rs
```diff
@@ -0,0 +1,403 @@
...
+    // todo
+    // Explicitly only run one test if specified in command line args
+    // This is used by test-convertor to auto-apply the snapshots of incompatible
+    // fail-annotated http specs
+
+    let mut specs = vec![];
+    let contents = fs::read_to_string(path)?;
+    let spec: ExecutionSpec = ExecutionSpec::from_source(path, contents)
+        .await
+        .map_err(|err| err.context(path.to_str().unwrap().to_string()))?;
```
meskill, line 276:
```suggestion
    let spec: ExecutionSpec = ExecutionSpec::from_source(path, contents)
        .await
        .with_context(|| path.display().to_string())?;
```

tests/executionspec/spec.rs
```diff
@@ -0,0 +1,403 @@
...
+    // TODO: reuse logic from server.rs to select the correct handler
+    if app_ctx.blueprint.server.enable_batch_requests {
+        handle_request::<GraphQLBatchRequest>(req, app_ctx).await
+    } else {
+        handle_request::<GraphQLRequest>(req, app_ctx).await
+    }
+}
+
+#[cfg(test)]
+pub mod test {
```
meskill, line 314: no need to wrap in separate module and that should be part of `runtime.rs`

tests/executionspec/spec.rs
```diff
@@ -0,0 +1,403 @@
...
+    fn from(value: Cause<String>) -> Self {
+        SDLError {
+            message: value.message.to_string(),
+            trace: value.trace.iter().map(|e| e.to_string()).collect(),
+            description: value.description,
+        }
+    }
+}
+
+async fn test_spec(spec: ExecutionSpec, opentelemetry: &InMemoryTelemetry) {
```
meskill, line 61: could we split this big function into smaller pieces to make it more readable?

meskill (changes requested):

tests/executionspec/spec.rs
webbdays: will do it.

tests/executionspec/parse.rs
```diff
@@ -0,0 +1,361 @@
...
+        // list.
+        if !only_specs.is_empty() {
+            only_specs
+        } else {
+            filtered_specs
+        }
+    }
+
+    pub async fn from_source(path: &Path, contents: String) -> anyhow::Result<Self> {
+        INIT.call_once(|| {});
```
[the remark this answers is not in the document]
webbdays: Yes, but its from old implementation. i didnt touched it.
Will clean it.

tests/executionspec/spec.rs
webbdays: ok.

webbdays:
https://github.com/tailcallhq/tailcall/pull/1727#discussion_r1570945823
we can do this way.
but, the test_names should include those filter strings.
currently, the tests are using annotations in the tests files.

tests/execution_spec.rs
```diff
@@ -1,1099 +1,13 @@
...
-        vec
-    };
-
-    for spec in spec.into_iter() {
-        assert_spec(spec, &opentelemetry).await;
-    }
+use executionspec::spec::load_and_test_execution_spec;
 
+fn run_execution_spec(path: &Path) -> datatest_stable::Result<()> {
+    let _result = tokio_test::block_on(load_and_test_execution_spec(path));
```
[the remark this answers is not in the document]
meskill: but here you're not propagating the error to datatest and in case it was Result::Err it will be ignored here, only panics will work

meskill:
> Hi @meskill, [#1727 (comment)](https://github.com/tailcallhq/tailcall/pull/1727#discussion_r1570945823) we can do this way. but, the test_names should include those filter strings. currently, we are using annotations in the tests files.

yeah, annotations were used to reimplement something like this, but it doesn't work with datatest as expected anymore since `only` annotation works only if we filter all set of specs. So I suggested to drop it and live with cargo test options

webbdays:
@meskill 
it works incase of only. can i know example.
in this case specs has only one spec.

webbdays:
@meskill 
should i remove opentelemtry usage?

meskill:
> @meskill should i remove opentelemtry usage?

yes, please, remove opentelemetry usage from execution_spec

tests/execution_spec.rs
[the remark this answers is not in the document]
webbdays: ok

webbdays:
@meskill 
done.
can you review now.

webbdays:
@meskill 
some how some settings in Cargo.toml are changed. pushed a commit with fix.

webbdays:
* To skip the test file from running we should save test file with extension .skip.md

.github/workflows/ci.yml
```diff
@@ -165,7 +165,7 @@ jobs:
           target: ${{ matrix.target }}
 
       - name: Run Cargo Test
-        run: cargo llvm-cov --all-features --lcov --output-path lcov.info
+        run: cargo llvm-cov --all-features --lcov --output-path lcov.info -- --skip ".skip.md"
```
meskill, line 168: ah, yeah, we have single test that we actually want to skip =)

changing ci is not a good option here because we want to also skip those tests when running locally.

Let's remove it from ci and move to execution_spec i.e. don't execute files with `.skip` there.

meskill (changes requested):

tests/executionspec/parse.rs
```diff
@@ -0,0 +1,293 @@
...
+            http,
+            http2_only,
+            file: Arc::new(MockFileSystem::new(self.clone())),
+            env: Arc::new(Env::init(env)),
+            cache: Arc::new(InMemoryCache::new()),
+            extensions: Arc::new(vec![]),
+        };
+
+        // TODO: move inside tailcall core if possible
+        init_metrics(&runtime).unwrap();
```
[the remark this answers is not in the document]
webbdays: Ok.

tests/executionspec/spec.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    // run query tests
+    run_query_tests_on_spec(spec, server, mock_http_client).await;
+}
+
+pub async fn load_and_test_execution_spec(path: &Path) -> anyhow::Result<()> {
+    // todo
+    // Explicitly only run one test if specified in command line args
+    // This is used by test-convertor to auto-apply the snapshots of incompatible
+    // fail-annotated http specs
```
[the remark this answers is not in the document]
webbdays: Ok.

.github/workflows/ci.yml
webbdays: --skip ".skip.md" skips all files with that filter.
when running locally we can run this command right.
without codecov
cargo test --test execution_spec -- --skip ".skip.md"

webbdays: > move to execution_spec i.e. don't execute files with `.skip` there.

how to do it there?
i think datatest-stable didnt provide any ways to skip.

webbdays: ok. i will see if possible with pattern

tests/executionspec/spec.rs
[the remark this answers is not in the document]
webbdays: this was done.

tests/executionspec/parse.rs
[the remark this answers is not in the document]
webbdays: this was done now.

.github/workflows/ci.yml
webbdays: Hi @meskill,
any idea of regex pattern such that we do not match the .skip.md files.
Thank you.

meskill: just leave regex as is and ignore the execution it in your test handler

webbdays: in what the datatest-stable will show it as passed not skipped.

webbdays: So, its better to go go with ci change if we want to use test files skipping.

webbdays: --skip is the best way here i think.

webbdays: @meskill ?

meskill: developers usually just call `cargo test` without additional options to run all tests. Introducing mandatory option for the call is inconvenient. 

just add additional logs from spec output about skipped test if you worry about misleading stats

webbdays: ok.

webbdays: so, we need Executionspec::filter_spec()

webbdays: then how do we deal with only case.

webbdays: i think the skip and only are for developmental purposes.
1. "only" is for running locally. such tests are not to push to codebase. we can reverage cargo test -- --test "only test name to run"
2. "skip" is to skip tests run locally as well as in ci because (sometimes we may want avoiding tests runs but still want to save it) for that we can use cargo test -- --skip "test pattern to skip"

webbdays: @meskill 
i think its not much effort to a developer.

webbdays: then how can we handle "only" as we are moving file by file.
previously, the logic is like
loads all specs and filter them out based on only and skip.
Which is not possible as we are moving test by test.

webbdays: we can handle the skip by printing skipped message.
but only is not like that.

webbdays: I think you understand.

meskill: sure, let me wrap it for you:
- `only` is used for experimenting purposes when developer locally want to focus on specific tests.  We don't mark `only` any tests on remote, only locally. That will be done by `cargo test` options now.
- `skip` is used to mark tests that are currently not working and we push that annotations to remote so they are not run by default. For that we need something like it worked previously either with special annotations or special file naming.

tests/executionspec/parse.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    pub server: Vec<(Source, String)>,
+    pub mock: Option<Vec<Mock>>,
+    pub env: Option<HashMap<String, String>>,
+    pub test: Option<Vec<APIRequest>>,
+    pub files: BTreeMap<String, String>,
+
+    pub check_identity: bool,
+    pub sdl_error: bool,
+}
```
tusharmath, line 54: move to `model.rs`

.github/workflows/ci.yml
webbdays: ok.

tests/executionspec/parse.rs
webbdays: ok.

tests/executionspec/runtime.rs
```diff
@@ -0,0 +1,382 @@
...
+use tailcall::http::{Method, Response};
+use tailcall::runtime::TargetRuntime;
+use tailcall::{blueprint, EnvIO, FileIO, HttpIO};
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use url::Url;
+
+use super::model::ExecutionSpec;
+
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct APIRequest {
```
tusharmath, line 27: move to `model.rs`

tests/executionspec/runtime.rs
```diff
@@ -0,0 +1,382 @@
...
+    #[serde(default = "default_status")]
+    pub status: u16,
+    #[serde(default)]
+    pub headers: BTreeMap<String, String>,
+    #[serde(default)]
+    pub body: serde_json::Value,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub text_body: Option<String>,
+}
```
tusharmath, line 53: move to model.rs

tests/executionspec/runtime.rs
```diff
@@ -0,0 +1,382 @@
...
+#[derive(Serialize, Deserialize, Clone, Debug)]
+struct UpstreamResponse(APIResponse);
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct Mock {
+    request: UpstreamRequest,
+    response: UpstreamResponse,
+    #[serde(default = "default_expected_hits")]
+    expected_hits: usize,
+}
```
tusharmath, line 75: move to model.rs

tests/executionspec/spec.rs
```diff
@@ -0,0 +1,321 @@
...
+    let contents = fs::read_to_string(path)?;
+    let spec: ExecutionSpec = ExecutionSpec::from_source(path, contents)
+        .await
+        .with_context(|| path.display().to_string())?;
+
+    match spec.runner {
+        Some(Annotation::Skip) => {
+            println!("{} ... {}", spec.path.display(), "skipped".blue());
+        }
+        Some(Annotation::Only) => {}
```
meskill, line 290: `only` doesn't do anything, let's drop it from available annotations

tests/executionspec/spec.rs
```diff
@@ -0,0 +1,321 @@
...
+
+pub async fn load_and_test_execution_spec(path: &Path) -> anyhow::Result<()> {
+    let contents = fs::read_to_string(path)?;
+    let spec: ExecutionSpec = ExecutionSpec::from_source(path, contents)
+        .await
+        .with_context(|| path.display().to_string())?;
+
+    match spec.runner {
+        Some(Annotation::Skip) => {
+            println!("{} ... {}", spec.path.display(), "skipped".blue());
```
meskill, line 288: looks a little bit off in the output.

```
test run_execution_spec::call-multiple-steps-piping.md                    ... ok
tests/execution/graphql-dataloader-batch-keys.md ... skipped
test run_execution_spec::graphql-dataloader-batch-keys.md                 ... ok
test run_execution_spec::batching-default.md                              ... ok
```

can we improve it?
You can take a look how ok tests are logged in https://github.com/nextest-rs/datatest-stable/blob/main/src/utils.rs#L34 and https://github.com/LukasKalbertodt/libtest-mimic/blob/master/src/printer.rs#L118

meskill:
We'll need to update docs https://tailcall.run/docs/contributors/tests/#annotation for this and add instructions how to filter tests

tests/executionspec/spec.rs
webbdays: looking into it.

tests/executionspec/spec.rs
webbdays: Ok.

tests/executionspec/spec.rs
webbdays: i think We need to track the max length of test file_name

tests/executionspec/runtime.rs
```diff
@@ -0,0 +1,382 @@
...
+    pub url: Url,
+    #[serde(default)]
+    pub headers: BTreeMap<String, String>,
+    #[serde(default)]
+    pub body: serde_json::Value,
+    #[serde(default)]
+    pub test_traces: bool,
+    #[serde(default)]
+    pub test_metrics: bool,
+}
```
tusharmath, line 39: Move to `model`

tests/executionspec/runtime.rs
```diff
@@ -0,0 +1,382 @@
...
+    response: UpstreamResponse,
+    #[serde(default = "default_expected_hits")]
+    expected_hits: usize,
+}
+
+#[derive(Clone, Debug)]
+struct ExecutionMock {
+    mock: Mock,
+    actual_hits: Arc<AtomicUsize>,
+}
```
tusharmath, line 81: Move to `model`

webbdays:
I have moved all structs to model? is it ok.

tusharmath:
@webbdays I have updated the PR. You just need to update the documentation as @meskill has pointed out.

webbdays:
@tusharmath 
I have updated the  doc for tests and made a pr.
https://github.com/tailcallhq/tailcallhq.github.io/pull/220

tests/executionspec/spec.rs
meskill: still an issue I think

tests/executionspec/spec.rs
meskill: if it's complicated let's not bother for now

tusharmath (approved):

PR #1733, opened 2024-04-15
fix(grpc): linking multiple proto files with the same package name

meskill:
**Summary:**  
Instead of searching for the specific FileDescriptorSet based on package name, collect all of the linked files into single FileDescriptorSet that will be used for searching services and messages

**Build & Testing:**

**Checklist:**

## Summary by CodeRabbit

- **New Features**
	- Enhanced error messaging for missing protobuf files in gRPC configuration.
	- Introduced a scenario test for handling multiple Protocol Buffer files with the same package name.

- **Bug Fixes**
	- Updated logic for constructing service names in gRPC methods to ensure accuracy.

- **Refactor**
	- Simplified the handling of file descriptor sets in gRPC configurations.
	- Improved merging logic for gRPC extensions.

- **Tests**
	- Added unit tests for new merging behaviors in gRPC configurations.


src/blueprint/operators/grpc.rs
```diff
@@ -166,8 +166,8 @@ pub fn compile_grpc(inputs: CompileGrpc) -> Valid<Expression, String> {
     Valid::from(GrpcMethod::try_from(grpc.method.as_str()))
         .and_then(|method| {
             Valid::from_option(
-                config_module.extensions.get_file_descriptor_set(&method),
-                format!("File descriptor not found for method: {}", grpc.method),
+                config_module.extensions.get_file_descriptor_set(),
```
tusharmath, line 169: We need to validate if the path to the method exists in the proto file.

src/config/config_module.rs
```diff
@@ -153,3 +163,93 @@ impl From<Config> for ConfigModule {
         ConfigModule { config, input_types, output_types, ..Default::default() }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    mod extensions {
+        mod merge_right {
```
tusharmath, line 170: Do we need these tests for merge_right? We are using macros now, so we should be able to depend on them.

tests/snapshots/execution_spec__test-grpc-invalid-proto-id.md_errors.snap
```diff
@@ -4,7 +4,7 @@ expression: errors
 ---
 [
   {
-    "message": "File descriptor not found for method: abc.NewsService.GetAllNews",
+    "message": "Protobuf files were not specified in the config",
```
tusharmath, line 7: I think this should be validated at compile time.

src/blueprint/operators/grpc.rs
meskill: we have multiple proto files and we'll check that method exists later when executing method resolving

src/config/config_module.rs
meskill: I've added impl MergeRight for external type FileDescriptorSet

tests/snapshots/execution_spec__test-grpc-invalid-proto-id.md_errors.snap
meskill: this is validation error snapshot for compilation stage

tests/execution/grpc-proto-with-same-package.md
```diff
@@ -0,0 +1,97 @@
...
+    repeated News news = 1;
+}
+```
+
+```graphql @server
+schema
+  @server(port: 8000, graphiql: true)
+  @upstream(httpCache: true, batch: {delay: 10})
+  @link(id: "errors", src: "errors.proto", type: Protobuf)
+  @link(id: "news", src: "news.proto", type: Protobuf) {
```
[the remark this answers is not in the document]
meskill: actually, copy-pasted from another test and kept the same approach

src/blueprint/operators/grpc.rs
tusharmath: Understood, but we are dropping a significant feature here, I don't see why we need to move that check to runtime when we can identify the issue at compile time.

tests/execution/grpc-proto-with-same-package.md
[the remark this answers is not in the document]
tusharmath: Let's add a simpler test here: Use types such as Foo and Bar in the same package. Also, we should test by making requests to the services defined in the two files. The current implementation has the error proto but it's not being used anywhere.

src/blueprint/operators/grpc.rs
meskill: The check it later in this file when calling `to_operation` and it's at compile time.

The error looks like this https://github.com/tailcallhq/tailcall/blob/2ddb53663be9b6204187076ce286bb56a0d8df71/tests/snapshots/execution_spec__test-grpc-service.md_errors.snap#L7

tusharmath (approved):

PR #1741, opened 2024-04-16
fix: duplicate read proto

### LINKED ISSUE #1734: Duplicated CLI output related to read files
### Prerequisites

### Describe the bug

When running `examples/grpc.graphql` example that has `@link` operator the read proto files are showed twice i.e.

```sh
[wsl] nixos in ❯ cargo run -- start examples/grpc.graphql
   Compiling tailcall v0.1.0 (/home/nixos/Projects/tailcall/tailcall)
    Finished dev [unoptimized + debuginfo] target(s) in 30.76s
     Running `target/debug/tailcall start examples/grpc.graphql`
 INFO File read: examples/grpc.graphql ... ok
 INFO File read: examples/../src/grpc/tests/proto/news.proto ... ok
 INFO File read: examples/../src/grpc/tests/proto/news.proto ... ok
 INFO N + 1 detected: 0
 INFO 🚀 Tailcall launched at [127.0.0.1:8000] over HTTP/1.1
 INFO 🌍 Playground: http://127.0.0.1:8000
 ```
 
 for file /src/grpc/tests/proto/news.proto

### Expected behavior

Output logs should show every file only once

### Actual behavior

Read files (at least related to protobuf) are showed twice for each file

### Additional context

Make sure it doesn't replicate for different types of link

### PR DESCRIPTION
**Pull request by Shylock-Hg**
**Summary:**  

According to @tusharmath 's comments. use HashMap to avoid duplicate read.

Current output:
```
shylock@localhost:~/Workspace/tailcall> ./target/debug/tailcall start examples/grpc.graphql 
 INFO File read: examples/grpc.graphql ... ok
 INFO File read: examples/../src/grpc/tests/proto/news.proto ... ok
 INFO N + 1 detected: 0
 INFO 🚀 Tailcall launched at [127.0.0.1:8000] over HTTP/1.1
 INFO 🌍 Playground: http://127.0.0.1:8000
```

**Issue Reference(s):**  
Close #1734
/claim #1734 

**Build & Testing:**

**Checklist:**

## Summary by CodeRabbit

- **New Features**
	- Introduced a caching mechanism to improve file reading efficiency in the application.
	- Enhanced resource reading capabilities with options for both cached and direct file access.
  
- **Enhancements**
	- Updated initialization processes for various readers to accommodate new caching features.
	- Added thread safety to file reading operations using mutex locks.

- **Refactor**
	- Modified method signatures to support new concurrent and caching functionalities.

### DIFF
--- a/src/resource_reader.rs
+++ b/src/resource_reader.rs
@@ -46,13 +46,6 @@ impl ResourceReader<Cached> {
 }
 
 #[async_trait::async_trait]
-impl<A: Reader + Send + Sync> Reader for ResourceReader<A> {
-    async fn read<T: ToString + Send>(&self, file: T) -> anyhow::Result<FileRead> {
-        self.0.read(file).await
-    }
-}
-
-#[async_trait::async_trait]
 pub trait Reader {
     async fn read<T: ToString + Send>(&self, file: T) -> anyhow::Result<FileRead>;
 }

### CODE REVIEW

File: src/resource_reader.rs [hunk 1]
```diff
@@ -13,14 +17,39 @@ pub struct FileRead {
 
 pub struct ResourceReader {
     runtime: TargetRuntime,
+    // Cache file content, path->content
+    cache: Arc<Mutex<HashMap<String, String>>>,
```
tusharmath, line 21: Why use an async-lock? I think a simple sync::Mutex should work for our usecase.

src/resource_reader.rs
```diff
@@ -13,14 +17,39 @@ pub struct FileRead {
...
 }
 
 impl ResourceReader {
-    pub fn init(runtime: TargetRuntime) -> Self {
-        Self { runtime }
+    pub fn init(runtime: TargetRuntime, cache: Arc<Mutex<HashMap<String, String>>>) -> Self {
```
tusharmath, line 25:
```suggestion
    pub fn init(runtime: TargetRuntime, cache: bool) -> Self {
```

If cache is true, we should initialize it inside of `init`

tusharmath: >And there is another issue, the first reading is called by ConfigReader::resource_reader, the second is called by ConfigReader::proto_reader, they are different object. So we need to get the HashMap from outside.

The cache should not be shared. If the cache needs to be shared, we should do that by sharing the ResourceReader.

tusharmath: I think that's just an additional unwrapping that makes the code difficult to read. You can create two resource readers
```rust
struct ResourceReader<A>(A);

impl<A> ResourceReader<A> {
  fn direct(runtime: TargetRuntime) -> ResourceReader<Direct>
  fn cached(runtime: TargetRuntime) -> ResourceReader<Cache>
}

// Reads the files directly without cache
Direct {runtime: TargetRuntime }

// Reads it and maintains a cache
pub Cache {
  reader: Direct,
  cache: Arc<Mutex<HashMap<String, FileRead>>
}

```

Can we do something like above? We will know at compile time, if the reader is caching or not.

src/resource_reader.rs
tusharmath: Performance is not a major concern in this part of the application.

src/generator/generator.rs
```diff
@@ -1,18 +1,23 @@
...
 
 pub struct Generator {
     proto_reader: ProtoReader,
 }
 impl Generator {
     pub fn init(runtime: TargetRuntime) -> Self {
-        Self { proto_reader: ProtoReader::init(runtime) }
+        Self {
+            proto_reader: ProtoReader::init(Arc::new(ResourceReader::init(runtime, false))),
+        }
```
tusharmath, line 20: Why are we changing the API here? I don't think it's required.

src/config/reader.rs
```diff
@@ -18,16 +18,17 @@ use crate::runtime::TargetRuntime;
 /// linked extensions to create a ConfigModule.
 pub struct ConfigReader {
     runtime: TargetRuntime,
-    resource_reader: ResourceReader,
+    resource_reader: Arc<ResourceReader>,
```
tusharmath, line 21:
```suggestion
    resource_reader: ResourceReader,
```
No need for an arc here, like TargetRuntime, implement clone on ResourceReader if required.

src/resource_reader.rs
tusharmath: It makes the code a lot more modular, type-safe and easy to test. The current implementation has a lot of checks before reading, increasing the potential

tusharmath (approved):

tusharmath (approved):

PR #1760, opened 2024-04-19
fix: parse empty response body as Value::Null

amitksingh1490:
**Problem 1**
Empty string could not be deserialized as JSON
**Problem 2**
We were wrapping null value in some, that is why the expected array error was thrown by graphql executor

## Summary by CodeRabbit

- **New Features**
	- Enhanced type handling to support list structures more effectively.
	- Improved API response handling to provide default values when expected data is missing.

- **Tests**
	- Added tests to ensure correct handling of empty arrays and objects in API responses.

### DIFF
--- a/src/blueprint/into_schema.rs
+++ b/src/blueprint/into_schema.rs
@@ -41,6 +41,7 @@ fn to_type(def: &Definition) -> dynamic::Type {
                 let field = field.clone();
                 let type_ref = to_type_ref(&field.of_type);
                 let field_name = &field.name.clone();
+                let is_list = field.of_type.is_list();
                 let mut dyn_schema_field = dynamic::Field::new(
                     field_name,
                     type_ref.clone(),
@@ -70,12 +71,17 @@ fn to_type(def: &Definition) -> dynamic::Type {
 
                                         let const_value =
                                             expr.eval(ctx, &Concurrent::Sequential).await?;
-
                                         let p = match const_value {
-                                            ConstValue::List(a) => FieldValue::list(a),
-                                            a => FieldValue::from(a),
+                                            ConstValue::List(a) => Some(FieldValue::list(a)),
+                                            a => {
+                                                if a == ConstValue::Null && is_list {
+                                                     FieldValue::NONE
+                                                } else {
+                                                    FieldValue::try_from(a).ok()
+                                                }
+                                            },
                                         };
-                                        Ok(Some(p))
+                                        Ok(p)
                                     }
                                     .instrument(span)
                                     .inspect_err(|err| tracing::error!(?err)),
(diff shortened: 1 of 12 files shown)


src/blueprint/into_schema.rs
```diff
@@ -70,12 +71,17 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
-
                                         let p = match const_value {
-                                            ConstValue::List(a) => FieldValue::list(a),
-                                            a => FieldValue::from(a),
+                                            ConstValue::List(a) => Some(FieldValue::list(a)),
+                                            a => {
+                                                if a == ConstValue::Null && is_list {
+                                                    FieldValue::NONE
+                                                } else {
+                                                    Some(FieldValue::from(a))
```
meskill, line 80: `try_from` here is just a blanket implementation that uses `from` thus it's infallible

src/blueprint/into_schema.rs
```diff
@@ -70,12 +71,17 @@ fn to_type(def: &Definition) -> dynamic::Type {
 
                                         let const_value =
                                             expr.eval(ctx, &Concurrent::Sequential).await?;
-
                                         let p = match const_value {
-                                            ConstValue::List(a) => FieldValue::list(a),
-                                            a => FieldValue::from(a),
+                                            ConstValue::List(a) => Some(FieldValue::list(a)),
+                                            a => {
+                                                if a == ConstValue::Null && is_list {
```
meskill, line 77: why return none only for list? 

I think it should return none if it's any null

tests/snapshots/execution_spec__test-null-in-object.md_test_0.snap
```diff
@@ -0,0 +1,18 @@
...
+expression: response
+---
+{
+  "status": 200,
+  "headers": {
+    "content-type": "application/json"
+  },
+  "body": {
+    "data": {
+      "hi": {
```
meskill, line 12: this is not quite right because the `hi` value should be null itself.

Imagine if nested fields were marked as required in this case we'll get an error although this is valid to have hi as null

meskill (changes requested):

src/blueprint/into_schema.rs
amitksingh1490: In case of object you need to return None on the field

src/blueprint/into_schema.rs
amitksingh1490: removed try_from as clippy was complaining

tests/snapshots/execution_spec__test-null-in-object.md_test_0.snap
amitksingh1490: Let me check the graphql spec.

PR #1789, opened 2024-04-24
fix: handle `Accept` header

### LINKED ISSUE #1766: feat: make tailcall "GraphQL over HTTP" compliant
Currently, tailcall doesn't support [GraphQL over HTTP Spec](https://graphql.org/learn/serving-over-http/)

**Technical Requirements**
- All existing tests should pass
- Add new tests for every compliance violation
- Ensure there is no performance degradation

**Additional Information**
Use https://graphql-http.com to check for spec compliance.

**Related Issues**
- For the `/graphql` route refer to https://github.com/tailcallhq/tailcall/issues/1790

Use the following configuration for testing.

```graphql
schema
  @server(
    port: 8000
    graphiql: true
    hostname: "0.0.0.0"
    headers: {cors: {allowOrigins: ["*"], allowHeaders: ["*"], allowMethods: [POST, GET, OPTIONS]}}
  )
  @upstream(baseURL: "http://jsonplaceholder.typicode.com") {
  query: Query
}

type Query {
}

type Post {
  id: Int!
  userId: Int!
  title: String!
  body: String!
}

```

webbdays:
**Summary:**  
1. Accept header handling

**Issue Reference(s):**  
#1766

**Build & Testing:**

- [yes ] I ran `cargo test` successfully.

**Checklist:**

- [ yes] I have performed a self-review of my code.
- [ yes] PR follows the naming convention of `<type>(<optional scope>): <title>`

### DIFF
--- a/src/http/request_handler.rs
+++ b/src/http/request_handler.rs
@@ -25,7 +25,6 @@ use crate::async_graphql_hyper::{GraphQLRequestLike, GraphQLResponse};
 use crate::blueprint::telemetry::TelemetryExporter;
 use crate::config::{PrometheusExporter, PrometheusFormat};
 
-
 pub const API_URL_PREFIX: &str = "/api";
 
 pub fn graphiql(req: &Request<Body>) -> Result<Response<Body>> {
--- a/src/http/request_handler.rs
+++ b/src/http/request_handler.rs
@@ -6,6 +6,7 @@ use std::sync::Arc;
 use anyhow::Result;
 use async_graphql::http::{playground_source, GraphQLPlaygroundConfig};
 use async_graphql::ServerError;
+use chrono::{Local, NaiveDate};
 use hyper::header::{self, CONTENT_TYPE};
 use hyper::http::Method;
 use hyper::{Body, HeaderMap, Request, Response, StatusCode};
@@ -13,6 +14,7 @@ use opentelemetry::trace::SpanKind;
 use opentelemetry_semantic_conventions::trace::{HTTP_REQUEST_METHOD, HTTP_ROUTE};
 use prometheus::{Encoder, ProtobufEncoder, TextEncoder, PROTOBUF_FORMAT, TEXT_FORMAT};
 use serde::de::DeserializeOwned;
+use serde_json::json;
 use tracing::Instrument;
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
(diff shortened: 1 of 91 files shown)


tests/execution/add-field-modify.md
```diff
@@ -43,6 +43,8 @@ type Query {
 ```yml @test
 - method: POST
   url: http://localhost:8080/graphql
+  headers:
+    Accept: application/graphql-response+json
```
tusharmath, line 47: Move this logic into execution spec so that by default we make the request with headers.

webbdays: I accept these are changes in many tests files.
Realistically, headers comes form request which we define in tests files. That should be the source of truth.
Also it will be the problem for us to write tests for non compliance.

src/http/request_handler.rs
```diff
@@ -264,6 +266,59 @@ async fn handle_request_inner<T: DeserializeOwned + GraphQLRequestLike>(
...
+                    ))
+                    .unwrap());
+            } else {
+                // on/after 1/1/2025, Accept header must contain
+                // "application/graphql-response+json"
+
+                let count = req
+                    .headers()
+                    .get("Accept")
+                    .unwrap()
```
tusharmath, line 294: let's not do `unwrap`

src/http/request_handler.rs
```diff
@@ -264,6 +266,59 @@ async fn handle_request_inner<T: DeserializeOwned + GraphQLRequestLike>(
...
+                    .to_str()
+                    .unwrap()
+                    .split(';')
+                    .filter(|i| i.trim() == "application/graphql-response+json")
+                    .count();
+                if count == 0 {
+                    let error_message = "Unsupported 'Accept' Header value in the request, required 'application/graphql-response+json'";
+
+                    // enforce this only on/after 1/1/2025, just warn until then.
+                    if NaiveDate::from_ymd_opt(2025, 1, 1) <= Some(Local::now().date_naive()) {
```
tusharmath, line 304: Is the date check required?

src/http/request_handler.rs
```diff
@@ -264,6 +266,59 @@ async fn handle_request_inner<T: DeserializeOwned + GraphQLRequestLike>(
...
+                                })
+                                .to_string(),
+                            ))
+                            .unwrap());
+                    } else {
+                        tracing::warn!(error_message);
+                    }
+                }
+            }
+
```
tusharmath, line 321: Let's put this header check in a separate function.

webbdays: ok. i also also think of that.
like creating separate module for spec checkers.

src/http/request_handler.rs
webbdays: ok.

tusharmath:
@webbdays Whenever the PR is ready please mark it as ready to review. I think I missed your comments here.

PR #1809, opened 2024-04-29
fix(grpc): show additional error info from expression

meskill:
**Summary:**  
Pass error extension info from grpc upstream in response. To make it mostly replace anyhow::Error with explicit EvaluationError and work with it

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/blueprint/into_schema.rs
+++ b/src/blueprint/into_schema.rs
@@ -2,6 +2,7 @@ use std::borrow::Cow;
 use std::sync::Arc;
 
 use async_graphql::dynamic::{self, FieldFuture, FieldValue, SchemaBuilder};
+use async_graphql::ErrorExtensions;
 use async_graphql_value::ConstValue;
 use futures_util::TryFutureExt;
 use tracing::Instrument;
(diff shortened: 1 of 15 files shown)


src/blueprint/into_schema.rs
```diff
@@ -68,8 +69,10 @@ fn to_type(def: &Definition) -> dynamic::Type {
                                         let ctx: ResolverContext = ctx.into();
                                         let ctx = EvaluationContext::new(req_ctx, &ctx);
 
-                                        let const_value =
-                                            expr.eval(ctx, &Concurrent::Sequential).await?;
+                                        let const_value = expr
+                                            .eval(ctx, &Concurrent::Sequential)
+                                            .await
+                                            .map_err(|err| err.extend())?;
```
tusharmath, line 75: Just a thought: We could use `into` here. Using `extend` doesn't seem very intuitive.

tusharmath (approved):

src/blueprint/into_schema.rs
meskill: no, `into` will use implicit implementation for Error https://docs.rs/async-graphql/latest/src/async_graphql/error.rs.html#322-330 that loses additional information

PR #1867, opened 2024-05-05
feat(js): JS based resolvers

ssddOnTop:


src/lambda/expression.rs
```diff
@@ -22,6 +22,7 @@ pub enum Expression {
     Cache(Cache),
     Path(Box<Expression>, Vec<String>),
     Protect(Box<Expression>),
+    Js(Arc<dyn WorkerIO<Option<ConstValue>, ConstValue>>),
```
tusharmath, line 25: Worker should not be stored in the Expression. It should be part of the runtime.

src/lambda/js_ctx.rs
```diff
@@ -0,0 +1,101 @@
...
+
+            let function = fn_as_value
+                .as_function()
+                .ok_or(anyhow::anyhow!("`hello` is not a function"))?;
+            let val: String = function.call((value, ))
+                .map_err(|_| anyhow::anyhow!("unable to parse value from js function: {} maybe because it's not returning a string?", name,))?;
+            Ok::<_, anyhow::Error>(serde_json::from_str(&val)?)
+        })
+    })
+}
```
tusharmath, line 101: Should reuse the cli runtime.

src/lib.rs
```diff
@@ -83,7 +84,7 @@ pub trait Cache: Send + Sync {
 pub type EntityCache = dyn Cache<Key = u64, Value = ConstValue>;
 
 #[async_trait::async_trait]
-pub trait WorkerIO<In, Out>: Send + Sync + 'static {
+pub trait WorkerIO<In, Out>: Debug + Send + Sync + 'static {
```
tusharmath, line 87: Do we need this?

src/blueprint/operators/js.rs
```diff
@@ -0,0 +1,64 @@
...
+    return js_disabled(inputs);
+
+    #[cfg(feature = "js")]
+    js_enabled(inputs)
+}
+
+#[cfg(not(feature = "js"))]
+fn js_disabled(_: CompileJs) -> Valid<Expression, String> {
+    Valid::fail("js is disabled or unsupported on this platform".to_string())
+}
```
amitksingh1490, line 24:
```suggestion
fn js_disabled(_: CompileJs) -> Valid<Expression, String> {
    Valid::fail("js is disabled or unsupported on this platform".to_string())
}
```

src/blueprint/operators/js.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub name: &'a str,
+    pub script: &'a Option<String>,
+}
+
+pub fn compile_js(inputs: CompileJs) -> Valid<Expression, String> {
+    #[cfg(not(feature = "js"))]
+    return js_disabled(inputs);
+
+    #[cfg(feature = "js")]
+    js_enabled(inputs)
```
amitksingh1490, line 18:
```suggestion
    js_enabled(inputs)
```

src/blueprint/operators/js.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::valid::{Valid, Validator};
+
+pub struct CompileJs<'a> {
+    pub name: &'a str,
+    pub script: &'a Option<String>,
+}
+
+pub fn compile_js(inputs: CompileJs) -> Valid<Expression, String> {
+    #[cfg(not(feature = "js"))]
+    return js_disabled(inputs);
```
amitksingh1490, line 15:
```suggestion
    return Valid::fail("js is disabled or unsupported on this platform".to_string());
```

src/blueprint/operators/js.rs
```diff
@@ -0,0 +1,64 @@
...
+        |script| {
+            Valid::from(
+                ctx.with(|ctx| {
+                    ctx.eval::<(), &str>(script)?;
+                    Ok::<_, anyhow::Error>(())
+                })
+                .map_err(|e| crate::valid::ValidationError::new(e.to_string())),
+            )
+            .map(|_| {
+                Expression::Js(std::sync::Arc::new(crate::javascript::Runtime::new(
```
amitksingh1490, line 42: drop ARC from here, Expression should not contain runtime info, it should be available in app_ctx

src/lambda/expression.rs
```diff
@@ -22,6 +23,7 @@ pub enum Expression {
     Cache(Cache),
     Path(Box<Expression>, Vec<String>),
     Protect(Box<Expression>),
+    Js(Arc<Runtime>),
```
tusharmath, line 26: - Move `Js` to `IO::JS`
- In TargetRuntime, hold an implementation of `WorkerIO`

src/javascript/enable_js/runtime.rs
```diff
@@ -47,18 +49,33 @@ impl LocalRuntime {
 
 pub struct Runtime {
     script: blueprint::Script,
+    function_name: String,
```
tusharmath, line 52: drop `function_name` use it from the worker_io::call method.

src/javascript/enable_js/runtime.rs
```diff
@@ -75,27 +92,13 @@ impl Drop for Runtime {
 
 #[async_trait::async_trait]
 impl WorkerIO<Event, Command> for Runtime {
-    async fn call(&self, name: String, event: Event) -> anyhow::Result<Option<Command>> {
+    async fn call(&self, event: Event) -> anyhow::Result<Option<Command>> {
```
tusharmath, line 95: put the `name` back.

src/javascript/enable_js/runtime.rs
```diff
@@ -105,6 +108,40 @@ impl WorkerIO<Event, Command> for Runtime {
     }
 }
 
+#[async_trait::async_trait]
+impl WorkerIO<Option<ConstValue>, ConstValue> for Runtime {
```
tusharmath, line 112:
```suggestion
impl WorkerIO<ConstValue, ConstValue> for Runtime {
```

src/core/javascript/runtime_no_js.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub fn new(_: blueprint::Script) -> Self {
+        panic!("JavaScript runtime is not supported in this build")
+    }
+}
+
+#[async_trait::async_trait]
+impl WorkerIO<Option<ConstValue>, ConstValue> for Runtime {
+    async fn call(&self, _: String, _: Option<ConstValue>) -> anyhow::Result<Option<ConstValue>> {
+        panic!("JavaScript runtime is not supported in this build")
+    }
```
tusharmath, line 18: Bail out instead of panic.

src/core/blueprint/operators/js.rs
```diff
@@ -0,0 +1,60 @@
...
+use crate::core::config;
+use crate::core::config::Field;
+use crate::core::lambda::IO;
+use crate::core::try_fold::TryFold;
+use crate::core::valid::Valid;
+use crate::{ConfigModule, Expression, ValidationError, Validator};
+
+pub struct CompileJs<'a> {
+    pub name: &'a str,
+    pub script: &'a Option<String>,
```
tusharmath, line 11: script might not be needed, because script is specified via link.

src/core/blueprint/operators/js.rs
```diff
@@ -0,0 +1,60 @@
...
+    Valid::fail("js is disabled or unsupported on this platform".to_string())
+}
+
+#[cfg(feature = "js")]
+fn js_enabled(inputs: CompileJs) -> Valid<Expression, String> {
+    let name = inputs.name;
+    let quickjs = rquickjs::Runtime::new().unwrap();
+    let ctx = rquickjs::Context::full(&quickjs).unwrap();
+
+    Valid::from_option(inputs.script.as_ref(), "script is required".to_string()).and_then(
```
tusharmath, line 33: Just validate if the config has a worker script linked. No need of using quickjs for this.

src/cli/mod.rs
```diff
@@ -1,8 +1,6 @@
 mod command;
 mod error;
 mod fmt;
-#[cfg(feature = "js")]
-pub mod javascript;
```
tusharmath, line 5: We should keep this feature flag. CLI need to have a quickjs based implementation of the worker.

src/cli/javascript/js_request.rs
```diff
@@ -9,9 +9,7 @@ use rquickjs::{FromJs, IntoJs};
 use serde::{Deserialize, Serialize};
 
 use crate::core::is_default;
-
-#[derive(Debug)]
-pub struct JsRequest(reqwest::Request);
```
tusharmath, line 14: Keep the JsRequest as it is. Move it to core, but drop the `FromJs` and `IntoJs` from core. Move those implementations to `cli::javascript::worker`

src/lib.rs
```diff
@@ -1,4 +1,4 @@
-mod core;
+pub mod core;
```
tusharmath, line 1: Do not expose core publicly. Control the things you wish to expose publicly down below.

src/core/lambda/io.rs
```diff
@@ -133,6 +137,17 @@ impl IO {
 
                     Ok(res.body)
                 }
+                IO::Js { name: method } => {
+                    let value = ctx
+                        .request_ctx
+                        .runtime
+                        .resolver_worker
+                        .call(method.clone(), ctx.value().cloned())
```
tusharmath, line 145: We need to pass the whole context.

src/core/javascript.rs
```diff
@@ -0,0 +1,27 @@
...
+#[derive(Debug)]
+pub enum Event {
+    Request(JsRequest),
+}
+
+#[derive(Debug)]
+pub enum Command {
+    Request(JsRequest),
+    Response(JsResponse),
+}
```
tusharmath, line 27: Rename file to worker.rs

src/core/javascript.rs
```diff
@@ -0,0 +1,27 @@
...
+    async fn call(&self, _: String, _: A) -> anyhow::Result<Option<B>> {
+        anyhow::bail!("JavaScript runtime is not supported in this build")
+    }
+}
+
+#[derive(Debug)]
+pub struct JsResponse(pub Response<String>);
+
+#[derive(Debug)]
+pub struct JsRequest(pub reqwest::Request);
```
tusharmath, line 16:
```suggestion
pub struct WorkerRequest(pub reqwest::Request);
```

src/core/javascript.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[async_trait::async_trait]
+impl<A: Send + Sync + 'static, B> WorkerIO<A, B> for DefaultJsRuntime {
+    async fn call(&self, _: String, _: A) -> anyhow::Result<Option<B>> {
+        anyhow::bail!("JavaScript runtime is not supported in this build")
+    }
+}
+
+#[derive(Debug)]
+pub struct JsResponse(pub Response<String>);
```
tusharmath, line 13:
```suggestion
pub struct WorkerResponse(pub Response<String>);
```

src/core/runtime.rs
```diff
@@ -26,6 +27,10 @@ pub struct TargetRuntime {
     /// A list of extensions that can be used to extend the runtime's
     /// functionality or integrate additional features.
     pub extensions: Arc<Vec<SchemaExtension>>,
+    /// Worker middleware for handling HTTP requests.
+    pub http_worker: Arc<dyn WorkerIO<Event, Command>>,
+    /// Worker middleware for resolving data.
+    pub resolver_worker: Arc<dyn WorkerIO<ConstValue, ConstValue>>,
```
tusharmath, line 33:
```suggestion
    pub worker: Arc<dyn WorkerIO<ConstValue, ConstValue>>,
```

src/cli/runtime/mod.rs
```diff
@@ -33,6 +34,30 @@ fn init_hook_http(http: Arc<impl HttpIO>, script: Option<blueprint::Script>) ->
...
+fn init_resolver_worker_io(
+    script: Option<blueprint::Script>,
+) -> Arc<dyn WorkerIO<async_graphql::Value, async_graphql::Value>> {
+    #[cfg(feature = "js")]
+    if let Some(script) = script {
+        return Arc::new(super::javascript::Runtime::new(script));
+    };
+
+    #[cfg(not(feature = "js"))]
+    tracing::warn!("JS capabilities are disabled in this build");
```
tusharmath, line 57: This will log twice. May be move this logic to where TargetRuntime is being initialized.

src/core/lambda/io.rs
```diff
@@ -133,6 +137,18 @@ impl IO {
 
                     Ok(res.body)
                 }
+                IO::Js { name: method } => {
+                    let value = ctx.value().cloned().unwrap_or_default();
+                    let value = ctx
+                        .request_ctx
+                        .runtime
+                        .worker
+                        .call(method.clone(), value)
```
tusharmath, line 146: Hot code should avoid clones as much as possible. Can you update the signature to pass the function name as an &str.

ssddOnTop: difficult..

src/core/lambda/io.rs
```diff
@@ -144,6 +160,12 @@ impl<'a, Ctx: ResolverContextLike<'a> + Sync + Send> CacheKey<EvaluationContext<
             IO::Http { req_template, .. } => req_template.cache_key(ctx),
             IO::Grpc { req_template, .. } => req_template.cache_key(ctx),
             IO::GraphQL { req_template, .. } => req_template.cache_key(ctx),
+            IO::Js { name: method } => {
+                // TODO
+                let mut hasher = std::hash::DefaultHasher::new();
+                method.hash(&mut hasher);
+                hasher.finish()
+            }
```
tusharmath, line 168: Intresting... I think you have hit an edge case.

While you are at it, can you add this documentation:

> Generating `cache_key` allows us to cache IO calls within the context of an GraphQL request. This ensures that if an IO is executed once, we don't execute it again and wait for the initiated request to finish or serve from cache if the data is already available. The cache_key is generated based on what's the IO type AND what's the value in ctx.


Hopefully the documentation above makes sense. The problem in your implementation is that the JS function will be called only once irrespective of what the input is because the JS function takes context as input and you don't use it for hashing.


Assuming that you can only insert one `@js` on one field, the caching doesn't make a lot of sense because the ctx is almost always going to be different. In essense — JS doesn't fit the current requirements of a typical IO, which is that it should be cacheable. 

There are two ways we can go from here — 
1. Move JS out of IO, but that's not right, JS can fail, and in other environments it might actually be an IO.
2. Drop the requirement of being cacheable for IO. That means `cache_key` should return an `Option<u64>` instead.


My inclination is towards the second one. It's possible that there other IO operations that aren't cacheable perhaps.

Phew!

tusharmath: resolved in #2009

src/core/mod.rs
```diff
@@ -83,7 +84,7 @@ pub type EntityCache = dyn Cache<Key = u64, Value = ConstValue>;
 #[async_trait::async_trait]
 pub trait WorkerIO<In, Out>: Send + Sync + 'static {
     /// Calls a global JS function
-    async fn call(&self, name: String, input: In) -> anyhow::Result<Option<Out>>;
+    async fn call(&self, name: Cow<'async_trait, str>, input: In) -> anyhow::Result<Option<Out>>;
```
tusharmath, line 87: Take `name` as a reference. Clone internally if required.

tusharmath (approved):

tusharmath (approved):

PR #1895, opened 2024-05-08
fix(grpc): type and field names for config from protobuf generation

meskill:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


generated/.tailcallrc.schema.json
```diff
@@ -1474,7 +1474,9 @@
           "enum": [
             "Str",
             "Num",
-            "Bool"
+            "Bool",
+            "Empty",
+            "Any"
```
tusharmath, line 1479: tailcallrc should also have the generated types.

src/core/proto_reader/reader.rs
```diff
@@ -178,21 +178,19 @@ mod test_proto_config {
...
-        let helper_map = reader
+        let file_descriptors = reader
             .resolve(reader.read_proto(&test_file, None).await?, Some(test_dir))
             .await?;
-        let files = test_dir.read_dir()?;
-        for file in files {
-            let path = file?.path();
+        for file in file_descriptors
+            .iter()
+            .filter(|desc| !desc.name.as_ref().unwrap().starts_with("google/protobuf/"))
```
tusharmath, line 186:
```suggestion
            .filter_map(|desc| !desc.name.as_ref().map(|a| a.starts_with("google/protobuf/")))
```

src/core/generator/graphql_type.rs
```diff
@@ -1,122 +1,116 @@
...
+    fn combine(mut self, other: Namespace) -> Self {
+        self.path.extend(other.path);
+
+        self
     }
 
-    fn source(&self) -> &str {
-        &self.input
+    fn id(&self) -> String {
+        self.path.join(PACKAGE_SEPARATOR)
```
tusharmath, line 50: `source` was providing the original name of the type.  What is it doing now?

src/core/generator/graphql_type.rs
```diff
@@ -189,12 +183,9 @@ mod tests {
     fn test_from_enum_variant() {
         let input: Vec<TestParams> = vec![
             // Enum variants
-            ((Entity::EnumVariant, None, "foo"), "FOO"),
-            ((Entity::EnumVariant, None, "a.b.c.foo"), "FOO"),
-            ((Entity::EnumVariant, Some("a.b.c"), "foo"), "FOO"),
-            ((Entity::EnumVariant, Some("a.b"), "d.e.foo"), "FOO"),
-            ((Entity::EnumVariant, Some(""), "a.b.c.foo"), "FOO"),
-            ((Entity::EnumVariant, None, "a_b_c_foo"), "A_B_C_FOO"),
```
tusharmath, line 197: Let's keep the older tests.

generated/.tailcallrc.schema.json
meskill: you mean `.tailcallrc.graphql`? 
yeah, that types are not populated on generation.
The issue is with graphql syntax at least since JsonSchema type is generated as 

```
input Schema {
  Obj: JSON
  Arr: Schema
  Opt: Schema
  Enum: [String!]
}
```

We can solve it by adding nesting to scalar types in schema instead of storing simple types in root enum so output could look like 
```
```

src/core/generator/graphql_type.rs
meskill: I removed `source`. The id for methods is generated based on added namespaces

src/core/proto_reader/reader.rs
meskill: it's a code in the test I think it's okay to leave the unwrap. I expect it to fail here in such a case instead of ignoring it. And that adds more logic in the test itself though it's already has too much conditions that makes it less declarative and understandable for a test

tusharmath (approved):

PR #1937, opened 2024-05-13
fix: improvements

amitksingh1490:
fix: parsing constant url at compile time (#1936)


tusharmath:
Moving this to draft. Can't see any significant change in performance from main.

PR #1939, opened 2024-05-13
fix: allow default values for input arguments

### LINKED ISSUE #1931: bug: allow default values for input arguments.
### Prerequisites

### Describe the bug

currently the default values passed to input arguments are not used to query the data.

### Expected behavior

1. In following example, we pass 1 as default value to id.
```graphql
type Query {
  user(id: Int = 1): User @http(path: "/users/{{.args.id}}")
}
```
2. when we make request from playground with following query then default id should be taken into account and request with default id should be made.
```graphql
# client(playground) side query.
query {
 user {
     name  
  }
}
```
```url
request made to upstream : http://jsonplaceholder.typicode.com/users/1
```

### Actual behavior

1. In following example, we pass 1 as default value to id.
```graphql
type Query {
  user(id: Int = 1): User @http(path: "/users/{{.args.id}}")
}
```
2. but when we make request from playground with following query then default id is not taken into account and request without id is made.
```graphql
# client(playground) side query.
query {
 user {
     name  
  }
}
```
```url
request made to upstream : http://jsonplaceholder.typicode.com/users/
```

ssddOnTop:
**Issue Reference(s):**  
Fixes #1931
/claim 1931


src/core/blueprint/into_schema.rs
```diff
@@ -42,17 +43,37 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
+                        Some((v.name.clone(), val))
+                    })
+                    .collect::<HashMap<String, serde_json::Value>>();
+                let args = if args.is_empty() {
+                    ConstValue::Null
+                } else {
+                    async_graphql_value::to_value(args).unwrap_or_default()
+                };
+
                 let mut dyn_schema_field = dynamic::Field::new(
```
tusharmath, line 63: Just call the the `dyn_schema_field.arguments(args)` Current implementation is a bit complicated and unnecessary.

ssddOnTop:
there is an issue with async-graphql as well, I raised a PR https://github.com/async-graphql/async-graphql/pull/1527 there if it gets merged, I'll finish touchups in this PR.

Issue #1869, opened 2024-05-06
fix: Enable JS build for all architecture


amitksingh1490:
**Issue Description:**

We recently migrated from `V8` to `QuickJS` for our `JavaScript` runtime. We initially utilized feature flags for disabling the build of `JS `features for architectures where V8 was incompatible.

Now, we aim to enable `JS` feature across all architectures. 

> Note: While this may potentially encounter issues on certain architectures, we should iteratively enable it for those where it works.

The expectation is to update the CI to build Tailcall on all supported platforms without any feature flags. 

**Additional Context:**

- Unsupported platforms
	- [ ] linux-x64-musl
	- [ ] linux-arm64-musl
	- [ ] linux-ia32-gnu
	- [ ] win32-x64-gnu
	- [ ] win32-arm64-msvc
	- [ ] win32-ia32-gnu

- How it works on CI
	https://github.com/tailcallhq/tailcall/blob/dd2114deb7c939d44798b945f8491b1801678125/.github/workflows/ci.yml#L266

tusharmath:
/bounty 180$

**IMPORTANT:** The bounty is priced at 30$ per platform. So in case only two platforms are supported we will pay only 60$.

tusharmath:
@varshith257 @mehulmathur16 The builds are failing on main. Which is not allowing us to make any releases. Can you please check this out URGENTLY.

https://github.com/tailcallhq/tailcall/actions/runs/9092466656/job/24989725401
https://github.com/tailcallhq/tailcall/actions/runs/9092466656/job/24989725711

mehulmathur16:
> @varshith257 @mehulmathur16 The builds are failing on the main, which is not allowing us to make any releases. Can you please check this out URGENTLY?
> 
> https://github.com/tailcallhq/tailcall/actions/runs/9092466656/job/24989725401
> https://github.com/tailcallhq/tailcall/actions/runs/9092466656/job/24989725711
> 

@tusharmath on it

varshith257:
> @varshith257 @mehulmathur16 The builds are failing on the main, which is not allowing us to make any releases. Can you please check this out URGENTLY?
> 
> https://github.com/tailcallhq/tailcall/actions/runs/9092466656/job/24989725401 https://github.com/tailcallhq/tailcall/actions/runs/9092466656/job/24989725711

Taking a look on it

varshith257:
[...]

PR #1941, opened 2024-05-13
fix: js build on some architectures

mehulmathur16:
This PR fixes tailcall build on these platforms:

linux-x64-musl
linux-arm64-musl
linux-ia32-gnu
win32-x64-gnu
win32-arm64-msvc

/claim #1869

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -263,7 +263,6 @@ jobs:
             os: ubuntu-latest
             rust: stable
             target: x86_64-unknown-linux-musl
-            features: --no-default-features --features cli
             libc: musl
             cross: true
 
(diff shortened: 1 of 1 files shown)


ssddOnTop:
maybe try this dockerfile (musl):

```Dockerfile
FROM alpine:latest

ENV RUSTUP_HOME=/usr/local/rustup \
    CARGO_HOME=/usr/local/cargo \
    PATH=/usr/local/cargo/bin:$PATH

RUN apk add --no-cache build-base curl openssl-dev clang

RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable
```

PR #1977, opened 2024-05-16
chore: Clean up merge logic for vectors of key-value structs.

### LINKED ISSUE #1701: Improve  merging logic in  struct
There's a need to improve the merging logic for  in the  struct to use a more idiomatic Rust approach, avoiding manual index handling. This issue was identified during a PR review. For more context, see the PR at https://github.com/tailcallhq/tailcall/pull/1699 and the specific comment at https://github.com/tailcallhq/tailcall/pull/1699#discussion_r1560651753.

TheWizardTower:
The issue is that the original code is not idiomatic rust, since it's directly indexing acc via acc[pos]. However, looking at merge_key_value_vecs, it already prefers the right value, so I _think_ we can just have the last two lines and get the same result.

Including the refactored code as a comment so we can have a discussion on the PR, we may be doing the merge twice for cache line performance reasons.

**Summary:**  
I've rewritten the logic to merge the two vector maps to be more idiomatic Rust, and avoid a bare vector index. However, I think we can just call `merge_key_value_vecs()` and achieve the same result, since that function biases towards the right values anyway.

It's possible you all are doing it for optimization reasons, so I would appreciate review from folks who are more familiar with the codebase than I am.

**Issue Reference(s):**  
Fixes #1701 

**Build & Testing:**

**Checklist:**

- [ ] I have updated the [documentation] accordingly. (not sure what documentation exists for this)


src/core/config/server.rs
```diff
@@ -97,15 +97,19 @@ pub struct Server {
...
+    //         None => acc.push(kv.clone()),
+    //         Some(pos) => {
+    //             match acc.get_mut(pos) {
+    //                 None => (), // Something deeply strange has happened.
+    //                 Some(lkv) => *lkv = kv.clone(),
+    //             }
+    //         }
+    //     }
+    //     acc
+    // });
```
tusharmath, line 111: delete commented code.

src/core/config/key_values.rs
```diff
@@ -27,15 +27,13 @@ pub struct KeyValue {
...
-        .map(|(k, v)| KeyValue { key: k.to_string(), value: v.to_string() })
+    current
+        .iter()
+        .chain(other.iter())
+        .map(|KeyValue { key, value }| (key, value))
+        .collect::<BTreeMap<_, _>>()
+        .into_iter()
+        .map(|(k, v)| KeyValue { key: k.to_owned(), value: v.to_owned() })
         .collect()
 }
```
tusharmath, line 38: There a little too much going on here. I think we can keep it simple and a bit more performant:

- Create one `HashMap<String, String>`
- Iterate over current and push all the values into the hashmap
- Next iterate over the second one and push the values into the hashmap
- Lastly convert the HashMap into Vec and return.

Your implementation doesn't basically the same thing, but it's not very clear from the implementation and does few unnecessary allocations on the way.

TheWizardTower: Switched over!

tusharmath (approved):

tusharmath:
LGTM! Thanks @TheWizardTower

PR #2002, opened 2024-05-21
feat(grpc): additional scalar types

### LINKED ISSUE #1896: feat(grpc): Implement additional Scalar Types
Add new Scalar types that could be used for creating config and map protobuf types.

Refer the [RFC](https://docs.google.com/document/d/1QPJ_WhobgtjLcPzp6_UfOIS0MRixQDVUj0FxM0mx5M8/edit?usp=sharing) about scalar types that needs to be implemented.

- [ ] BigInt 
- [ ] UnsignedInt
- [ ] Bytes

meskill:
**Summary:**  

**Issue Reference(s):**  
Fixes #1896

**Build & Testing:**

**Checklist:**


tests/core/snapshots/test-scalars.md_10.snap
```diff
@@ -0,0 +1,27 @@
...
+{
+  "status": 200,
+  "headers": {
+    "content-type": "application/json"
+  },
+  "body": {
+    "data": null,
+    "errors": [
+      {
+        "message": "internal: invalid value for scalar \"PhoneNumber\", expected \"FieldValue::Value\"",
```
tusharmath, line 14: This isn't a very meaningful error for the user.

tests/execution/test-scalars.md
```diff
@@ -1,4 +1,4 @@
-# test-scalar-email
+# Test scalars
 
 ```graphql @config
 # this is already defined scalars in tailcall
```
tusharmath, line 4:
```suggestion
# these are pre-defined scalars in Tailcall
```

tests/core/snapshots/test-scalars.md_10.snap
meskill: this is the error from async_graphql we don't specify it. All of the scalar errors looks like this rn

tests/execution/test-scalars.md
```diff
@@ -46,6 +49,21 @@ type Query {
   body:
     query: '{ any1: any(value: { test: "abc" } ), any2: any(value: "string-value") }'
 
+- method: POST
+  url: http://localhost:8000/graphql
+  body:
+    query: "{ bigInt1: bigInt(value: 32147483647), bigInt2: bigInt(value: -32147483647) }"
```
tusharmath, line 55:
```suggestion
    query: "{ bigInt1: bigInt(value: "32147483647"), bigInt2: bigInt(value: "-32147483647") }"
```

- Scalars should always be sent as a string. 
- The test for bigint should contain a real big int.

src/core/scalar/big_int.rs
```diff
@@ -0,0 +1,50 @@
+use async_graphql_value::ConstValue;
+use schemars::schema::Schema;
+use schemars::{schema_for, JsonSchema};
+
+/// Represents signed integer type 64 bit size
+#[derive(JsonSchema, Default)]
+pub struct BigInt(pub i64);
```
tusharmath, line 7: BigInt are actually strings. This is not truly a bigint.

src/core/generator/from_proto.rs
```diff
@@ -223,11 +223,18 @@ fn graphql_type_from_ref(name: &str) -> Result<GraphQLType<Unparsed>> {
...
         "double" | "float" => "Float",
-        "int32" | "int64" | "fixed32" | "fixed64" | "uint32" | "uint64" => "Int",
+        "int32" | "sint32" | "fixed32" | "sfixed32" => "Int",
+        "int64" | "sint64" | "fixed64" | "sfixed64" => "Int64Str",
+        "uint32" => "UInt32",
+        "uint64" => "Uint64Str",
         "bool" => "Boolean",
-        "string" | "bytes" => "String",
+        "string" => "String",
+        "bytes" => "Bytes",
```
tusharmath, line 237: I think for numbers we should use the following:

| Rust Types                     | Protobuf Primitive            | GraphQL Scalar   |
|--------------------------------|-------------------------------|------------------|
| `i8`, `i16`, `i32`             | `int32`                       | `Int`            |
| `u8`, `u16`, `u32`             | `uint32`                      | `Int`            |
| `i64`                          | `int64`                       | `BigInt`         |
| `u64`                          | `uint64`                      | `BigInt`         |
| `f32`                          | `float`                       | `Float`          |
| `f64`                          | `double`                      | `Float`          |
| `bool`                         | `bool`                        | `Boolean`        |
| `char`                         | `string`                      | `String`         |
| `usize`, `isize`               | `int64` or `uint64`           | `BigInt`         |

Generally the consumer of GraphQL is going to be a JS developer who doesn't work with so many types usually. It would be easier on them to reduce it to some simple one that are compatible with something that exists in JS. This helps build a better intuition for a developer who is consuming a GraphQL API. It's hard for a JS engineer to make sense of what should `UInt32` map to natively.

tusharmath (changes requested):
Fix comments.

src/core/generator/from_proto.rs
meskill: I'm not sure that wrapping so many different types into single one is a robust solution since the conversion from BigInt to any other int type will lose the precision or fail. And for grpc case, API expects a value with defined restrictions that could be expressed with respective scalar type. In that case scalar acts as validation step before sending invalid data to the upstream or generating conversion error. Also it acts as documentation on type level that helps to understand API expectations without a need to search for API docs or explore validation errors.

I feel like having such validation as scalar type (if we improve error message) will be easier to understand compared to conversion error or API error that could appear in such cases.

meskill: I think my approach to scalars is similar to graphql-scalars package https://the-guild.dev/graphql/scalars/docs#why whilst single `BigInt` type simplifies definitions but doesn't reflect actual API restrictions

PR #2008, opened 2024-05-22
fix: add support to list type arguments

### LINKED ISSUE #1991: bug: allow support to list arguments.
### Prerequisites

### Describe the bug

following tailcall configurations doesn't work, it throws error saying ` can't use list type 'ids' here`.

```graphql
type Query {
   f(ids: [Int!]!): B @http(path: "api/v1/users?ids={{.args.ids}}")
}
```

```graphql
type Query {
   f(ids: [Int!]!): B @http(path: "api/v1/users", query: [{key: "ids", value: "{{.args.ids}}"}])
}
```

### Expected behavior

When passed list of arguments, it should get applied to path and should construct url accordingly.
eg.
```url
  www.examples.com/api/v1/users?ids=1,2,3,4,5
```

### Actual behavior

it throws error saying ` can't use list type 'ids' here`.

### Technical Requirements
- When passed list of arguments, it should get applied to path and should construct url accordingly.
- Add unit tests wherever necessary to cover the new functionality.

laststylebender14:
**Summary:**  
- added support for list type arguments.
- added missing integration tests.

**Issue Reference(s):**  
Fixes #1991

**Build & Testing:**

**Checklist:**

### DIFF
--- a/tests/execution/test-params-as-body-with-list.md
+++ b/tests/execution/test-params-as-body-with-list.md
@@ -1,4 +1,4 @@
-# Http with args as body
+# Http with args as body with list
 
 ```graphql @config
 schema @server(port: 8000) @upstream(baseURL: "http://jsonplaceholder.typicode.com") {


tests/execution/test-list-argument.md
```diff
@@ -0,0 +1,32 @@
...
+
+type T1 {
+  numbers: [Int]
+}
+```
+
+```yml @mock
+- request:
+    method: GET
+    url: http://localhost:3000/api?q=1,2,3
```
tusharmath, line 20: There are multiple ways to encode this — 

```curl
# Option 1
/api?q=1,2,3

# Option 2
/api?q=1&q=2&q=3

# Option 3
/api?q[]=1&q[]=2&q[]=3
```

I think default should be option 2, but the user should be able to decide which option to use for encoding lists in the url. 

This setting should be specified in `upstream` however it can be overridden in `@http` if required.

tusharmath: The main problem that I see here, is that this way of handling lists is very different from how we handle lists in Bulk APIs.

src/core/config/config.rs
```diff
@@ -386,6 +386,9 @@ pub struct Arg {
     pub modify: Option<Modify>,
     #[serde(default, skip_serializing_if = "is_default")]
     pub default_value: Option<Value>,
+    /// Flag to indicate if the type inside the list is required.
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub list_type_required: bool,
```
tusharmath, line 391: drop these changes if they aren't working from this PR. Reduce reviewer overhead :)

src/core/mustache.rs
```diff
@@ -48,6 +48,21 @@ impl Mustache {
         }
     }
 
+    pub fn render_http(&self, value: &impl RequestString, method: &reqwest::Method) -> String {
```
tusharmath, line 51: mustache doesn't care about HTTP. We shouldn't have this a dependency to this function. Mustache doesn't even care it being a URL.

PR #2009, opened 2024-05-22
chore: make cache_key return `Option<u64>`

ssddOnTop:


src/core/lambda/io.rs
```diff
@@ -51,10 +51,12 @@ impl Eval for IO {
     ) -> Pin<Box<dyn Future<Output = Result<ConstValue, EvaluationError>> + 'a + Send>> {
         if ctx.request_ctx.upstream.dedupe {
             Box::pin(async move {
-                let key = self.cache_key(&ctx);
+                let key = self
+                    .cache_key(&ctx)
+                    .ok_or(anyhow::anyhow!("Unable to generate Cache Key"))?;
```
amitksingh1490, line 56: I think this should not fail
it should skip deduping and log a warning
and call self.eval_inner(ctx)
@tusharmath Please suggest

src/core/lambda/cache.rs
```diff
@@ -40,7 +40,9 @@ impl Eval for Cache {
     ) -> Pin<Box<dyn Future<Output = Result<ConstValue, EvaluationError>> + 'a + Send>> {
         Box::pin(async move {
             if let Expression::IO(io) = self.expr.deref() {
-                let key = io.cache_key(&ctx);
+                let key = io
+                    .cache_key(&ctx)
+                    .ok_or(anyhow::anyhow!("Unable to generate Cache Key"))?;
```
tusharmath, line 45: WE should only cache if the key is available.

tusharmath (approved):

PR #2019, opened 2024-05-24
fix: resolve union type in response

### LINKED ISSUE #2020: bug: Using Union type in output generates validation error in runtime
### Prerequisites

### Describe the bug

Specifying union type as output generates error like below.

```
internal: invalid value for union \"FooBar\", expected \"FieldValue::WithType\"
```

### Steps to reproduce

I created a pr with changes in test related to unions https://github.com/tailcallhq/tailcall/pull/2019/files that demonstrates the issue.

### Additional Context

To be able to actually use unions `async_graphql` library requires to specify type name for every value that is returned as a union. Pr with test has commented code that fixes only test example and allows to get response properly but proper solution requires more general way to solve this.

meskill:
**Summary:**  

**Issue Reference(s):**  
Fixes #2020 

**Build & Testing:**

**Checklist:**


src/core/blueprint/into_schema.rs
```diff
@@ -34,14 +38,120 @@ fn to_type_ref(type_of: &Type) -> dynamic::TypeRef {
...
+/// Entity that analyzes resolved data and decides what variant type of union
+/// this data satisfies to. It's required to specify exact type in
+/// `async_graphql` otherwise validation error will be generated at runtime
+#[derive(Clone)]
+struct UnionTypeResolver {
+    type_name: String,
+    schemas: Arc<Option<Vec<(String, JsonSchema)>>>,
+}
+
+impl UnionTypeResolver {
```
tusharmath, line 50: Let's move this to `union_resolver.rs`

src/core/blueprint/into_schema.rs
```diff
@@ -34,14 +38,120 @@ fn to_type_ref(type_of: &Type) -> dynamic::TypeRef {
...
 }
 
-fn to_type(def: &Definition) -> dynamic::Type {
+/// Entity that analyzes resolved data and decides what variant type of union
+/// this data satisfies to. It's required to specify exact type in
+/// `async_graphql` otherwise validation error will be generated at runtime
+#[derive(Clone)]
+struct UnionTypeResolver {
+    type_name: String,
+    schemas: Arc<Option<Vec<(String, JsonSchema)>>>,
```
tusharmath, line 47: Keep only a `Vec<(String, JsonSchema)>`

src/core/blueprint/into_schema.rs
tusharmath: I have a different proposal for this implementation. Ideally we would identify and store the discriminator while converting to blueprint. If we can't find the discriminator for a union type, we should throw a validation error, stating that there is no possible way to discriminate between the variants of the given union type.


Once we identify the discriminator we should create new IR:


```rust
enum IR {
  // ...
  Discriminate(Discriminator, Box<IR>)
  // ...
}

struct Discriminator(Vec<FieldName, TypeName>)

```

This particular IR when evaluated will check to identify the type of the output of the inner IR and store it in EvaluationCtx. In `into_schema` we can just read it and set it in the field value.

Let's discuss on discord if it isn't clear.

I feel this approach will be significantly superior to all existing Union type implementations.

tusharmath (changes requested):
The union type check doesn't need to be a deep check. A discriminator needs to be identified and stored in the expression at compile time.

src/core/ir/evaluation_context.rs
```diff
@@ -23,9 +24,11 @@ pub struct EvaluationContext<'a, Ctx: ResolverContextLike<'a>> {
 
     // Overridden Arguments for Async GraphQL Context
     graphql_ctx_args: Option<Arc<Value>>,
+
+    pub type_name: Option<TypeName>,
```
tusharmath, line 28: Add some doc about what is being stored here.

src/core/ir/discriminator.rs
```diff
@@ -0,0 +1,1054 @@
...
+    Vec(Vec<String>),
+}
+
+#[derive(Clone)]
+pub struct Discriminator {
+    types: Vec<String>,
+    fields_info: IndexMap<String, FieldInfo>,
+}
+
+impl std::fmt::Debug for Discriminator {
```
tusharmath, line 25: Do we need the debugs?

src/core/ir/discriminator.rs
```diff
@@ -0,0 +1,1054 @@
...
+use anyhow::{bail, Result};
+use async_graphql::Value;
+use indenter::indented;
+use indexmap::IndexMap;
+
+use crate::core::config::Type;
+use crate::core::valid::{Cause, Valid, Validator};
+
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub enum TypeName {
```
tusharmath, line 14: Can you add more documentation about what each type does and the overall algorithm.

src/core/ir/discriminator.rs
meskill: I'm suggesting to leave the debug fmt because it'll help in debugging since it outputs the info stored in `FieldInfo` as type name strings instead of raw usize. I've also added `tracing::debug!` in Discriminator in case we need to debug something in runtime

src/core/ir/evaluation_context.rs
meskill: added doc

src/core/ir/discriminator.rs
meskill: added more documentation

Cargo.toml
```diff
@@ -159,6 +159,7 @@ base64 = "0.22.1"
 tailcall-hasher = { path = "tailcall-hasher" }
 serde_json_borrow = "0.3.0"
 Inflector = "0.11.4"
+indenter = "0.3.3"
```
tusharmath, line 162: Do we need this? It seems like it's only used to implement Debug.

src/core/valid/valid.rs
```diff
@@ -157,6 +157,16 @@ impl<A, E> Valid<A, E> {
...
+impl<A: std::fmt::Debug, E: std::fmt::Debug> Valid<A, E> {
+    pub fn unwrap(self) -> A {
+        self.0.unwrap()
+    }
+
+    pub fn unwrap_err(self) -> ValidationError<E> {
+        self.0.unwrap_err()
+    }
+}
+
```
tusharmath, line 169: can be moved to `test` config. Unwrap isn't a feature that Valid need to offer.

src/core/ir/discriminator.rs
```diff
@@ -0,0 +1,1113 @@
...
+    }
+}
+
+/// Representation for set of types if some condition is hold.
+/// The condition is represented as a bit inside the usize number
+/// where bit position from the right in binary representation of usize
+/// is the index of type in the set and the if the value of the bit is
+/// 1 then the condition is hold
+#[derive(Copy, Clone, Default)]
+struct Repr(usize);
```
tusharmath, line 271: Why not use a u128?

src/core/ir/discriminator.rs
```diff
@@ -0,0 +1,1121 @@
...
+    }
+
+    /// Create new BitVec where condition is hold
+    /// for type with passed index
+    fn set_at(self, id: usize) -> Self {
+        Self(self.0 | (1 << id))
+    }
+
+    /// Search for first type in list for which condition is hold
+    fn first_covered_type<'types>(&self, types: &'types [String]) -> &'types str {
```
tusharmath, line 314: We don't need to pass the slice of String to this BitVec. Keep the BitVec generic and multi-purpose.

src/core/ir/discriminator.rs
```diff
@@ -0,0 +1,1121 @@
...
+        let mut result = Vec::new();
+
+        while x.0 != 0 {
+            result.push(x.first_covered_type(types));
+
+            *x = *x & (*x - 1);
+        }
+
+        result
+    }
```
tusharmath, line 330: Two points:
1. We shouldn't bind the BitVec's implementation to this special usecase of array of strings.
2. Could we benefit from performance, if we mutated the original array in place?

src/core/ir/discriminator.rs
tusharmath: We need to have better bit_vec specific names here.

src/core/ir/discriminator.rs
```diff
@@ -0,0 +1,1121 @@
...
+#[derive(Copy, Clone, Default)]
+struct BitVec(usize);
+
+impl std::fmt::Debug for BitVec {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("{:0b}", self.0))
+    }
+}
+
+impl Deref for BitVec {
```
tusharmath, line 279: Not a good idea to implement Deref here. We should use Deref when the primary value is what's inside for eg: Box, Arc, Mutex etc. BitVec, is a great DS that encapsulates the idea of a vector for boolean values. We could have a `as_usize` method if really needed.

src/core/ir/discriminator.rs
```diff
@@ -0,0 +1,1121 @@
...
+/// BitVecesents the type name for the resolved value.
+/// It is used when the GraphQL executor needs to resolve values of a union
+/// type. In order to select the correct fields, the executor must know the
+/// exact type name for each resolved value. And in cases when the output is
+/// list of union type it should resolve exact type for every entry in list.
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub enum TypeName {
+    Single(String),
+    Vec(Vec<String>),
+}
```
tusharmath, line 22: How about storing a ref of string?

src/core/ir/discriminator.rs
```diff
@@ -0,0 +1,1121 @@
...
+        for (field, info) in &self.fields_info {
+            if obj.contains_key(field.as_str()) {
+                *possible_types &= *info.presented_in;
+            } else {
+                *possible_types &= !*info.required_in;
+            }
+
+            if possible_types.is_empty() {
+                // no possible types, something is wrong with the resolved value
+                bail!("Failed to find corresponding type for value")
```
tusharmath, line 246: integration tests for all possible bailout conditions.

Cargo.toml
meskill: yes, this only for debug, but debug is valuable

src/core/valid/valid.rs
meskill: since it easily could be done with `valid.to_result().unwrap()` anyway, is there a point of hiding it?

src/core/ir/discriminator.rs
meskill: u128 would imply using multiple registries and multiple cpu instructions to perform arithmetic operations while for usize I believe it is covered with single cpu registry and instruction

src/core/ir/discriminator.rs
meskill: replaced Deref/DerefMut with required operations traits

src/core/ir/discriminator.rs
meskill: The unconformable part is that async_graphql API requires 'static str or String to pass https://docs.rs/async-graphql/latest/async_graphql/dynamic/struct.FieldValue.html#method.with_type

I've put an experimental implementation as separate pr for this using `String::leak` here https://github.com/tailcallhq/tailcall/pull/2210

src/core/ir/discriminator.rs
meskill: if we want to have something multi-purpose for bitvec we can use https://docs.rs/bitvec/latest/bitvec/index.html.

But here I prefer to have specific implementation for this use-case since we can built more precise and more performant solution (ofc until usize repr is enough for us, otherwise bitvec should be better)

src/core/ir/discriminator.rs
meskill: 1. I'm not sure we should start premature abstractions at least until we have clear cases to do so. Bitvec representation have it's own drawbacks
2. I'm not sure I've got the question and what array do you mean?

src/core/ir/discriminator.rs
meskill: added tests

src/core/valid/valid.rs
tusharmath: We need to ask this question —
1. Is Valid a wrapper on Result?
2. Is Valid is a value that allows horizontal composition of errors?

If we consider definition 1, then yes, exposing the common Result methods on top of Valid makes sense. However if we consider that Valid is a type of its own and usage of Result type internally is merely an implementation detail, then exposing unwrap doesn't seem right. For eg: Let us change the implementation of Valid to not use a Result type but something custom, in that case we will have an overhead of supporting this API.

src/core/ir/discriminator.rs
tusharmath: Rust docs: https://doc.rust-lang.org/1.23.0/std/ops/trait.Deref.html
> Deref should only be implemented for smart pointers to avoid confusion.

Also look at — https://stackoverflow.com/questions/45086595/is-it-considered-a-bad-practice-to-implement-deref-for-newtypes

src/core/ir/eval.rs
```diff
@@ -12,10 +12,10 @@ use crate::core::serde_value_ext::ValueExt;
 
 impl IR {
     #[tracing::instrument(skip_all, fields(otel.name = %self), err)]
-    pub fn eval<'a, Ctx>(
+    pub fn eval<'a, 'b, Ctx>(
         &'a self,
-        ctx: &'a mut EvalContext<'a, Ctx>,
-    ) -> Pin<Box<impl Future<Output = Result<ConstValue, Error>> + 'a>>
+        ctx: &'b mut EvalContext<'a, Ctx>,
+    ) -> Pin<Box<dyn Future<Output = Result<ConstValue, Error>> + Send + 'b>>
```
tusharmath, line 18: IMPORTANT: This is introducing a dynamic dispatch. Please revert.

src/core/ir/discriminator.rs
meskill: well, there are mixed opinions on it based on discussion =)

But, I agree that we shouldn't overuse it and I've already replaced it

src/core/ir/eval.rs
meskill: I'm trying to resolve the issue about not having the Mutex in EvaluationContext here.

Since we removed Eval trait from main I'm struggling to fix the code now because of the errors rustc generates. I raised the discussion https://users.rust-lang.org/t/rpitit-allows-more-flexible-code-in-comparison-with-raw-rpit-in-inherit-impl/113417 about this. 

In short I can't specify impl here without breaking usage later because of the mutable reference to ctx. dyn works right now. Adding Eval trait will also solve this.

tusharmath (approved):

PR #2053, opened 2024-05-29
feat: allows user to have byte based control over httpcache

### LINKED ISSUE #2035: feat: allow users to specify the http cache size in terms of total bytes rather than number of entries.
### Description:
Currently, in our system, there is provision to set the HTTP cache size in terms of the number of entries a cache can hold, which may not be the most efficient approach. I propose allowing users to specify the cache size in terms of total bytes rather than the number of entries.

Presently, users define the cache size as the maximum number of entries it can hold using the `@upstream(httpCache: X) ` directive. 

 Following example limits the cache to 42 entries.
```graphql
@upstream(httpCache: 42)
``` 
 Following example specifies a cache of 2 MB.
```graphql
@upstream(httpCache: 2097152)
``` 

This feature request suggests transitioning to a byte-based specification to provide more granular control over cache size allocation and better accommodate diverse data sizes and resource requirements.

### Technical requirements
- implement the above defined feature.
- write unit test cases when necessary.
- update the documentation with byte based httpCache.
- update data tests to support the byte based httpCache.

sirish123:
**Summary:**  
The 4 major files changed were that of upstream.rs (in two locations /blueprint and /core) , http.rs and cache.rs, **httpcachepercenatge** parameter was introduced so the user is allowed to specify the max percentage of memory allowed for the cache. If **httpcachepercenatge** is specified then it uses **HttpCacheManagerByteBased** which uses Lru policy and takes the weight(size) of a item stored for eviction if reaches the threshold or else uses the default **HttpCacheManager** which uses the number of entries

**Issue Reference(s):**  
Fixes #2035 

**Build & Testing:**

**Checklist:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5074,6 +5074,7 @@ checksum = "7b2093cf4c8eb1e67749a6762251bc9cd836b6fc171623bd0a9d324d37af2417"
 name = "tailcall"
 version = "0.1.0"
 dependencies = [
+ "Inflector",
  "anyhow",
  "async-graphql",
  "async-graphql-extension-apollo-tracing",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,6 +158,7 @@ tokio-test = "0.4.4"
 base64 = "0.22.1"
 tailcall-hasher = { path = "tailcall-hasher" }
 serde_json_borrow = "0.3.0"
+Inflector = "0.11.4"
 
 [dev-dependencies]
 tailcall-prettier = { path = "tailcall-prettier" }
--- a/src/core/config/config.rs
+++ b/src/core/config/config.rs
@@ -610,6 +610,19 @@ pub struct AddField {
 }
 
 impl Config {
+    pub fn is_root_operation_type(&self, type_name: &str) -> bool {
+        let type_name = type_name.to_lowercase();
+
+        [
+            &self.schema.query,
+            &self.schema.mutation,
+            &self.schema.subscription,
+        ]
+        .iter()
+        .filter_map(|&root_name| root_name.as_ref())
+        .any(|root_name| root_name.to_lowercase() == type_name)
+    }
+
     pub fn port(&self) -> u16 {
         self.server.port.unwrap_or(8000)
     }
@@ -885,5 +898,44 @@ mod tests {
         expected.insert("Bar".to_string());
 
         assert_eq!(actual, expected);
+    }
+
+    #[test]
+    fn test_is_root_operation_type_with_query() {
+        let mut config = Config::default();
+        config.schema.query = Some("Query".to_string());
+
+        assert!(config.is_root_operation_type("Query"));
+        assert!(!config.is_root_operation_type("Mutation"));
+        assert!(!config.is_root_operation_type("Subscription"));
+    }
+
+    #[test]
+    fn test_is_root_operation_type_with_mutation() {
+        let mut config = Config::default();
+        config.schema.mutation = Some("Mutation".to_string());
+
+        assert!(!config.is_root_operation_type("Query"));
+        assert!(config.is_root_operation_type("Mutation"));
+        assert!(!config.is_root_operation_type("Subscription"));
+    }
+
+    #[test]
+    fn test_is_root_operation_type_with_subscription() {
+        let mut config = Config::default();
+        config.schema.subscription = Some("Subscription".to_string());
+
+        assert!(!config.is_root_operation_type("Query"));
+        assert!(!config.is_root_operation_type("Mutation"));
+        assert!(config.is_root_operation_type("Subscription"));
+    }
+
+    #[test]
+    fn test_is_root_operation_type_with_no_root_operation() {
+        let config = Config::default();
+
+        assert!(!config.is_root_operation_type("Query"));
+        assert!(!config.is_root_operation_type("Mutation"));
+        assert!(!config.is_root_operation_type("Subscription"));
     }
 }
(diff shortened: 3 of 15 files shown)


generated/.tailcallrc.graphql
```diff
@@ -368,6 +368,11 @@ directive @upstream(
   """
   httpCache: Int
   """
+  Providing httpCachePercentage allocates a maximum of that percentage of RAM to the 
+  HTTP cache. Defaults to `0` if unspecified.
+  """
+  httpCachePercentage: String
```
tusharmath, line 374:
```suggestion
  httpCachePercentage: Float
```

src/cli/runtime/http.rs
```diff
@@ -59,6 +60,28 @@ impl RequestCounter {
...
+        .map(|p| if p > 100.0 { 100.0 } else { p })
+        .unwrap_or(0.0);
+
+    // Calculate the memory based on the percentage, without losing precision
+    let memory_by_percentage = (total_memory as f64 * clamped_percentage) as u64;
+
+    // Clamp the memory_by_percentage to u64::MAX
+
+    std::cmp::min(memory_by_percentage, u64::MAX)
+}
```
tusharmath, line 83: Can move this logic to a separate file `cache_memory.rs`. Should be calculated once and stored statically.

src/cli/runtime/http.rs
```diff
@@ -281,4 +318,70 @@ mod tests {
...
+        let url3 = format!("http://localhost:{}/test-3", port);
+        let resp = make_request(&url3, &native_http).await;
+        assert_eq!(resp.headers.get("x-cache-lookup").unwrap(), "MISS");
+
+        let resp = make_request(&url3, &native_http).await;
+        assert_eq!(resp.headers.get("x-cache-lookup").unwrap(), "HIT");
+
+        let resp = make_request(&url1, &native_http).await;
+        assert_eq!(resp.headers.get("x-cache-lookup").unwrap(), "MISS");
+    }
```
tusharmath, line 386: Move tests to `cache_size.rs`

tailcall-http-cache/src/cache.rs
```diff
@@ -69,6 +128,35 @@ impl CacheManager for HttpCacheManager {
     }
 }
 
+#[async_trait::async_trait]
+impl CacheManager for HttpCacheManagerByteBased {
```
tusharmath, line 132:
```suggestion
impl CacheManager for HttpMemoryCapCache {
```

tusharmath (changes requested):
Minor refactors required.

generated/.tailcallrc.graphql
sirish123: Wont this break the current system as httpcache is u64?

generated/.tailcallrc.graphql
```diff
@@ -368,6 +368,11 @@ directive @upstream(
   """
   httpCache: Int
```
sirish123, line 369: suggested new struct might break this part 
as current definiton `@upstream(httpCache: 100)`

generated/.tailcallrc.graphql
sirish123: @tusharmath another reason I chose **_string_** was that , **_float_** does not support `Eq` And the _**upstream**_ data structure was designed to support `Eq` hence the workaround with string

src/cli/runtime/http.rs
sirish123: done

src/cli/runtime/http.rs
sirish123: done

tailcall-http-cache/src/cache.rs
sirish123: done

generated/.tailcallrc.graphql
tusharmath: We can make breaking changes until we reach version 1

sirish123: > @tusharmath another reason I chose **_string_** was that , **_float_** does not support `Eq` And the _**upstream**_ data structure was designed to support `Eq` hence the workaround with string

@tusharmath I tried solving this issue but could not find a way the error is as follows  
image

One thing I found was ordered float https://docs.rs/ordered-float/latest/ordered_float/ but this did not work as well, because it violated  _**schemars::JsonSchema**_ and **_Serialize_**

sirish123: @tusharmath all other comments addressed and changed as per suggestions only this has been taken in as a string as f32 is not possible

tusharmath: Keep it as a `u64` with a range in between `0-100`. Anything beyond that will be considered as `100`.

src/core/config/upstream.rs
```diff
@@ -82,6 +81,11 @@ pub struct Upstream {
     /// Providing httpCache size enables Tailcall's HTTP caching, adhering to the [HTTP Caching RFC](https://tools.ietf.org/html/rfc7234), to enhance performance by minimizing redundant data fetches. Defaults to `0` if unspecified.
     pub http_cache: Option<u64>,
 
+    #[serde(default, skip_serializing_if = "is_default")]
+    /// Providing httpCachePercentage allocates a maximum of that percentage of
+    /// RAM to the HTTP cache. Defaults to `0` if unspecified.
+    pub http_cache_percentage: Option<String>,
```
tusharmath, line 87: Having two cache settings is quite confusing. I won't think we should do that.

generated/.tailcallrc.graphql
sirish123: Ok

PR #2093, opened 2024-06-02
chore: introduce field index

ssddOnTop:
ext of #2091 and #2090

**TODO**

~~- [ ] Impl serde_json::Value -> BorrowedValue~~
- [ ]


src/core/ir/jit/mod.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(unused)]
```
tusharmath, line 1: drop unused from the file. use it in places where there are todos only

PR #2139, opened 2024-06-08
feat: enum aliases

### LINKED ISSUE #2127: feat: Support Alias Mapping for Enum Values in GraphQL
### Description:
We need to extend the current functionality of the GraphQL `enum` type to support alias mappings for enum values. example, for the `enum Foo`, we want to map the value `A` to the aliases "Alpha" and "Ace". 

This feature will allow for more flexibility in handling enum values, particularly when integrating with external systems that may use different representations for the same logical value. The value should be parsed and the aliases should be correctly mapped to their respective enum values.

The config should look like this:

```graphql
enum Foo {
  A @alias(options: ["Alpha", "Ace"])
  B
}
```

In above example, the enum value `A` should be recognized for `A`, `Alpha`, and `Ace`.

### Acceptance Criteria:
1. Support for specifying aliases for enum values in the config.
2. Aliases should be correctly recognized and mapped to the corresponding enum values.
3. Add integration tests for this use case.
4. Update the documentation on the website

Shylock-Hg:
**Summary:**  

**Issue Reference(s):**  
Close #2127 
/claim #2127 

**Build & Testing:**

**Checklist:**


src/core/ir/mod.rs
```diff
@@ -54,6 +56,41 @@ impl IR {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct Map {
+    pub input: Box<IR>,
+    // accept key return value instead of
+    pub map: BTreeMap<String, ConstValue>,
```
tusharmath, line 63:
```suggestion
    pub map: HashMap<String, String>,
```
Why need a const value?

generated/.tailcallrc.schema.json
```diff
@@ -93,6 +93,21 @@
...
+      "properties": {
+        "options": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "uniqueItems": true
+        }
+      }
+    },
```
tusharmath, line 110: .tailcallrc.graphql should also have generated documentation for directives.

src/core/ir/mod.rs
```diff
@@ -54,6 +56,41 @@ impl IR {
...
+                } else {
+                    Err(EvaluationError::ExprEvalError(format!(
+                        "Can't find mapped key: {}.",
+                        key
+                    )))
+                }
+            } else {
+                Err(EvaluationError::ExprEvalError(
+                    "Mapped key must be string value.".to_owned(),
+                ))
```
tusharmath, line 88: Based on my previous comment, we should be able to match with any arbitary ConstValue.

tusharmath: Actually the implementation isn't right. 

We need to store a `HashMap<ConstValue, String>` so that the incoming value can be mapped to any EnumVariant.

tusharmath: You are right, but we want to be able to map any ConstValue to a String. For I can do the following 

```graphql
enum Chars {
  A @alias(options: ["alpha", 1, {this_can_also_work: true}]),
  B,
  C
}

```

tusharmath: Can use `serde_json::Value` also.

tusharmath: In that case, lets completely drop const_value from the Map, keep it as a `HashMap<String, String>`

tusharmath (approved):

PR #2150, opened 2024-06-09
chore: replace IoId with CallId

ssddOnTop:
ext of #2090

### DIFF
--- a/src/core/counter.rs
+++ b/src/core/counter.rs
@@ -29,9 +29,9 @@ impl Count for Counter {
 }
 
 #[derive(Default, Debug)]
-pub struct MutexCounter(Mutex<usize>);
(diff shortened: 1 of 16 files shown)


src/core/counter.rs
```diff
@@ -1,15 +1,42 @@
...
+
+    fn next(&self) -> Self::Item {
         let curr = self.0.get();
         self.0.set(curr + 1);
         curr
     }
 }
+
+#[derive(Default, Debug)]
+pub struct MutexCounter(Mutex<usize>);
```
tusharmath, line 32:
```suggestion
pub struct AtomicCounter(Mutex<usize>);
```

tusharmath (approved):

PR #2170, opened 2024-06-11
chore(store): drop struct Data

ssddOnTop:
ext of: #2090

### DIFF
--- a/src/core/ir/jit/fixtures/jsonplaceholder-mutation.graphql
+++ b/src/core/ir/jit/fixtures/jsonplaceholder-mutation.graphql
@@ -53,7 +53,7 @@ type User {
   phone: String
   website: String
   address: Address
-  todo: [Todo] @http(path: "/todos?userId={{.value.id}}")
+  todo: [Todo] @http(path: "/todos}}")
 }
 
 type Todo {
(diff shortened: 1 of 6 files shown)


src/core/ir/jit/fixtures/jsonplaceholder-mutation.graphql
```diff
@@ -53,7 +53,7 @@ type User {
   phone: String
   website: String
   address: Address
-  todo: [Todo] @http(path: "/todos?userId={{.value.id}}")
+  todo: [Todo] @http(path: "/todos}}")
```
tusharmath, line 56: this is wrong.

src/core/ir/jit/synth.rs
```diff
@@ -235,21 +267,10 @@ mod tests {
...
-                },
-            ),
             // Insert /user/:id
             (
-                CallId::new(1),
-                Data {
-                    data: Some(OwnedValue::from_str(USER).unwrap()),
-                    extras: Default::default(),
-                },
+                IoId::new(12070278117653709596),
```
tusharmath, line 272: Can't we use the Enum for these Ids?

Issue #2193, opened 2024-06-14
chore: implement cache key for all types


ssddOnTop:
**Issue**
Currently, CacheKey returns `Option<IoId>` because we are unable to generate IoId for IO type JS.

**TODO**
Generate IoId for JS from the ctx value.

tusharmath:
Need to think of an alternative.

PR #2194, opened 2024-06-14
fix: implement cache key for all IO types

ssddOnTop:
fixes: #2193

### DIFF
--- a/src/core/ir/io.rs
+++ b/src/core/ir/io.rs
@@ -1,10 +1,12 @@
 use core::future::Future;
+use std::hash::{Hash, Hasher};
 use std::pin::Pin;
 use std::sync::Arc;
 
 use async_graphql::from_value;
 use async_graphql_value::ConstValue;
 use reqwest::Request;
+use tailcall_hasher::TailcallHasher;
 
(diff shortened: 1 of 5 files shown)


src/core/ir/io.rs
```diff
@@ -167,12 +166,18 @@ impl IO {
...
-    fn cache_key(&self, ctx: &EvaluationContext<'a, Ctx>) -> Option<IoId> {
+    fn cache_key(&self, ctx: &EvaluationContext<'a, Ctx>) -> IoId {
         match self {
             IO::Http { req_template, .. } => req_template.cache_key(ctx),
             IO::Grpc { req_template, .. } => req_template.cache_key(ctx),
             IO::GraphQL { req_template, .. } => req_template.cache_key(ctx),
-            IO::Js { .. } => None,
+            IO::Js { name } => {
+                let mut hasher = TailcallHasher::default();
+                let val = ctx.value().map(|v| v.to_string()).unwrap_or_default();
```
tusharmath, line 176: This hashing technique is going to be very very slow. We are effectively serializing the data twice.

ssddOnTop: but const value don't have Hash implemented

tusharmath:
Need to think of an alternative.

PR #2220, opened 2024-06-17
[stack]: redesign errors - 3

mehulmathur16:


src/core/rest/directive.rs
```diff
@@ -225,6 +227,9 @@ mod tests {
         let query = generate_query_with_directive(&query, DEFAULT_QUERY_PARAM);
         let directive = query_to_directive(&query);
         // Will panic
-        Rest::try_from(&directive).unwrap();
+        match Rest::try_from(&directive) {
+            Ok(_) => panic!("Test should have failed but it succeeded"),
+            Err(e) => panic!("{}", e),
+        }
```
tusharmath, line 233: Why can't we use `unwrap` here?

PR #2234, opened 2024-06-20
fix: types across multiple configs

shashitnak:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/config/config_module.rs
```diff
@@ -21,9 +22,6 @@ use crate::core::valid::{Valid, Validator};
 pub struct ConfigModule {
     pub config: Config,
     pub extensions: Extensions,
-    pub input_types: HashSet<String>,
-    pub output_types: HashSet<String>,
-    pub interface_types: HashSet<String>,
 }
```
meskill, line 25: I'm not sure that's the right way to solve this.

As I understand the problem is that when using links we create `ConfigModule` from initial config and we merge it with provided links. And resolved data like `input_types` is resolved only once for the first config, loosing all of the additional info. 

Ideally, we should not calculate any info until we end the merge process and collect all the types. Or as a simpler solution like this will work

```rust
    async fn ext_links(
        // ...
    ) -> anyhow::Result<ConfigModule> {
        // ...

        let mut result = ConfigModule::from(config_module.config);

        result.extensions = config_module.extensions;

        Ok(result)
    }
```

meskill (changes requested):

src/core/config/config_module.rs
shashitnak: That is correct. The suggested solution is much simpler. I have made these changes and also added a comment explaining why this is being done

src/core/config/config_module.rs
```diff
@@ -119,4 +107,75 @@ impl ConfigModule {
...
+    }
+
+    /// Returns a list of all the arguments in the configuration
+    fn arguments(&self) -> Vec<(&String, &Arg)> {
+        self.types
+            .iter()
+            .flat_map(|(_, type_of)| type_of.fields.iter())
+            .flat_map(|(_, field)| field.args.iter())
+            .collect::<Vec<_>>()
+    }
```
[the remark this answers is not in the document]
shashitnak: removed them since they are not required with the new changes

PR #2259, opened 2024-06-23
chore: JIT executor

tusharmath:
- **refactor: make store polymorphic**
- **lint fixes**
- **chore: update linter version**
- **add exec**


src/core/jit/mod.rs
```diff
@@ -9,3 +10,4 @@ pub use synth::*;
 
 // NOTE: Only used in tests and benchmarks
 pub mod common;
+mod ir;
```
tusharmath, line 13:
```suggestion
pub mod ir;
```

PR #2260, opened 2024-06-23
chore: dropped pub fields from config module

### LINKED ISSUE #2256: refactor: Drop `pub` from all the fields of ConfigModule
**Technical Requirements**
- Derive getters to get access by reference to the internal fields.
- There is only one setter usage for ConfigModule that can be changed to use `config_module::Extension` which has pub fields.

dekkku:
**Summary:**  

**Issue Reference(s):**  
Fixes #2256 
/claim #2256

**Build & Testing:**

**Checklist:**


src/core/config/config_module.rs
```diff
@@ -15,13 +15,22 @@ use crate::core::rest::{EndpointSet, Unchecked};
...
+    interface_types: HashSet<String>,
+}
+
+impl ConfigModule {
+    pub fn config_mut(&mut self) -> &mut Config {
+        &mut self.config
+    }
+    pub fn set_extensions(&mut self, extensions: Extensions) {
+        self.extensions = extensions;
+    }
```
tusharmath, line 33:
```suggestion
    pub fn set_config(self, config: Config) -> Config {
        let c = Self::from(config);
         c.extensions = self.extensions;
    }
    
    pub fn with_extensions(&mut self, extensions: Extensions) {
        self.extensions = extensions;
    }
```

be-hunter: updated.

src/core/config/reader.rs
```diff
@@ -131,16 +129,16 @@ impl ConfigReader {
...
                 }
             }
         }
 
         // Recreating the ConfigModule in order to recompute the values of
         // `input_types`, `output_types` and `interface_types`
-        let mut final_config_module = ConfigModule::from(config_module.config);
-        final_config_module.extensions = config_module.extensions;
+        let mut final_config_module = ConfigModule::from(config_module.config().clone());
+        final_config_module.with_extensions(extensions);
```
tusharmath, line 141: This is just a bit awkward. I think the main problem is that we are modifying the `config` above when there is a `LinkType::Config` and then we use `merge_right` on `ConfigModule`. 

I think a better approach would be to pull out the config from config_module modify it wherever required and then recreate the ConfigModule. We should not use the `merge_right` feature in ConfigModule, it's derived and is logically incorrect.

src/core/config/reader.rs
```diff
@@ -131,16 +129,16 @@ impl ConfigReader {
...
         Ok(final_config_module)
```
tusharmath, line 142:
```suggestion
        Ok(ConfigModule::from(config).with_extensions(extensions))
```

This is a much better usage. Notice that the `config` passed to the `from` method. It is a config that is extracted in begining of the function and then modified whenever we hit a `LinkType::Config`.

Feel free to connect on discord if it's not clear.

tusharmath (approved):

Issue #2295, opened 2024-06-27
refactor: drop `to_sdl` method on `config_module`


tusharmath:
Use the `to_sdl` method in `Config` wherever required.

PR #2296, opened 2024-06-27
fix: drop `to_sdl` from `ConfigModule`

ssddOnTop:
**Issue Reference(s):**  
/claim #2295

### DIFF
--- a/src/core/config/into_document.rs
+++ b/src/core/config/into_document.rs
@@ -2,7 +2,7 @@ use async_graphql::parser::types::*;
 use async_graphql::{Pos, Positioned};
 use async_graphql_value::{ConstValue, Name};
 
-use super::{Config};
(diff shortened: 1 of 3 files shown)


src/core/config/into_document.rs
```diff
@@ -59,7 +59,7 @@ fn config_document(config: &ConfigModule) -> ServiceDocument {
     };
     definitions.push(TypeSystemDefinition::Schema(pos(schema_definition)));
     for (type_name, type_def) in config.types.iter() {
-        let kind = if config.interface_types.contains(type_name) {
+        let kind = if config.interface_types().contains(type_name) {
```
tusharmath, line 62: Move this to a variable outside the for loop.

tusharmath (approved):

PR #2302, opened 2024-06-28
feat: allows users to specify cache as percentage of memory(#2035)

### LINKED ISSUE #2035: feat: allow users to specify the http cache size in terms of total bytes rather than number of entries.

sirish123:
**Summary:**  
httpCache Parameter has been changed to accommodate the cache as a percentage of memory
`httpCache:{
enable: bool,
size: u64
}` 

**Issue Reference(s):**  
Fixes #2035 

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/core/config/upstream.rs
+++ b/src/core/config/upstream.rs
@@ -2,6 +2,7 @@ use std::collections::BTreeSet;
 
 use derive_setters::Setters;
 use serde::{Deserialize, Serialize};
+use tailcall_macros::{DirectiveDefinition, InputDefinition};
 
 use crate::core::is_default;
 use crate::core::macros::MergeRight;
@@ -45,7 +46,10 @@ pub struct Proxy {
     Default,
     schemars::JsonSchema,
     MergeRight,
+    DirectiveDefinition,
+    InputDefinition,
 )]
+#[directive_definition(locations = "Schema")]
 #[serde(deny_unknown_fields)]
 #[serde(rename_all = "camelCase", default)]
 /// The `upstream` directive allows you to control various aspects of the
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5230,6 +5230,7 @@ dependencies = [
  "tailcall-macros",
  "tailcall-prettier",
  "tailcall-tracker",
+ "tailcall-typedefs-common",
  "tailcall-version",
  "temp-env",
  "tempfile",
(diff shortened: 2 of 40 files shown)


sirish123:
@tusharmath  docs updated at https://github.com/tailcallhq/tailcallhq.github.io/pull/295 and conflicts resolved.

src/core/config/upstream.rs
```diff
@@ -28,7 +28,13 @@ impl Default for Batch {
         }
     }
 }
-
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug, schemars::JsonSchema, MergeRight)]
+#[serde(rename_all = "camelCase", default)]
+#[derive(Default)]
+pub struct HttpCache {
+    pub enable: bool,
+    pub size: u64,
```
ssddOnTop, line 36: now that we are taking `size` we do not need `enable`.  If user  specifies size, it's obvious to turn httpCache on.

and secondly, you should skip serialization if default

src/core/config/upstream.rs
```diff
@@ -28,7 +28,13 @@ impl Default for Batch {
         }
     }
 }
-
+#[derive(Serialize, Deserialize, PartialEq, Eq, Clone, Debug, schemars::JsonSchema, MergeRight)]
+#[serde(rename_all = "camelCase", default)]
```
ssddOnTop, line 32: should not use default here, we can do that wherever we use this struct as a field

src/core/config/upstream.rs
sirish123: @ssddOnTop _**enable**_ was suggested by @tusharmath  if user wants _**size**_ to represent percentage rather than number of entries example below.

```graphql
schema
  @server(port: 4000)
  @upstream(
    baseURL: "https://api.example.com"
    # highlight-start
    httpCache: {size: 42} # Fixed number of entries
    # OR
    httpCache: {enable: true, size: 2} # 2% of available memory (i.e. RAM)
    # highlight-end
  ) {
  query: Query
}
```

src/core/config/upstream.rs
sirish123: @ssddOnTop On removing **default** doesn't the user have to specify **enable** as now its a required field, it makes sense if we remove **_enable_** though than only **_size_** is the attribute present and it should be mentioned explicitly.
image

PR #2335, opened 2024-07-02
fix: query plan arguments

shashitnak:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/jit/exec.rs
```diff
@@ -70,6 +71,28 @@ where
     async fn init(&mut self) {
         join_all(self.plan.as_children().iter().map(|field| async {
             let ctx = Context::new(&self.request);
+            let mut arg_map = indexmap::IndexMap::new();
+            for arg in field.args.iter() {
+                let name = async_graphql::Name::new(&arg.name);
+                let value: Option<anyhow::Result<ConstValue>> = arg
```
tusharmath, line 77: `exec` can't be bound to `ConstValue` or any specific value. Use the type param `Value` here.

src/core/jit/exec.rs
```diff
@@ -70,6 +71,28 @@ where
     async fn init(&mut self) {
         join_all(self.plan.as_children().iter().map(|field| async {
             let ctx = Context::new(&self.request);
+            let mut arg_map = indexmap::IndexMap::new();
+            for arg in field.args.iter() {
```
tusharmath, line 75: Field will only have arg names but not the actual args. Actual arg information is in the request.

src/core/jit/context.rs
```diff
@@ -8,21 +8,29 @@ use crate::core::ir::ResolverContextLike;
 pub struct Context<'a, Input, Output> {
     request: &'a Request<Input>,
     value: Option<&'a Output>,
+    args: Option<&'a indexmap::IndexMap<async_graphql::Name, async_graphql::Value>>,
```
tusharmath, line 11:
```suggestion
    args: Option<&'a indexmap::IndexMap<&'str, Input>>,
```

src/core/jit/exec.rs
shashitnak: since in `Arg` type field `value` is `async_graphql_value::Value` and `default_value` is `async_graphql_value::ConstValue` so the value of any argument will also have to be `ConstValue`. If we want this to be generic I think we also will have to update `Arg` type to use generic types. Should I update that as well?

src/core/jit/exec.rs
shashitnak: If I understand correctly, the way we are parsing the query, we will find the value of arguments in the `Arg` type when query is passed as `query {user(id: 1) {id}}` and when it is passed as `query user($id: Int!) {user(id: $id) { id }}` then we will find the value in variables

src/core/jit/synth/synth_borrow.rs
```diff
@@ -4,13 +4,13 @@ use crate::core::jit::model::{Children, Field};
...
-pub struct SynthBorrow<'a> {
-    selection: Vec<Field<Children>>,
+pub struct SynthBorrow<'a, ParsedValue: Clone, Input: Clone> {
+    selection: Vec<Field<Children<ParsedValue, Input>, ParsedValue, Input>>,
     store: Store<Value<'a>>,
 }
 
-impl<'a> SynthBorrow<'a> {
-    pub fn new(plan: ExecutionPlan, store: Store<Value<'a>>) -> Self {
+impl<'a, ParsedValue: Clone, Input: Clone> SynthBorrow<'a, ParsedValue, Input> {
```
tusharmath, line 12: SynthBorrow should not be parameterized.

src/core/generator/snapshots/tailcall__core__generator__generator__test__should_generate_combined_config.snap
```diff
@@ -89,11 +89,12 @@ type Query {
...
   posts: [Post] @http(path: "/posts?_limit=11")
-  user(id: Int!): User @http(path: "/users/{{.args.id}}")
-  users: [User] @http(path: "/users")
+  user(id: Int!): User2 @http(path: "/users/{{.args.id}}")
+  user2(id: Int = 1): User2 @http(path: "/users/{{.args.id}}")
+  users: [User2] @http(path: "/users")
 }
 
-type User {
+type User2 {
```
tusharmath, line 97: Why is this renamed?

src/core/jit/context.rs
```diff
@@ -8,21 +10,50 @@ use crate::core::ir::ResolverContextLike;
...
+    {
+        let mut map = indexmap::IndexMap::new();
+        for (key, value) in args.iter() {
+            map.insert(async_graphql::Name::new(key), value.clone().into());
+        }
+        Self {
+            request: self.request,
+            value: self.value,
+            args: Some(args),
+            arg_const_value: Some(Arc::new(map)),
```
meskill, line 55: why store `args` and `arg_const_value` separately and why use only the second one later?

meskill (changes requested):

src/core/jit/model.rs
```diff
@@ -1,14 +1,32 @@
...
     pub name: String,
     pub type_of: crate::core::blueprint::Type,
-    pub value: Option<async_graphql_value::Value>,
-    pub default_value: Option<async_graphql_value::ConstValue>,
+    pub value: Option<Input>,
+    pub default_value: Option<Input>,
+}
+
+impl<Input> Arg<Input> {
+    pub fn map_value<Output, Error>(
```
tusharmath, line 18:
```suggestion
    pub fn try_map<Output, Error>(
```

src/core/jit/model.rs
```diff
@@ -45,31 +85,53 @@ impl FieldId {
...
 
-const EMPTY_VEC: &Vec<Field<Nested>> = &Vec::new();
-impl Field<Nested> {
-    pub fn nested(&self) -> &Vec<Field<Nested>> {
-        match &self.extensions {
-            Some(Nested(children)) => children,
-            _ => EMPTY_VEC,
-        }
+impl<Extensions, Input> Field<Extensions, Input> {
+    pub fn map_args<Output, Error>(
```
tusharmath, line 98:
```suggestion
    pub fn try_map<Output, Error>(
```

tests/jit_spec.rs
```diff
@@ -105,8 +124,72 @@ mod tests {
...
+        match executor.run(request).await {
+            Ok(_) => panic!("Should fail with unresolved variable"),
+            Err(err) => assert_eq!(
+                err.to_string(),
+                "ResolveInputError: Variable `id` is not defined"
+            ),
+        };
+
+        let request = Request::new(query);
+        let request = request.variables(Variables::from_iter([("id".into(), ConstValue::from(1))]));
```
tusharmath, line 179:
```suggestion
        let request = request.variables([("id".into(), ConstValue::from(1))]);
```

No need exposing `Variables` publicly just for tests.

src/core/jit/input_resolver.rs
```diff
@@ -0,0 +1,68 @@
...
+    plan: ExecutionPlan<Input>,
+}
+
+impl<Input> InputResolver<Input> {
+    pub fn new(plan: ExecutionPlan<Input>) -> Self {
+        Self { plan }
+    }
+}
+
+impl<Input, Output> InputResolver<Input>
```
tusharmath, line 45: I don't think is really needed. We can implement this behaviour by adding a `try_map` in `ExecutionPlan`.

src/core/jit/common/json_placeholder.rs
```diff
@@ -82,6 +84,10 @@ impl JsonPlaceholder {
             store
         });
 
+        let input_resolver = InputResolver::new(plan);
+
+        let plan = input_resolver.resolve_input(&Variables::default()).unwrap();
+
         Synth::new(plan, store)
```
tusharmath, line 91: May be `ConstSynth` could perform this transformation of the plan using `plan.try_map` while synthesizing?

src/core/jit/exec.rs
```diff
@@ -70,14 +75,31 @@ where
...
+                    .value
+                    .clone()
+                    // TODO: default value resolution should happen in the InputResolver
+                    .or_else(|| arg.default_value.clone());
+
+                if let Some(value) = value {
+                    arg_map.insert(name, value);
+                } else if !arg.type_of.is_nullable() {
+                    // TODO: throw error here
+                    todo!()
```
tusharmath, line 91: Can handle error here instead of `todo`

src/core/jit/model.rs
tusharmath: Also do we need need to map args or just the value inside it?

src/core/jit/model.rs
meskill: renamed

src/core/jit/model.rs
meskill: yes, renamed and changed to work only with values

tests/jit_spec.rs
meskill: applied the suggestion

src/core/jit/input_resolver.rs
meskill: when we discussed this you suggested to implement value resolution as separate step e.g. to have Field<Input> -> Field<ResolvedInput> conversion

src/core/jit/exec.rs
meskill: not that easy here since the error is generic in exec and we can't actually return any concrete error here.

I think it should be better resolved after changes in error handling for jit

src/core/jit/model.rs
```diff
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::fmt::{Debug, Formatter};
 
+use async_graphql_value::ConstValue;
```
tusharmath, line 4: Model shouldn't include concrete types. Keep the model parametric for maximum flexibility.

meskill: not my changes, the usage of ConstValue is came from skip/include pr. I just replaced it with import

src/core/jit/builder.rs
```diff
@@ -51,6 +54,7 @@ pub struct Builder {
     pub document: ExecutableDocument,
 }
 
+// TODO: make generic over Value (Input) type
```
tusharmath, line 57: Not sure if we can make this happen because we depend on async_graphql for parsing the request and just abstracting over the input type won't be enough.

src/core/jit/model.rs
```diff
@@ -33,12 +34,27 @@ impl<V> FromIterator<(String, V)> for Variables<V> {
...
 
 #[derive(Debug, Clone)]
-pub struct Arg {
+pub struct Arg<Input> {
     pub id: ArgId,
     pub name: String,
     pub type_of: crate::core::blueprint::Type,
-    pub value: Option<async_graphql_value::Value>,
-    pub default_value: Option<async_graphql_value::ConstValue>,
+    pub value: Option<Input>,
```
tusharmath, line 41: Why is `value` needed in the plan?

meskill: this is the value passed from the request
image

it's a part of the request and thus part of the plan

src/core/jit/model.rs
```diff
@@ -33,12 +33,27 @@ impl<V> FromIterator<(String, V)> for Variables<V> {
...
+    pub default_value: Option<Input>,
```
tusharmath, line 41: So `default_value` is defined in the schema, coming from blueprint and value is specific to a query, right?

meskill: right

tusharmath (approved):

PR #2345, opened 2024-07-04
feat: New transformer to flatten out types with single field

### LINKED ISSUE #2343: transformer to flatten out types with single field
Write a `config/transformer` that unwraps types with one field.

**FROM**
```graphql
type Query {
  foo: Foo
}

# Type with only one field
type Foo { 
  bar: Bar
}

# Type with only one field
type Bar {
  a: Int
}
```

**TO**

```graphql
type Query {
  foo: Int
}
```

This can be achieved by using the `addField` and `omit` operators. 

**Technical Requirements**
- Should be configurable via the `Preset` settings.
- Should be enabled by default in config generation.
- Should be added before BetterTypeName but after TreeShaking transformers.
- Add integration tests in `tests/execution`

**IMPORTANT**
- The "Before" is the server schema and the "After" is the expected client schema.
- Link:
	- https://tailcall.run/docs/tailcall-dsl-graphql-custom-directives/#addfield-directive 
	- https://tailcall.run/docs/tailcall-dsl-graphql-custom-directives/#modify-directive
	- https://tailcall.run/docs/contributors/integration-testing/

LeoLeSommer:
**Summary:**  

(That's my first contribution ever on an open source project, I hope it will not be too bad)

I took some liberty on implementation, I'll comment the PR to explain my choices.

/claim #2343

Write a config/transformer that unwraps types with one field.

FROM:

```graphql
type Query {
  foo: Foo
}

# Type with only one field
type Foo { 
  bar: Bar
}

# Type with only one field
type Bar {
  a: Int
}
```

TO:

```graphql
type Query {
  foo: Int
}
```

**Issue Reference(s):**  
Fixes #2343

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/core/config/transformer/flatten_single_types.rs
+++ b/src/core/config/transformer/flatten_single_types.rs
@@ -1,0 +1,176 @@
+use std::collections::BTreeMap;
+
+use crate::core::config::{Config, Field, Type};
+use crate::core::transform::{self, Transform};
+use crate::core::valid::Valid;
+
+/// Flatten single types
+/// If a type has only one field, it is considered a single type.
+/// If a field references a single type, it will be flattened.
+#[derive(Default)]
+pub struct FlattenSingleTypes;
+
+impl Transform for FlattenSingleTypes {
+    type Value = Config;
+    type Error = String;
+
+    fn transform(&self, mut config: Config) -> Valid<Config, String> {
+        let single_types = get_single_types(&config);
+        let root_types = get_root_types(&config);
+        let mut still_referenced_single_types = Vec::<String>::new();
+
+        config.types.iter_mut().for_each(|(_, ty)| {
+            let fields_referencing_single_type = ty
+                .fields
+                .iter_mut()
+                .filter(|(_, field)| single_types.contains_key(&field.type_of));
+
+            // Skip fields with resolver and prevent array of array
+            let (skipped_fields, fields_to_flatten): (Vec<_>, Vec<_>) =
+                fields_referencing_single_type.partition(|(_, field)| {
+                    field.has_resolver() || (field.list && single_types[&field.type_of].list)
+                });
+
+            // Keep track of single types that are still referenced
+            still_referenced_single_types.extend(
+                skipped_fields
+                    .iter()
+                    .map(|(_, field)| field.type_of.clone()),
+            );
+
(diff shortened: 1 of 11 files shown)


src/core/config/transformer/flatten_single_types.rs
```diff
@@ -0,0 +1,176 @@
...
+        let single_types = get_single_types(&config);
+        let root_types = get_root_types(&config);
+        let mut still_referenced_single_types = Vec::<String>::new();
+
+        config.types.iter_mut().for_each(|(_, ty)| {
+            let fields_referencing_single_type = ty
+                .fields
+                .iter_mut()
+                .filter(|(_, field)| single_types.contains_key(&field.type_of));
+
```
LeoLeSommer, line 27: I did skip the flatten when there is a resolver cause i was anticipating some side effects.

src/core/config/transformer/preset.rs
```diff
@@ -26,6 +27,7 @@ impl Transform for Preset {
             .pipe(super::TreeShake.when(self.tree_shake))
             .pipe(super::TypeMerger::new(self.merge_type))
             .pipe(super::ImproveTypeNames.when(self.use_better_names))
+            .pipe(super::FlattenSingleTypes.when(self.flatten_single_types))
```
LeoLeSommer, line 30: Issue wa asking to put the flatten of single types before ImproveTypeNames but i noticed some strange behaviour when it is before so i placed it after.

src/core/config/transformer/preset.rs
```diff
@@ -36,8 +38,9 @@ impl Default for Preset {
         Self {
             merge_type: 1.0,
             consolidate_url: 0.5,
-            use_better_names: true,
             tree_shake: true,
+            flatten_single_types: true,
```
tusharmath, line 42: I don't think we should enable this by default.

tusharmath:
@LeoLeSommer Appreciate your time and effort but I don't think you understood the problem. Feel free to follow up on discord, but in general you can't delete stuff from the config. We need to use `modify` and `addField` directives to make sure the final configuration looks like what we intend it to look like. `addField` and `modify` will handle the resolvers internally to make sure the generated configuration actually works. Currently it seems like you are are just creating new type with some fields that might resolve with values that don't match the actual type.

PR #2357, opened 2024-07-05
fix(jit): implement `skip` and `include`

ssddOnTop:


src/core/jit/builder.rs
```diff
@@ -9,26 +9,67 @@ use async_graphql::Positioned;
...
     pub document: ExecutableDocument,
+    variables: Variables<async_graphql_value::ConstValue>,
 }
 
 impl Builder {
-    pub fn new(blueprint: &Blueprint, document: ExecutableDocument) -> Self {
+    pub fn new(
+        blueprint: &Blueprint,
+        document: ExecutableDocument,
+        variables: Variables<async_graphql_value::ConstValue>,
```
tusharmath, line 28: Variables will change with each request but the plan is created only once. Ideally, this information should be used dynamically at the time of execution and not planning - UNLESS the it's always false or always true. We don't need to worry about the "always" cases right now though.

src/core/jit/model.rs
```diff
@@ -50,10 +52,39 @@ pub struct Field<Extensions> {
     pub name: String,
     pub ir: Option<IR>,
     pub type_of: crate::core::blueprint::Type,
+    pub include: Option<Include>,
```
tusharmath, line 55:
```suggestion
    pub ignore: Ignore,
```

```rust
enum Ignore {
    Never, // Default
    Always, // If a literal is used.    
    When(String)  // Represents when the variable name is true.
}
```

src/core/jit/request.rs
```diff
@@ -6,28 +6,43 @@ use serde::Deserialize;
...
+    pub fn new() -> Self {
+        Self(HashMap::new())
+    }
+    pub fn get(&self, key: &str) -> Option<&Value> {
+        self.0.get(key)
+    }
+    pub fn insert(&mut self, key: String, value: Value) {
+        self.0.insert(key, value);
+    }
+}
```
tusharmath, line 34: Move to `model.rs`

src/core/jit/request.rs
```diff
@@ -6,28 +6,43 @@ use serde::Deserialize;
...
             query: value.query,
             operation_name: value.operation_name,
             variables: match value.variables.into_value() {
-                async_graphql_value::ConstValue::Object(val) => {
-                    HashMap::from_iter(val.into_iter().map(|(k, v)| (k.to_string(), v)))
-                }
-                _ => HashMap::new(),
+                async_graphql_value::ConstValue::Object(val) => Variables(HashMap::from_iter(
+                    val.into_iter().map(|(k, v)| (k.to_string(), v)),
+                )),
```
tusharmath, line 44: Should have helpful constructors to create Variables from iters so that we don't need to write all this here.

src/core/jit/request.rs
```diff
@@ -6,28 +6,43 @@ use serde::Deserialize;
...
+                _ => Variables(HashMap::new()),
```
tusharmath, line 45:
```suggestion
                _ => Variables::default(),
```

src/core/jit/request.rs
```diff
@@ -47,7 +62,7 @@ impl<A> Request<A> {
         Self {
             query: query.to_string(),
             operation_name: None,
-            variables: HashMap::new(),
+            variables: Variables::new(),
```
tusharmath, line 65:
```suggestion
            variables: Variables::default(),
```

src/core/jit/model.rs
ssddOnTop: won't work.. we need 2 information, i.e. if it is include or skip directive and the Value associated with it.

tusharmath:
```rust
enum Ignore {
    Never, // Default
    Always, // If a literal is used.    
    WhenTrue(Variable)  // Ignore when the variable is true.
    WhenFalse(Variable)  // Ignore when the variable is false.
}

struct Variable(String); 
```

src/core/jit/model.rs
```diff
@@ -1,6 +1,38 @@
+use std::collections::HashMap;
 use std::fmt::{Debug, Formatter};
 
+use async_graphql_value::Value;
```
tusharmath, line 4: don't refer to concrete type in the model.

src/core/jit/model.rs
```diff
@@ -50,9 +82,77 @@ pub struct Field<Extensions> {
...
+    #[default]
+    Never, // Default
+    Always,              // If a literal is used.
+    SkipIf(Variable),    // when @skip is used
+    IncludeIf(Variable), // when @include is used
+}
+
+impl Ignore {
+    #[inline]
+    pub fn new(include: bool, value: &Value) -> Self {
```
tusharmath, line 109: inline this function to the builder where it's called.

src/core/jit/model.rs
```diff
@@ -50,9 +82,77 @@ pub struct Field<Extensions> {
...
+                Value::String(st) => Ignore::SkipIf(Variable::new(st.as_str())),
+                Value::Boolean(bool) => {
+                    if *bool {
+                        Ignore::Always
+                    } else {
+                        Ignore::Never
+                    }
+                }
+                _ => Ignore::default(),
+            }
```
tusharmath, line 135: Lot of duplicate code. This pattern resembles the XOR operation.

src/core/jit/model.rs
```diff
@@ -50,9 +82,77 @@ pub struct Field<Extensions> {
...
+                    } else {
+                        Ignore::Always
+                    }
+                }
+                _ => Ignore::default(),
+            }
+        } else {
+            match value {
+                Value::Variable(var) => Ignore::SkipIf(Variable::new(var.as_str())),
+                Value::String(st) => Ignore::SkipIf(Variable::new(st.as_str())),
```
tusharmath, line 126: We don't need to support string.

src/core/jit/model.rs
```diff
@@ -50,9 +82,77 @@ pub struct Field<Extensions> {
...
+                        Ignore::Never
+                    }
+                }
+                _ => Ignore::default(),
+            }
+        }
+    }
+
+    #[inline(always)]
+    pub fn ignore<Value: JsonLike>(&self, variables: &Variables<Value>) -> bool {
```
tusharmath, line 140:
```suggestion
    pub fn check<Value: JsonLike>(&self, variables: &Variables<Value>) -> bool {
```

src/core/jit/builder.rs
```diff
@@ -29,7 +29,28 @@ impl Builder {
...
+    fn include(
+        &self,
+        directives: &[Positioned<async_graphql::parser::types::Directive>],
+    ) -> Ignore {
+        for directive in directives {
+            let include = match &*directive.node.name.node {
+                "skip" => false,
+                "include" => true,
+                _ => continue,
+            };
```
tusharmath, line 42: Add a unit test for 

> Neither @skip nor @include has precedence over the other. In the case that both the @skip and @include directives are provided on the same field or fragment, it must be queried only if the @skip condition is false and the @include condition is true. Stated conversely, the field or fragment must not be queried if either the @skip condition is true or the @include condition is false.

src/core/jit/model.rs
```diff
@@ -50,9 +81,46 @@ pub struct Field<Extensions> {
     pub name: String,
     pub ir: Option<IR>,
     pub type_of: crate::core::blueprint::Type,
+    pub ignore: Ignore,
```
tusharmath, line 84:
```suggestion
    pub skip_when: Option<Variable>,
    pub include_when: Option<Variable>,
```

tusharmath (approved):

PR #2361, opened 2024-07-06
chore(JsonLike): introduce lifetime

ssddOnTop:


src/core/json/json_like.rs
```diff
@@ -15,55 +15,69 @@ pub trait JsonLike {
...
-    fn as_option_ok(&self) -> Result<Option<&Self::Output>, &str>;
-    fn get_path<T: AsRef<str>>(&self, path: &[T]) -> Option<&Self::Output>;
-    fn get_key(&self, path: &str) -> Option<&Self::Output>;
-    fn new(value: &Self::Output) -> &Self;
-    fn group_by<'a>(&'a self, path: &'a [String]) -> HashMap<String, Vec<&'a Self::Output>>;
+    fn as_array_ok(&'json self) -> Result<&'json Vec<Self::Output>, &str>;
+    fn as_object_ok(&'json self) -> Result<&'json Self::Obj, &str>;
+    fn as_str_ok(&'json self) -> Result<&'json str, &str>;
+    fn as_string_ok(&'json self) -> Result<&'json str, &str>;
+    fn as_i64_ok(&'json self) -> Result<i64, &str>;
```
meskill, line 22: what's the point of this lifetime here?

not, at least, it puts stronger boundaries than it was before, see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4113f351fc618e6a287634376447bc15

ssddOnTop:
this PR is not needed

Issue #2381, opened 2024-07-09
wrong type for the body parameter of the grpc directive


nekinie:
### Prerequisites

- [x] I have checked the existing issues to ensure this is not a duplicate.
- [x] I am running the latest version of Tailcall.

### Describe the bug
 
The type for the body parameter of the GRPC directive is string, but the comment around its decleration indicates that static objects should work. I beleve the type should be JSON to support flat objects in the same way that the args parameter for the call directive.

https://github.com/tailcallhq/tailcall/blob/main/generated/.tailcallrc.graphql#L112

tusharmath:
Thanks @nekinie you are right, I think this is a quick fix. BTW have you checkout our auto-generation guide — https://tailcall.run/docs/graphql-configuration-generation-with-tailcall/

It should reduce your onboarding time drastically!

tusharmath:
fixed in https://github.com/tailcallhq/tailcall/pull/2383

nekinie:
I gave this a test today and everything looks great, thank you :)

PR #2383, opened 2024-07-09
fix(grpc): allow users to pass value along with mustache template

ssddOnTop:
/claim #2381

### DIFF
--- a/tests/execution/grpc-json.md
+++ b/tests/execution/grpc-json.md
@@ -70,6 +70,7 @@ type News {
   response:
     status: 200
     textBody: \0\0\0\0#\x08\x02\x12\x06Note 2\x1a\tContent 2\"\x0cPost image 2
+  expectedHits: 3
 ```
 
 ```yml @test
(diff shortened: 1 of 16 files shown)


tests/execution/grpc-json.md
```diff
@@ -0,0 +1,73 @@
...
+```graphql @config
+schema
+  @server(port: 8000)
+  @upstream(baseURL: "http://localhost:50051")
+  @link(id: "news", src: "news.proto", type: Protobuf) {
+  query: Query
+}
+
+type Query {
+  newsById: News! @grpc(method: "news.NewsService.GetNews", body: {id: 2})
```
tusharmath, line 48:
```suggestion
  newsById: News! @grpc(method: "news.NewsService.GetNews", body: {id: 2})
  newsById(id: Int): News! @grpc(method: "news.NewsService.GetNews", body: {id: "{{.args.id}}"})
```

The above should also work.

ssddOnTop: Both works

tusharmath (approved):

Issue #2389, opened 2024-07-09
use `RWLock` in `dedupe`


tusharmath:
Move from mutex to rw_lock.

tusharmath:
We don't see any practical improvements in performance so would be closing this issue.

PR #2390, opened 2024-07-09
chore: use `RWLock` in `dedupe`

ssddOnTop:
/claim #2389


tusharmath:
Can you raise PR WRT this one with `dedupe: true` and change the query to `query {greet}`. I want to see how it impacts the performance of the server.

tusharmath:
LEt's close this if there is not real performance improvement.

PR #2392, opened 2024-07-10
chore: impl Linters

### LINKED ISSUE #1322: Validate schema using linters

ssddOnTop:
**Issue Reference(s):**  
Fixes #1322

### DIFF
--- a/src/core/config/transformer/preset.rs
+++ b/src/core/config/transformer/preset.rs
@@ -11,6 +11,7 @@ pub struct Preset {
     consolidate_url: f32,
     tree_shake: bool,
     use_better_names: bool,
+    linter: bool,
 }
 
 impl Transform for Preset {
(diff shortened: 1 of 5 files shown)


src/core/config/transformer/preset.rs
```diff
@@ -38,6 +40,7 @@ impl Default for Preset {
             consolidate_url: 0.5,
             use_better_names: true,
             tree_shake: true,
+            linter: false,
```
tusharmath, line 43:
```json
{
  "preset": {
    "nomenclature": {
      "types": "Pascal",
      "enum": "Pascal",
      "interface": "Pascal",
      "fields": "Camel",
      "enumValues": "ScreamingSnake"
    }
  }
}
```
Use this setting to display lint errors.

PR #2403, opened 2024-07-10
fix(generator): improve path variable detection logic in config generator

### LINKED ISSUE #2347: bug: improve path variable detection logic in config generator.
## Present Behaviour

Currently, the path variable detection logic is simplistic. It checks if the URL has a number between forward slashes, which works correctly in the following scenarios:

For URL `/api/v1/users/12` and `/api/v1/users/12/comments`  is correctly converted to 
```graphql
@http(path: "/api/v1/users/{{.args.p1}}")
@http(path: "/api/v1/users/{{.args.p1}}/comments")
```
However, the logic fails in the following scenarios where the path variable is a string or contains non-numeric/mixed characters:

For URL `/v1/albums/4a` and `URL /v1/albums/wpa` it is incorrectly converted to 
```graphql
album(p1: Int!): Album @http(path: "/v1/albums/{{.args.p1}}a")
album(p1: Int!): Album @http(path: "/v1/albums/wpa")
```
## Expected Behaviour

The path variable detection logic should be enhanced to detect path variables in all scenarios, including numbers, strings, or mixed characters, and generate a valid @http directive:

For URL `/v1/albums/4a`and ` /v1/albums/wpa`, it should be converted to 
```graphql
album(p1: String!): Album @http(path: "/v1/albums/{{.args.p1}}").
album(p1: String!): Album @http(path: "/v1/albums/{{.args.p1}}").
```

Technical Requirements:
- update the path variable detection logic to handle cases like above.
- update the test cases if necessary.
- add integration test that cover's this case.

bnchi:
**Summary:**  

- Adds improvements to the way of how variables are being detected in the generator by allowing the user to add an optional route deceleration 
- In case the user didn't specify the route config a best effort strategy is applied and the way of how it's being done is documented in the code  

**Issue Reference(s):**  
Fixes #2347

**Build & Testing:**

**Checklist:**


bnchi:
read for review

src/core/generator/json/http_directive_generator.rs
```diff
@@ -108,13 +168,27 @@ impl<'a> HttpDirectiveGenerator<'a> {
 
         self.http
     }
+
+    fn deterime_arg_type_from_route_segment(segment: &str) -> String {
```
tusharmath, line 172:
```suggestion
    fn determine_arg_type_from_route_segment(segment: &str) -> String {
```

src/core/generator/json/http_directive_generator.rs
```diff
@@ -169,4 +243,65 @@ mod test {
...
+        expected_field.args.insert(
+            "p2".to_string(),
+            Arg {
+                type_of: "Int".to_string(),
+                required: true,
+                ..Default::default()
+            },
+        );
+
+        assert_eq!(actual_field, expected_field);
```
tusharmath, line 305: Simplify tests by adding helper functions. Checkout our guide here - https://tailcall.run/docs/contributors/testing/#unit-tests

src/core/generator/json/http_directive_generator.rs
```diff
@@ -50,36 +50,96 @@ impl<'a> UrlUtility<'a> {
 
 pub struct HttpDirectiveGenerator<'a> {
     url: &'a Url,
+    route: Option<&'a String>,
```
tusharmath, line 53:
```suggestion
    route: Option<&'a Rest>,
```

We have already implemented a lot of machinery for `Rest` we should use that. 
Side Note: May be rename `Rest` to `Route` in a separate PR.

tusharmath:
Moving to draft to reduce noise and improve CI efficiency. Once you are ready just mark it as "ready to review". Feel free to give a shoutout on the #contributors channel on Discord if you want immediate attention.

src/core/generator/json/http_directive_generator.rs
bnchi: The implementation here is a bit different than Rest, from my understanding Rest builds a type map of user defined types. 

[](https://tailcall.run/docs/tailcall-dsl-graphql-custom-directives/#rest-directive)

where $id is already known to be an _Int!_.

In here we don't really know the exact type of the segment unless we try to infer it, and when we try to infer we're already at a point where we can just build the _Field_ if that make sense. 

We can however let the user explicitly define the type of $id and the configuration will change a little bit to be like : 

```json
[
      {
        "curl":{
            "src":"https://jsonplaceholder.typicode.com/posts/1",
            "fieldName":"posts",
            "rest": {
              "route": "/posts/$id",
               "variables": {
                 "$id": "Int!"
               }
            }
        }
      }
  ],
```

originally it was like this(which is a lot more simple) : 

```
[
      {
        "curl":{
            "src":"https://jsonplaceholder.typicode.com/posts/1",
            "fieldName":"posts",
            "path": "/posts/$id" // optional new configuration to describe the path
        }
      }
  ],
```

tusharmath:
I think you have raised a valid point about why it's not Rest. And if we go this route the user needs to specify so much more information that he might as well just write the config by hand.

Closing the PR for now. Thank you for spending time on this 🙏 

/tip $50

PR #2424, opened 2024-07-13
refactor: impl is query method

laststylebender14:
**Summary:**  
- renamed ExecutionPlan to OperationPlan.
- GraphQL request can only hold single operation per request. i.e either Query or Mutation but not both simultaneously.

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/jit/context.rs
```diff
@@ -8,21 +8,34 @@ use crate::core::ir::ResolverContextLike;
...
+            value: self.value,
+            is_query: self.is_query,
+        }
     }
 }
 
 impl<'a, Input, Output> Context<'a, Input, Output> {
     pub fn new(request: &'a Request<Input>) -> Self {
-        Self { request, value: None }
+        Self { request, value: None, is_query: false }
```
meskill, line 26: I suggest to pass `is_query` as arguments to `new` instead of separate `with_is_query` since it's significant part of the processed query and it'll be harder to forget about it

PR #2437, opened 2024-07-15
chore: parameterize Synth

ssddOnTop:
same as: #2435 
but it uses `jit::Result<Value>`

Fixes: #2463
/claim #2463

PR #2442, opened 2024-07-15
chore: introduce `try_map` in `OperationPlan`

ssddOnTop:
ext of: #2437


src/core/jit/try_map.rs
```diff
@@ -0,0 +1,137 @@
+use crate::core::jit::model::{Arg, Field, Flat, Nested, OperationPlan};
+
+pub trait TryMap<Input, Value, Output, Error> {
+    fn try_map(self, map: &impl Fn(Input) -> Result<Value, Error>) -> Result<Output, Error>;
+}
```
tusharmath, line 5: Drop the trait. Use a trait only if there is a type-bound needed somewhere.

src/core/jit/try_map.rs
```diff
@@ -0,0 +1,137 @@
...
+            ir: self.ir,
+            type_of: self.type_of,
+            extensions: self.extensions,
+            skip: self.skip,
+            include: self.include,
+            args: self
+                .args
+                .into_iter()
+                .map(|arg| arg.try_map(map))
+                .collect::<Result<_, _>>()?,
```
tusharmath, line 70: Use `self.args` setter instead.

ssddOnTop: how?
will have to clone args to do the same thing

tusharmath (approved):

PR #2444, opened 2024-07-16
chore: impl field method on context

laststylebender14:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/ir/resolver_context_like.rs
```diff
@@ -68,3 +69,87 @@ impl<'a> ResolverContextLike for ResolverContext<'a> {
         self.inner.ctx.add_error(error)
     }
 }
+
+#[derive(Debug)]
+pub struct SelectionField {
+    name: String,
+    args: Vec<(String, String)>,
+    directive: async_graphql::ServerResult<Vec<ConstDirective>>,
```
meskill, line 77:
```suggestion
    directives: async_graphql::ServerResult<Vec<ConstDirective>>,
```

meskill: we don't use ServerResult anyway after that. Consider storing just vec or `Option<Vec>`

src/core/jit/context.rs
```diff
@@ -1,22 +1,26 @@
...
+use super::{Field, Nested, Request};
+use crate::core::ir::{ResolverContextLike, SelectionField};
 
 /// Rust representation of the GraphQL context available in the DSL
-#[derive(Getters, Debug, Clone)]
+#[derive(Debug, Clone)]
 pub struct Context<'a, Input, Output> {
     request: &'a Request<Input>,
     value: Option<&'a Output>,
     args: Option<indexmap::IndexMap<Name, Input>>,
```
meskill, line 12: in that case `args` becomes redundant since args are also present in field. But we can't just migrate it because it'll broke the some cases with default value. Maybe add todo to migrate?

src/core/jit/exec.rs
```diff
@@ -118,7 +117,7 @@ where
                     if let Some(array) = value.as_array() {
                         join_all(field.nested_iter().map(|field| {
                             join_all(array.iter().enumerate().map(|(index, value)| {
-                                let ctx = ctx.with_value(value); // Output::JsonArray::Value
+                                let ctx = ctx.with_value(value).with_field(field); // Output::JsonArray::Value
```
meskill, line 120: seems like we always setting both field and value. Would it be better to merge it into single operation?

meskill (approved):

PR #2451, opened 2024-07-17
chore: add directive support for field method in context

laststylebender14:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/core/jit/builder.rs
+++ b/src/core/jit/builder.rs
@@ -607,4 +607,22 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn test_directives() {
+        let plan = plan(
+            r#"
+            query($includeName: Boolean! = true) {
+                users {
+                    id @options(paging: $includeName)
+                    name @include(if: $includeName)
+                }
+            }
+            "#,
+            &Variables::new(),
+        );
+
+        assert!(plan.is_query());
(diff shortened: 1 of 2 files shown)


src/core/jit/builder.rs
```diff
@@ -145,6 +166,30 @@ impl Builder {
...
+                        if directive.name.node == "skip" || directive.name.node == "include" {
+                            continue;
+                        }
+                        let mut arguments = Vec::with_capacity(directive.arguments.len());
+                        for (name, value) in &directive.arguments {
+                            arguments.push((
+                                name.clone(),
+                                value.position_node(value.node.clone().into_const_with(
+                                    |name| {
+                                        self.get_variable_value(
```
meskill, line 181: the `src/core/jit/input_resolver.rs` implements the same logic but only for `field.args`.

we need to merge this together somehow and remove duplication

meskill (changes requested):

src/core/jit/builder.rs
meskill: we can implement proper defaulting later, right now we should provide some abstractions to easier integrate this. Because of that I see the reason to use single entity to resolve variables in both args and directives

meskill:
can we close this one in favor of https://github.com/tailcallhq/tailcall/pull/2469 ?

PR #2452, opened 2024-07-17
fix(jit): union support

meskill:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/jit/exec.rs
```diff
@@ -171,8 +184,41 @@ where
     }
 }
 
+#[derive(Clone)]
+pub struct ExecResult<V> {
+    pub value: V,
+    pub type_name: Option<TypeName>,
+}
```
tusharmath, line 191:
```suggestion
pub struct TypedValue<V> {
    pub value: V,
    pub name: Option<TypeName>,
}
```

src/core/jit/exec.rs
```diff
@@ -171,8 +184,41 @@ where
...
+
+pub struct ExecResultRef<'a, V> {
+    pub value: &'a V,
+    pub type_name: Option<&'a TypeName>,
+}
```
tusharmath, line 196: This seems like is morally equivalent to a `Cow<'a, TypedValue<&'a V>>`.

src/core/jit/model.rs
```diff
@@ -282,27 +320,31 @@ impl<Extensions: Debug, Input: Debug> Debug for Field<Extensions, Input> {
 
 /// Stores field relationships in a flat structure where each field links to its
 /// parent.
-#[derive(Clone)]
-pub struct Flat(FieldId);
+#[derive(Clone, Debug)]
+pub struct Flat {
+    parent_id: FieldId,
+    as_type: Option<String>,
```
tusharmath, line 326: If this is the expected type, we should store it in `Field` instead of `Flat`.

src/core/jit/model.rs
```diff
@@ -282,27 +320,31 @@ impl<Extensions: Debug, Input: Debug> Debug for Field<Extensions, Input> {
...
 
 /// Store field relationships in a nested structure like a tree where each field
 /// links to its children.
 #[derive(Clone, Debug)]
-pub struct Nested<Input>(Vec<Field<Nested<Input>, Input>>);
+pub struct Nested<Input> {
+    /// usual fields without fragment definition
+    fields: Vec<Field<Nested<Input>, Input>>,
+    /// fields from the fragment based on the fragment's type
+    by_type: IndexMap<String, Vec<Field<Nested<Input>, Input>>>,
```
tusharmath, line 346: Like `as_type` This should also be stored on the field.

src/core/jit/exec.rs
meskill: no, `type_name` is also a reference here while in base struct its an owned value

src/core/jit/model.rs
meskill: refactored

src/core/jit/snapshots/tailcall__core__jit__builder__tests__resolving_operation-2.snap
```diff
@@ -44,27 +45,31 @@ expression: plan.into_nested()
                         id: 1,
                         name: "id",
                         type_of: ID!,
+                        parent_type: "Post",
```
tusharmath, line 48: We are storing the parent type for performance reasons? Can't we perform a simple lookup and find the parent information in O(1) complexity?

tusharmath (approved):

src/core/jit/snapshots/tailcall__core__jit__builder__tests__resolving_operation-2.snap
meskill: not exactly. The type of the parent could be different to this property because it could contain interface or union variant whilst type of the parent is some concrete type/union. 

perhaps, different naming could help? like `on_type`?

tusharmath: Understood. Lets rename it to `type_condition` add a bit of documentation. 

More: https://spec.graphql.org/October2021/#sec-Type-Conditions

meskill: done

PR #2454, opened 2024-07-17
chore: taillcall prettier anyhow cleanup

mehulmathur16:


tests/core/spec.rs
```diff
@@ -131,12 +131,12 @@ async fn check_server_config(spec: ExecutionSpec) -> Vec<Config> {
 
                 let actual = tailcall_prettier::format(actual, &tailcall_prettier::Parser::Gql)
                     .await
-                    .context(context.clone())
+                    .map_err(|_| anyhow!("{}", context.clone()))
```
tusharmath, line 134: We are dropping the actual error and just displaying the file path.

tusharmath (approved):

PR #2461, opened 2024-07-18
chore: make `validate` function generic in `Scalar` trait

### LINKED ISSUE #2460: chore: make `validate` function generic in `Scalar` trait
Currently, the existing `validate` functions is bound to use `ConstValue` because it's used for `async_graphql::Schema`

But to use it with a parameterized Store (PR: #2437) we need a new function that returns a function that takes a generic implementation of `JsonLikeOwned`

**Technical Requirements**
- Update scalar validations to accept the following signature.
	```rust
	fn validate<Value: JsonLike<'a>>(&'a self) -> fn(&'a Value) -> bool;
	```
- Ensure all existing tests pass
- Write a special handler for async_graphql to handle ConstValue

ssddOnTop:
**Issue Reference(s):**  
Fixes #2460
/claim #2460

### DIFF
--- a/src/core/scalar/bytes.rs
+++ b/src/core/scalar/bytes.rs
@@ -3,7 +3,7 @@ use schemars::schema::Schema;
 use schemars::{schema_for, JsonSchema};
 use tailcall_macros::ScalarDefinition;
 
-use crate::core::json::JsonLike;
+use crate::core::json::{JsonLike, JsonLikeOwned};
 
 /// Represents list of bytes
 #[derive(JsonSchema, Default, ScalarDefinition)]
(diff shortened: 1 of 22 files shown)


src/core/scalar/bytes.rs
```diff
@@ -14,6 +14,10 @@ impl super::Scalar for Bytes {
         |value| value.as_str().is_some()
     }
 
+    fn validate_generic<Value: JsonLikeOwned>(&self) -> fn(&Value) -> bool {
```
tusharmath, line 17: if we have validate_generic, then wy do we need `validate`?

ssddOnTop: for current impl with async-graphql..

tusharmath:
```suggestion
    fn validate<'a, Value: JsonLike<'a>>(&'a self) -> fn(&'a Value) -> bool {
```

tusharmath (approved):

PR #2469, opened 2024-07-19
chore: impl custom directive for JIT

laststylebender14:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/ir/resolver_context_like.rs
```diff
@@ -84,17 +84,30 @@ impl From<async_graphql::SelectionField<'_>> for SelectionField {
...
         Self::from_jit_field(value)
     }
 }
 
 impl SelectionField {
-    fn from_jit_field<Input: ToString>(
-        field: &crate::core::jit::Field<Nested<Input>, Input>,
+    fn from_jit_field(
+        field: &crate::core::jit::Field<
+            Nested<async_graphql_value::ConstValue>,
```
meskill, line 108: maybe just use import, all that `async_graphql_value::` looks redundant

src/core/jit/model.rs
```diff
@@ -314,3 +319,58 @@ impl<Input> OperationPlan<Input> {
...
+    pub fn try_map<Output, Error>(
+        self,
+        map: impl Fn(Input) -> Result<Output, Error>,
+    ) -> Result<Directive<Output>, Error> {
+        Ok(Directive {
+            name: self.name,
+            arguments: self
+                .arguments
+                .into_iter()
+                .filter_map(|(k, v)| map(v).ok().map(|v| (k, v)))
```
meskill, line 372: why do you ignore any possible errors here making the call infallible?

src/core/jit/model.rs
```diff
@@ -314,3 +319,58 @@ impl<Input> OperationPlan<Input> {
         self.flat.len()
     }
 }
+
+/// DirectiveAdapter helps to bridge the gap between async_graphql's
+/// ConstDirective and our custom Directive implementation.
+pub trait DirectiveAdapter {
+    fn name(&self) -> &str;
+    fn arguments(&self) -> Vec<(String, &ConstValue)>;
```
meskill, line 327: seems inconsistent to have owned String and reference. Could we return both references and evade string cloning?

meskill: Also, returning Iterator could be preferred since it's not involving allocation for the vec if not needed

meskill (dismissed):

src/core/document.rs
```diff
@@ -61,7 +63,9 @@ fn get_formatted_docs(docs: Option<String>, indent: usize) -> String {
     formatted_docs
 }
 
-pub fn print_directives<'a>(directives: impl Iterator<Item = &'a ConstDirective>) -> String {
+pub fn print_directives<'a, Input: DirectiveAdapter + 'a>(
+    directives: impl Iterator<Item = &'a Input>,
+) -> String {
```
tusharmath, line 68: Little unclear to me why this should change?

src/core/jit/model.rs
```diff
@@ -360,3 +379,54 @@ impl<Input> OperationPlan<Input> {
...
+    fn name(&self) -> &str {
+        &self.name.node
+    }
+
+    fn arguments(&self) -> impl Iterator<Item = (&str, &ConstValue)> {
+        self.arguments
+            .iter()
+            .map(|(k, v)| (k.node.as_str(), &v.node))
+    }
+}
```
tusharmath, line 400: We can create a `DirectiveAdapter::from(const_directive)` instead which performs a ZERO copy transformation.

tusharmath (approved):

tusharmath (approved):

PR #2475, opened 2024-07-19
feat(2468): support for `skipNull: true` in HTTP directive query parameters

### LINKED ISSUE #2468: Add support for `skipNull: true` in HTTP directive query parameters
#### Summary
The current GraphQL schema allows for HTTP requests to be configured via the `@http` directive. However, there is a need to support skipping null query parameters to avoid sending unnecessary or invalid data in requests. 

After https://github.com/tailcallhq/tailcall/issues/2467 the default behaviour will not skip the null params

#### Example
Consider the following GraphQL type definition:

```graphql
type User {
  posts: [Post]
    @http(
      path: "/posts"
      query: [{ key: "author_ids[]", value: "{{.value.authorID}}", skipNull: true }]
      batchKey: ["author_ids[]"]
    )
}
```

In the above example, if `{{.value.authorID}}` resolves to null, the `author_ids[]` parameter would be skipped entirely in the HTTP request.

skipNull: true
when specified in query params will skip the param whole value is null the default value of this argument is false

ssddOnTop:
**Issue Reference(s):**  
Fixes #2468
/claim #2468


src/core/config/key_values.rs
```diff
@@ -3,19 +3,23 @@ use std::ops::Deref;
 
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
+use crate::core::is_default;
+
+type Value = (String, Option<bool>);
```
tusharmath, line 8: Unclear, why this change is required?

src/core/config/key_values.rs
```diff
@@ -24,6 +28,11 @@ impl FromIterator<(String, String)> for KeyValues {
 pub struct KeyValue {
     pub key: String,
     pub value: String,
+    #[serde(default, skip_serializing_if = "is_default")]
+    #[serde(rename = "skipNull")]
```
tusharmath, line 32: Use the camelCase annotation of serde on the whole type.

src/core/config/key_values.rs
```diff
@@ -24,6 +28,11 @@ impl FromIterator<(String, String)> for KeyValues {
...
+    /// When specified in query params will skip the param whole value is null
+    /// the default value of this argument is false
+    pub skip_null: Option<bool>,
```
tusharmath, line 35:
```suggestion
    /// Determines whether to ignore query parameters with empty values when forming URLs.
    ///
    /// When set to true, query parameters without values are completely ignored
    /// during URL formation. When false (default), parameters without values
    /// are included in the URL.
    pub skip_null: Option<bool>,
```

src/core/http/data_loader.rs
```diff
@@ -111,10 +112,7 @@ impl Loader<DataLoaderRequest> for HttpDataLoader {
             for dl_req in dl_requests.iter() {
                 let url = dl_req.url();
                 let query_set: HashMap<_, _> = url.query_pairs().collect();
-                let id = query_set.get(query_name).ok_or(anyhow::anyhow!(
-                    "Unable to find key {} in query params",
-                    query_name
-                ))?;
+                let id = query_set.get(query_name).unwrap_or(&Cow::Borrowed(""));
```
tusharmath, line 115: Why did we drop the error here?

tests/execution/batching-group-by-optional-key-skip-null.md
```diff
@@ -0,0 +1,64 @@
...
+
+```graphql @config
+schema
+  @server(port: 8000, queryValidation: false)
+  @upstream(baseURL: "http://jsonplaceholder.typicode.com", httpCache: 42, batch: {delay: 1, maxSize: 1000}) {
+  query: Query
+}
+
+type Query {
+  posts: [Post] @http(path: "/posts?id=11&id=3&foo=1")
```
tusharmath, line 11:
```suggestion
  posts: [Post] @http(path: "/posts")
```
Don't think we need the query params here.

src/core/config/key_values.rs
ssddOnTop: to avoid defining tuple again and again

generated/.tailcallrc.graphql
```diff
@@ -536,6 +536,12 @@ input Step {
 
 input KeyValue {
   key: String!
+  """
+  Determines whether to ignore query parameters with empty values when forming URLs.When 
+  set to true, query parameters without values are completely ignored during URL formation. 
+  When false (default), parameters without values are included in the URL.
+  """
+  skipNull: Boolean
```
tusharmath, line 544:
```suggestion
  skipEmpty: Boolean
```

src/core/config/key_values.rs
```diff
@@ -3,27 +3,40 @@ use std::ops::Deref;
...
     pub key: String,
     pub value: String,
+    #[serde(default, skip_serializing_if = "is_default")]
+    /// Determines whether to ignore query parameters with empty values when
+    /// forming URLs.
+    ///
+    /// When set to true, query parameters without values are completely ignored
+    /// during URL formation. When false (default), parameters without values
+    /// are included in the URL.
+    pub skip_null: Option<bool>,
```
tusharmath, line 39:
```suggestion
    pub skip_empty: Option<bool>,
```

src/core/config/key_values.rs
```diff
@@ -3,27 +3,40 @@ use std::ops::Deref;
 
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
+use crate::core::is_default;
+
+type Value = (String, Option<bool>);
+
 #[derive(Clone, Debug, Default, Eq, PartialEq, schemars::JsonSchema)]
-pub struct KeyValues(pub BTreeMap<String, String>);
+pub struct KeyValues(pub BTreeMap<String, Value>);
```
tusharmath, line 11:
```suggestion
pub struct KeyValues<Value>(BTreeMap<String, Value>);
```
Make it polymorphic.

ssddOnTop: Value is not something from JL, it's `type Value = (String, Option<bool>);`

src/core/config/key_values.rs
```diff
@@ -3,27 +3,40 @@ use std::ops::Deref;
...
     pub key: String,
     pub value: String,
+    #[serde(default, skip_serializing_if = "is_default")]
+    /// Determines whether to ignore query parameters with empty values when
+    /// forming URLs.
+    ///
+    /// When set to true, query parameters without values are completely ignored
+    /// during URL formation. When false (default), parameters without values
+    /// are included in the URL.
+    pub skip_empty: Option<bool>,
```
tusharmath, line 39: We should keep the original `KeyValue` as it is because its being used in multiple places. Create a new one calle `URLQuery` type that has `key`, `value` and `skip_empty`.

src/core/config/server.rs
```diff
@@ -289,28 +289,68 @@ mod tests {
...
+            KeyValue {
+                key: "1".to_string(),
+                value: "1, Left".to_string(),
+                skip_empty: None,
+            },
+            KeyValue {
+                key: "2".to_string(),
+                value: "2, Left".to_string(),
+                skip_empty: None,
+            },
```
tusharmath, line 306: Revert these changes.

src/core/config/telemetry.rs
```diff
@@ -145,14 +145,22 @@ mod tests {
         let exporter_otlp_1 = Telemetry {
             export: Some(TelemetryExporter::Otlp(OtlpExporter {
                 url: "test-url".to_owned(),
-                headers: vec![KeyValue { key: "header_a".to_owned(), value: "a".to_owned() }],
+                headers: vec![KeyValue {
+                    key: "header_a".to_owned(),
+                    value: "a".to_owned(),
+                    skip_empty: None,
+                }],
```
tusharmath, line 152: Revert these changes.

tusharmath:
closing in favour of https://github.com/tailcallhq/tailcall/pull/2669

PR #2479, opened 2024-07-20
chore: drop lifetime from `JsonLike`

ssddOnTop:


ssddOnTop:
@tusharmath @meskill 

I wasn't able to fix lifetime issues for `insert_key` can I get some help pls 😓

meskill:
> @tusharmath @meskill 
> 
> I wasn't able to fix lifetime issues for `insert_key` can I get some help pls 😓  
>  

can you demonstrate what the issue?

ssddOnTop:
> can you demonstrate what the issue?

JsonLike works fine now.. `insert_key` functions is ugly but if I bring back lifetime for JsonObjectLike then we can change signature of the function to `fn insert_key(mut self, &'key, value: Self::Value<'a>) -> Self`

Before diving into actual problem, here's a brief background:

if I merge this branch to the PR: https://github.com/tailcallhq/tailcall/pull/2437 I face additional problems in creation of Value::Object and Value::Array, but moreover now it's very difficult to define generic values, we have so many nested types and defining them with lifetime is difficult.

However, that PR works fine and shows same result as main branch with and without JIT. But I am unable to write unit test with BorrowedValue

meskill:
> > can you demonstrate what the issue?
> 
> JsonLike works fine now.. `insert_key` functions is ugly but if I bring back lifetime for JsonObjectLike then we can change signature of the function to `fn insert_key(mut self, &'key, value: Self::Value<'a>) -> Self`
> 
> Before diving into actual problem, here's a brief background:
> 
> if I merge this branch to the PR: #2437 I face additional problems in creation of Value::Object and Value::Array, but moreover now it's very difficult to define generic values, we have so many nested types and defining them with lifetime is difficult.
> 
> However, that PR works fine and shows same result as main branch with and without JIT. But I am unable to write unit test with BorrowedValue

Sorry, I don't have much time in coming days and I'm not sure I've got the issue right.

I've tried to play around with this pr to see what can be improved. I've noticed that signature for `insert_key` is a little unusual and replacing it with more convenient signature may actually justify having lifetime parameter for `JsonObjectLike` trait, but this of course involves some problems to solve. 

I have put my experiments to https://github.com/tailcallhq/tailcall/pull/2481 in case it could be helpful.

src/core/json/json_like.rs
```diff
@@ -1,35 +1,46 @@
...
 /// A trait for objects that can be used as JSON values
-pub trait JsonLike<'a>: Sized {
-    type JsonObject: JsonObjectLike<'a, Value = Self>;
+pub trait JsonLike: Sized {
+    type JsonObject<'a>: JsonObjectLike<'a>
+    where
+        Self: 'a;
+    type Output<'a>: JsonLike
+    where
+        Self: 'a;
```
tusharmath, line 13: @meskill `Output` needs to be of type `Self` here not any JsonLike. That constraint is the responsibility of JsonLike and not of the implementor. The implementor needs to ensure it matches the trait.

meskill:
superseded by https://github.com/tailcallhq/tailcall/pull/2586

Issue #2482, opened 2024-07-21
refactor: Make `JsonPlaceholder.rs` parametric


tusharmath:
Currently `json_placeholder.rs` is tightly coupled with `ConstValue` implementation. Refactor that out and use a generic `Value` type-parameter and leverage `JsonLike` internally.

ssddOnTop:
Looks like duplicate issue

ssddOnTop:
feel free to attempt even while I am working on this one.. I am not sure if I'll be able to close the issue.

Anshgrover23:
@ssddOnTop thanks bro. i am also trying. best luck

Anshgrover23:
@ssddOnTop sorry bro i am not able to do i think i am closing my attempt

PR #2484, opened 2024-07-21
chore: parameterize placeholder

ssddOnTop:
related to: #2482

This works but only for values with Owned type.. i.e. Const and Serde value.

HRTB: `for<'a> JsonLike` would NOT work with Borrowed value. 

I think entire implementation at #2437 needs change 😅


ssddOnTop:
closing in favour of #2505

PR #2485, opened 2024-07-21
chore: drop HRTB from `Synth`

ssddOnTop:
some HTRB blunders are fixed here..

**TODO**
~~- [ ] drop JsonLikeOwned from Scalar~~

### DIFF
--- a/src/core/jit/synth/synth.rs
+++ b/src/core/jit/synth/synth.rs
@@ -132,8 +132,7 @@ impl<'a, Value: JsonLike<'a> + Clone + 'a> Synth<Value> {
     ) -> Result<Value, Positioned<Error>> {
         let include = self.include(node);
         if include && node.is_scalar {
-            let scalar = scalar::get_scalar(node.type_of.name())
-                .unwrap_or(scalar::ScalarType::JSON(Default::default()));
+            let scalar = scalar::get_scalar(node.type_of.name());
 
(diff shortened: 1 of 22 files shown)


src/core/scalar/date.rs
```diff
@@ -15,7 +15,18 @@ pub struct Date {
...
+        |value| {
+            if let Some(date_str) = value.as_str() {
+                if DateTime::parse_from_rfc3339(date_str).is_ok() {
+                    return true;
+                }
+            }
+            false
+        }
+    }
+
```
tusharmath, line 28: Extreme code duplication.

src/core/jit/synth/synth.rs
```diff
@@ -288,10 +295,14 @@ mod tests {
 
     const CONFIG: &str = include_str!("../fixtures/jsonplaceholder-mutation.graphql");
 
-    fn init<Value: JsonLikeOwned + DeserializeOwned + Serialize + Clone>(
+    fn extend_lifetime_str<'b>(r: &'b str) -> &'static str {
+        unsafe { std::mem::transmute::<&'b str, &'static str>(r) }
+    }
```
tusharmath, line 300: find an alternative strategy so that unsafe isn't required.

ssddOnTop: can we make try_map some external function then?

tusharmath (approved):

PR #2495, opened 2024-07-22
fix(#2489): Modify batchKey logic

karatakis:
**Summary:** 

- `batchKey` is now used to state which query parameter is used for the batching
- `batchPath` is used to instruct Tailcall how to groupBy items

/claim #2489

**Issue Reference(s):**  

Fixes #2489

**Build & Testing:**

- [X] I ran `cargo test` successfully.
- [X] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [X] I have added relevant unit & integration tests.
- [ ] I have updated the [documentation] accordingly.
- [X] I have performed a self-review of my code.
- [X] PR follows the naming convention of `<type>(<optional scope>): <title>`

[documentation]: https://github.com/tailcallhq/tailcallhq.github.io/tree/develop/docs

PR #2497, opened 2024-07-22
fix: allow users to specify extraction path

### LINKED ISSUE #2489: Batching with nested HTTP responses
Given this example:

```gql
type Transaction {
  id: ID!
  slug: String!
  bank_account_id: String! 
  BankAccounts: BankAccountResponse @http(
      path: "/v1/bank_accounts"
      query: [{ key: "bank_account_ids[]", value: "{{.value.bank_account_id}}" }]
      batchKey: ["bank_account_ids[]"]
    )
}

type BankAccountResponse {
  bank_accounts: [BankAccount!]!
}

type BankAccount {
  id: ID!
  bic: String!
  iban: String!
}
```

Without batching, this work fine to get the BankAccount information. 

However, with batching, the query is indeed batched but all the `Transactions.BankAccounts` are null.

As discussed on discord, the suggestion was to use `["bank_accounts", "id"]` as the batching key however this doesn't match the query param and the error `Unable to find key id in query params` is returned.

dekkku:
**Summary:**  
problem: when we specify the batchKey: ["a","b"] we essentially expect the path to target and assume the last value to be the query param key.
eg.
in the discussion example, `batchKey: ["bank_accounts", "id"]` , tells the server `id` will be used for the extraction of json item from batched endpoint's response and `id` will be used form a batched end point which isn't the case when extraction property name is different than the name used to form batched endpoint. so we need more information in order to solve this problem.

proposed solution: 
```graphql
      batch: {batchKey: "bank_account_ids[]", extractionPath: ["bank_accounts", "id"]}
```
 here we tell form the endpoint with batchKey and use extraction path for extract the json object from batched endpoint.
i.e when creating batched endpoint use `bank_account_ids[]` and in the batched endpoint response go to `bank_accounts.id` to figure out merging.

**Issue Reference(s):**  
Fixes #2489
/claim #2489

**Build & Testing:**

**Checklist:**

### DIFF
--- a/examples/grpc-reflection.graphql
+++ b/examples/grpc-reflection.graphql
@@ -10,7 +10,7 @@ type Query {
   news: NewsData! @grpc(method: "news.NewsService.GetAllNews")
   newsById(news: NewsInput!): News! @grpc(method: "news.NewsService.GetNews", body: "{{args.news}}")
   newsByIdBatch(news: NewsInput!): News!
-    @grpc(method: "news.NewsService.GetMultipleNews", body: "{{args.news}}", batchKey: ["news", "id"])
+    @grpc(method: "news.NewsService.GetMultipleNews", body: "{{args.news}}", batch: {batchKey: "id", extractionPath: ["news", "id"]})
 }
 
 type News {
--- a/examples/grpc-reflection.graphql
+++ b/examples/grpc-reflection.graphql
@@ -10,7 +10,11 @@ type Query {
   news: NewsData! @grpc(method: "news.NewsService.GetAllNews")
   newsById(news: NewsInput!): News! @grpc(method: "news.NewsService.GetNews", body: "{{args.news}}")
   newsByIdBatch(news: NewsInput!): News!
-    @grpc(method: "news.NewsService.GetMultipleNews", body: "{{args.news}}", batch: {batchKey: "id", extractionPath: ["news", "id"]})
+    @grpc(
+      method: "news.NewsService.GetMultipleNews"
+      body: "{{args.news}}"
+      batch: {batchKey: "id", extractionPath: ["news", "id"]}
+    )
 }
 
 type News {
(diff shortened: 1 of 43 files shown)


angristan:
This is really, really nice! 🫶 

I was able to check this PR locally and test it out. It's even better than I expected, because you basically solved my two concerns from https://github.com/tailcallhq/tailcall/discussions/2471. 

Although it fits my use case, I feel like making the batching work with nested resources should be separeted from "extracting" or flattening the resource. For example, it's the not case here but I could want to access `meta` (or any other value) that is alongside `bank_accounts` in the response while still having the batching work.

More importantly it seems there's an batching issue 🤔 

```
INFO POST http://localhost:3002/v1/transactions HTTP/1.1
 INFO GET http://localhost:3002/v1/membershipsfilters%5Bids%5D%5B%5D=2696aeb5-eb94-45e5-8bb1-3cd163a1b668&filters%5Bids%5D%5B%5D=5182052c-180e-451a-ab48-1874f83d03d7&filters%5Bids%5D%5B%5D=66f1bbf4-89f0-4c5c-becd-2fc5d1fb6947&filters%5Bids%5D%5B%5D=f601e869-4384-4df6-a286-39e95ce6ce08 HTTP/1.1
 INFO GET http://localhost:3006/v1/teams/?ids[]=4e3a67d5-1856-4fb3-9aa6-d4238d6e9c01&ids%5B%5D=5df40b69-a4d0-448e-85ec-3429a7815c34 HTTP/1.1
 INFO GET http://localhost:3002/v1/bank_accounts?bank_account_ids[]=ae73e825-fea4-49d5-a25f-e2f792a8314a&bank_account_ids%5B%5D=e7f60769-8513-4215-9dde-1574b00d770b HTTP/1.1
 INFO GET http://localhost:3002/v1/bank_accounts?bank_account_ids[]=270a7f73-c941-468c-898c-6e91ac14ebc7&bank_account_ids%5B%5D=e7f60769-8513-4215-9dde-1574b00d770b HTTP/1.1
```

Here I am also batching other resources (here memberships and teams), which seem to work fine. But there is two calls to `bank_accounts` with two IDs each time, one of which is the same in the two calls.

tusharmath:
Thanks @dekkku closing it in favour of https://github.com/tailcallhq/tailcall/pull/2495

It was a much smaller and simple change

PR #2508, opened 2024-07-23
chore: enum validation conflict fixes

laststylebender14:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/jit/synth/synth.rs
```diff
@@ -129,25 +131,42 @@ impl<'a, Value: JsonLike<'a> + Clone + 'a> Synth<Value> {
...
+                .map(|v| self.plan.field_validate_enum_value(node, v))
+                .unwrap_or(false)
+            {
+                Ok(parent.clone())
+            } else {
+                Err(
+                    ValidationError::EnumInvalid { type_of: node.type_of.name().to_string() }
+                        .into(),
+                )
+                .map_err(|e| self.to_location_error(e, node))
```
meskill, line 169: we can have single `map_err` call to the whole if-else block that way reducing duplication in every branch

src/core/jit/context.rs
```diff
@@ -84,11 +84,11 @@ impl<'a> ResolverContextLike for Context<'a, ConstValue, ConstValue> {
...
 
     fn is_query(&self) -> bool {
-        self.is_query
+        self.plan.is_query()
     }
 
-    fn add_error(&self, _error: ServerError) {
-        todo!()
+    fn add_error(&self, error: ServerError) {
+        self.plan.add_error(error.into())
```
tusharmath, line 91: Plans should ideally remain immutable and reusable between queries. `Context` is a good place to keep mutable content instead.

src/core/jit/error.rs
```diff
@@ -51,52 +56,67 @@ impl ErrorExtensions for Error {
...
-        match self {
-            // async_graphql::parser::Error has special conversion to ServerError
-            Error::ParseError(error) => error.into(),
-            error => error.into_server_error_with_pos(None),
+#[derive(Debug, Serialize, Clone)]
+pub struct LocationError<Err> {
+    pub error: Err,
+    pub pos: Pos,
+    pub path: Vec<PathSegment>,
+}
```
tusharmath, line 87:
```suggestion
pub struct Positioned<Value> {
    pub value: Value,
    pub pos: Pos,
    pub path: Vec<String>,
}
```

Can generalize it.

src/core/jit/model.rs
```diff
@@ -388,6 +420,22 @@ impl<Input> OperationPlan<Input> {
     pub fn size(&self) -> usize {
         self.flat.len()
     }
+
+    pub fn field_is_scalar<Extensions>(&self, field: &Field<Extensions, Input>) -> bool {
+        self.index.type_is_scalar(field.type_of.name())
```
tusharmath, line 425: This information should be added to the field at the time of planning instead of dynamically performing a string comparison everytime.

src/core/jit/model.rs
```diff
@@ -388,6 +420,22 @@ impl<Input> OperationPlan<Input> {
...
+    }
+
+    pub fn field_is_enum<Extensions>(&self, field: &Field<Extensions, Input>) -> bool {
+        self.index.type_is_enum(field.type_of.name())
+    }
```
tusharmath, line 430: Again, the fact that Field is an enum should be known at the time of its creation.

src/core/jit/model.rs
```diff
@@ -304,11 +302,21 @@ impl Debug for Flat {
...
 pub struct Nested<Input>(Vec<Field<Nested<Input>, Input>>);
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct OperationPlan<Input> {
     flat: Vec<Field<Flat, Input>>,
     operation_type: OperationType,
     nested: Vec<Field<Nested<Input>, Input>>,
+    errors: Arc<Mutex<Vec<LocationError<Error>>>>,
+    pub index: Arc<Index>,
```
tusharmath, line 311: Drop `errors` and `index`. Lets attempt to make `OperationPlan` remains immutable.

src/core/jit/context.rs
meskill: we could've reviewed this before merging https://github.com/tailcallhq/tailcall/pull/2518 and now this pr contains both changes.

better revert this merge and fix it separately

src/core/jit/model.rs
meskill: we could basically encode this info into field itself, but in that case the field will become another representation of the index. 
While it's lightweighed to store boolean flags in the field, for cases like enum validations we need also to store enum variants that's imply string cloning

src/core/jit/error.rs
laststylebender14: done, moved to model and added `PathSegment` and  `Post`.

PR #2512, opened 2024-07-24
Refactor Scalar Trait to Enum-Based Implementation

Anshgrover23:
**Summary:**  

Refactor the scalar type system by replacing the Scalar trait with an enum-based ScalarType. This simplifies the implementation and improves maintainability.

- Removed the Scalar trait.
- Introduced the ScalarType enum with validate, schema, and name methods.
- Refactored scalar_definition.rs to work with the new ScalarType.
**Issue Reference(s):**  
Fixes #2511
Closes #2511 
/claim #2511

**Build & Testing:**

- [x] I ran `cargo test` successfully.
- [x] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [ ] I have added relevant unit & integration tests.
- [ ] I have updated the [documentation] accordingly.
- [x] I have performed a self-review of my code.
- [x] PR follows the naming convention of `<type>(<optional scope>): <title>`

[documentation]: https://github.com/tailcallhq/tailcallhq.github.io/tree/develop/docs

PR #2515, opened 2024-07-24
refactor: clean up scalar 

### LINKED ISSUE #2511: Clean up Scalar
- Drop the trait `Scalar` and leverage the enum to implement `validate` and `schema`.
- Scalars currently store some sort of value which is typically neither constructed or used so drop the inner values and make them an empty struct.

ssddOnTop:
**Issue Reference(s):**  
Fixes #2511
/claim 2511


Anshgrover23:
@ssddOnTop hi bro i have tried the same thing. as you do so i also have to delete those files that u have deleted.

Anshgrover23:
@ssddOnTop can we split the bounty on this because i have also worked for this?. if u want to, or if u dont want then you can continue buddy.

ssddOnTop:
Hi @Anshgrover23 I wouldn't mind splitting bounty if I took over an issue that you've been attempting, but I have a few points to make..

1. As I was assigned to the issue, you shouldn't have attempted in the first place 😓 
2. I reviewed your [code](https://github.com/tailcallhq/tailcall/pull/2512/commits/8afdb2d65ec70f061a5eb466630b190cfba21ed1) and.. it's not reducing anything but it just adds up to the mess so I decided to raise a PR with my version 

hope you understand

src/core/scalar.rs
```diff
@@ -0,0 +1,312 @@
...
+            ScalarType::Int16 => eval_signed(value, i16::try_from),
+            ScalarType::Int32 => eval_signed(value, i32::try_from),
+
+            ScalarType::UInt8 => eval_unsigned(value, u8::try_from),
+            ScalarType::UInt16 => eval_unsigned(value, u16::try_from),
+            ScalarType::UInt32 => eval_unsigned(value, u32::try_from),
+        }
+    }
+    pub fn get_scalar(name: &str) -> ScalarType {
+        CUSTOM_SCALARS.get(name).cloned().unwrap_or(Self::Empty)
```
tusharmath, line 88: Should return an option, let the caller decide what to do if nothing is found. Add docs.

src/core/scalar.rs
```diff
@@ -0,0 +1,312 @@
...
+            }
+        );
+        if let Some(format) = format {
+            value["format"] = serde_json::json!(format);
+        }
+
+        let metadata = serde_json::from_value(value).unwrap();
+        Schema::Object(SchemaObject { metadata: Some(Box::new(metadata)), ..Default::default() })
+    }
+    pub fn schema(&self) -> Schema {
```
tusharmath, line 116: Can't we get the schema using schemar for each enum variant?

src/core/scalar.rs
```diff
@@ -0,0 +1,312 @@
...
+        set.extend(CUSTOM_SCALARS.keys().map(|k| k.as_str()));
+        set
+    };
+}
+
+///
+/// Check if the type is a predefined scalar
+pub fn is_predefined_scalar(type_name: &str) -> bool {
+    SCALAR_TYPES.contains(type_name)
+}
```
tusharmath, line 263: Move to the default implementation of Scalar

src/core/scalar.rs
```diff
@@ -0,0 +1,312 @@
...
+        hm
+    };
+}
+lazy_static! {
+    static ref SCALAR_TYPES: HashSet<&'static str> = {
+        let mut set = HashSet::new();
+        set.extend(["String", "Int", "Float", "ID", "Boolean"]);
+        set.extend(CUSTOM_SCALARS.keys().map(|k| k.as_str()));
+        set
+    };
```
tusharmath, line 256: Move to default implementation of Scalar

src/core/scalar.rs
```diff
@@ -0,0 +1,312 @@
...
+            ScalarType::UInt128,
+            ScalarType::Bytes,
+        ];
+        let mut hm = HashMap::new();
+
+        for scalar in scalars {
+            hm.insert(scalar.name(), scalar);
+        }
+        hm
+    };
```
tusharmath, line 248: Move to default implementation of Scalar

src/core/scalar.rs
```diff
@@ -0,0 +1,312 @@
...
+        // It is easy to forget to add a new scalar type to the CUSTOM_SCALARS map
+        // This test ensures that all scalar types are correctly defined
+        macro_rules! check_custom_scalars {
+            ($enum_type:ty, $hashmap:expr) => {{
+                let mut all_present = true;
+                for variant in <$enum_type>::iter() {
+                    let variant_name = format!("{:?}", variant);
+                    if !$hashmap.contains_key(&variant_name) {
+                        all_present = false;
+                        println!("Missing variant: {}", variant_name);
```
tusharmath, line 283: Println?

tusharmath (changes requested):

src/core/scalar.rs
ssddOnTop: ??

src/core/scalar.rs
ssddOnTop: if scalar is not found then validate function should always return `true` by default

src/core/scalar.rs
ssddOnTop: Introduced a Doc macro which gets doc comments from varients' instance. We can not do this using schemars

PR #2516, opened 2024-07-24
refactor(core): modularize error.rs

mehulmathur16:


src/core/error.rs
```diff
@@ -1,159 +0,0 @@
...
-impl Display for cache::Error {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        match self {
-            cache::Error::SerdeJson(error) => write!(f, "Serde Json Error: {}", error),
-            cache::Error::Kv(error) => write!(f, "Kv Error: {}", error),
-        }
-    }
-}
-
-pub type Result<A, E> = std::result::Result<A, E>;
```
tusharmath, line 159: Instead of creating a module for errors, we should create mods for cache and worker and move the errors into that module.

tusharmath (approved):

PR #2522, opened 2024-07-25
fix: handle field of list types in case of non-ir field.

laststylebender14:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/core/jit/exec.rs
+++ b/src/core/jit/exec.rs
@@ -126,7 +126,6 @@ where
         // Has to be an Object, we don't do anything while executing if its a Scalar
         else {
             join_all(field.nested_iter().map(|child| {
-                let value = if value.is_some() { value } else { ctx.value() };
                 let value = value.map(|v| v.get_key(&child.name).unwrap_or(v));
                 let ctx = if let Some(v) = value {
                     ctx.with_value_and_field(v, child)


src/core/jit/exec.rs
```diff
@@ -97,6 +97,51 @@ where
...
+            }
+            // TODO:  We should throw an error stating that we expected
+            // a list type here but because the `Error` is a
+            // type-parameter, its not possible
+        }
+        // TODO: Validate if the value is an Object
+        // Has to be an Object, we don't do anything while executing if its a Scalar
+        else {
+            join_all(field.nested_iter().map(|child| {
+                let value = if value.is_some() { value } else { ctx.value() };
```
meskill, line 129: this line is redundant because of the way you're calling this function from `exec`

src/core/jit/exec.rs
```diff
@@ -147,19 +164,8 @@ where
...
-                let ctx = if let Some(v) = value {
-                    ctx.with_value_and_field(v, child)
-                } else {
-                    ctx.with_field(child)
-                };
-                let data_path = data_path.clone();
-                async move { self.execute(child, &ctx, data_path).await }
-            }))
-            .await;
+            self.execute_inner(field, ctx, data_path.clone(), ctx.value())
```
meskill, line 167: seems like we handling additional case when there is not IR and the value is a list. We should add an integration test

src/core/jit/exec.rs
```diff
@@ -97,6 +97,51 @@ where
...
+        ctx: &'b Context<'b, Input, Output>,
+        data_path: DataPath,
+        value: Option<&'b Output>,
+    ) -> Result<(), Error> {
+        if field.type_of.is_list() {
+            // Check if the value is an array
+            if let Some(array) = value.and_then(|v| v.as_array()) {
+                join_all(field.nested_iter().map(|field| {
+                    join_all(array.iter().enumerate().map(|(index, value)| {
+                        let new_value = value.get_key(&field.name).unwrap_or(value);
```
meskill, line 113: could you remind me, why do we need `get_key` here and why in case of missing entry in value it's ok to use just the value itself?

as I understand, value that we set in the context is later used in an IR and that requires actually parent value to the IR be resolved properly. But here it could contain different data actually

meskill (changes requested):

src/core/jit/exec.rs
meskill: let's take closer look to this example:

when we iterating over field `b` inside type `Bar` you see an IR that contains `value.x`. This `x` comes from the data of type `Bar` and since the IR resolution happens at the beginning of the exec when iterating the fields of Bar, we need to store the Bar itself in the context, so on next recursive call the IR will get the `Bar` type and its `x` field.

What happens in your code that you're trying to acquire `Bar.b` and set the value into the context. And if for some reason the value `Bar.b` is present than you'll set some float value in the context and when trying to resolve an IR you'll get non-existent `x` value. I think it's not that destructive right now only because in most cases when we have IR for field the field itself is not present in the parent and therefore your logic `unwrap_or(value)` sets the parent value back to context and everything just works.

This issue is actually broke many tests for https://github.com/tailcallhq/tailcall/pull/2452/files with cryptic messages because I rely a lot on the value itself in those.

In conclusion, you should not try to set the value of the field in the context after resolving the value from IR. But you need to go one level deeper for field when processing non-IR branch since the value should change in that case. But this should not be for the child field, but for the currently processed field so the invariant that we're storing actually the parent value in the context is hold.

Issue #2483, opened 2024-07-21
Support `@modify` on inputs


tusharmath:
Context Discussion - https://github.com/tailcallhq/tailcall/discussions/2476

The [@modify](https://tailcall.run/docs/tailcall-dsl-graphql-custom-directives/#modify-directive) only works on output types. We need to support `@modify` on input types as well

**Proposal**

```graphql
type Query {
  ask(question: Question): Answer
}

type Answer {
  text: [String]
}

input Question {
  prompt: String
  model: String
  system: String @modify(name: "systemPrompt")
}
```

The generated client schema should look like:

```graphql
type Query {
  ask(question: Question): Answer
}

type Answer {
  text: [String]
}

input Question {
  prompt: String
  model: String
  systemPrompt: String # Field is renamed
}
```

Internally Tailcall should handle these field name conversions efficiently.

**Technical Requirements**
- Ensure both unit and integration tests are added wherever necessary.

bnchi:
The current behavior seem to be correct ?!

Current config

```graphql
schema
  @server(port: 8000)
  @upstream(baseURL: "http://localhost:3000", httpCache: 42, batch: {delay: 100}) {
  query: Query
}

type Query {
  searchUser(input: UserSearch): [User]!
    @http(
      method: POST
      path: "/users/search"
      body: "{{.args.input}}"
    )
}

input UserSearch {
  emailAddress: String! @modify(name: "email_address")
}

type User {
  id: Int!
  name: String!
  email: String!
}
```

The generated client schema for _UserSearch_ from the playground : 
```graphql
Input UserSeach {
 email_address: String!
}```

angristan:
Ok, let me try to make it more clear 🙏  Keep in mind that the issue I'm seeing is that I want to have consistency in the casing of my schema. Having mixed snake case and camel case because of the underlying resolvers is not ideal IMO.

First, `@modify` doesn't seem to be allowed on inputs:

<img width="1021" alt="image" src="https://github.com/user-attachments/assets/2a82396c-c939-4998-90f0-73ffcf3db82e">

But it does kinda work.

## `emailAddress: String! @modify(name: "email_address")`

Example schema:

```gql
schema @server(port: 8000) @upstream(baseURL: "http://localhost:8008") {
  query: Query
}

type Query {
  searchUser(input: UserSearch): [User]!
    @http(method: POST, path: "/users/search", body: "{{.args.input}}")
}

input UserSearch {
[...]

PR #2524, opened 2024-07-25
feat(2483): Add @modify to work on input fields

karatakis:
**Summary:**  
Make `@modify` work for input fields

**Issue Reference(s):**  
Fixes https://github.com/tailcallhq/tailcall/issues/2483

**Build & Testing:**

**Checklist:**


karatakis:
To make it better integrated with IR infrastructure, I propose to replace `pub renames: HashMap<Vec<String>, String>` with pub renames: HashMap<Vec<String>, Option<IR>>` and collect the IR of all subfields of an input type, with the addition of a field `pub resolver: Option<IR>` for the self operations. All the changes should be performed at the `InputFieldDefinition` struct.

karatakis:
After playing with the IR approach, it needs a custom `InputFieldEvaluator`. Feel free to comment if we want this approach instead of the hardcoded solution I provided.

src/core/blueprint/blueprint.rs
```diff
@@ -154,6 +154,7 @@ pub struct SchemaDefinition {
 pub struct InputFieldDefinition {
     pub name: String,
     pub of_type: Type,
+    pub renames: HashMap<Vec<String>, String>,
```
tusharmath, line 157: Blueprint should have the final state of the schema.

src/core/blueprint/mustache.rs
```diff
@@ -209,6 +211,7 @@ mod test {
                 of_type: type_,
                 default_value: None,
                 description: None,
+                renames: HashMap::new(),
```
tusharmath, line 214: Mustache doesn't care about renames. It's a general-purpose templating DSL.

src/core/blueprint/into_schema.rs
```diff
@@ -100,13 +102,55 @@ fn to_type(def: &Definition) -> dynamic::Type {
...
+                            for arg in &field_copy.args {
+                                let renames = &arg.renames;
+                                let arg_key = Name::new(&arg.name);
+                                if let Some(data) = index_map.get(&arg_key) {
+                                    let data = perform_rename(vec![], data, renames);
+                                    new_map.insert(arg_key, data);
+                                }
+                            }
+                            ConstValue::Object(new_map)
+                        };
```
tusharmath, line 146: This logic needs to move into `IR` under a special variant perhaps - `IR::ModifyArgs`

src/core/blueprint/blueprint.rs
```diff
@@ -154,6 +154,7 @@ pub struct SchemaDefinition {
 pub struct InputFieldDefinition {
     pub name: String,
     pub of_type: Type,
+    pub resolver: Option<IR>,
```
tusharmath, line 157: Inputs don't need to have resolvers. Even in the implementation in this PR, you are piping the IR to the resolvers of the fields in the output type.

tusharmath (changes requested):

src/core/blueprint/blueprint.rs
karatakis: Removed, thanks for the insight.

src/core/blueprint/into_schema.rs
```diff
@@ -102,24 +102,33 @@ fn to_type(def: &Definition) -> dynamic::Type {
 
                         let req_ctx = ctx.ctx.data::<Arc<RequestContext>>().unwrap();
                         let field_name = &field.name;
+                        let input_resolvers = field_copy
+                            .args
+                            .iter()
+                            .filter_map(|input_argument| input_argument.resolver.clone())
+                            .reduce(|first, second| first.pipe(second));
```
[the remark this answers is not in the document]
karatakis: Implemented correctly. Thanks for the insight.

src/core/ir/model.rs
```diff
@@ -108,6 +109,14 @@ impl Cache {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct InputTransforms {
+    pub subfield_types: HashMap<(String, String), String>,
+    pub subfield_renames: HashMap<(String, String), String>,
+    pub arg_name: String,
+    pub arg_type: String,
+}
```
tusharmath, line 118: I think this should be a recursive structure, something like:

```rust
pub enum Transform {
    Object(HashMap<FieldName, Transform>),
    Rename {from: String, to: String}
}
```

src/core/ir/eval.rs
```diff
@@ -95,6 +99,97 @@ impl IR {
...
+                                    (name.clone(), value.clone())
+                                }
+                            })
+                            .collect();
+
+                        fn handle_args<'a>(
+                            args: &'a ConstValue,
+                            input_transforms: &'a InputTransforms,
+                            type_of: &'a str,
+                        ) -> Cow<'a, ConstValue> {
```
tusharmath, line 129: This function should be implemented inside of `JsonLike`. It should be a default implementation.

src/core/blueprint/input_field_resolver.rs
```diff
@@ -0,0 +1,107 @@
...
+                    (None, Some(input_resolvers)) => Some(input_resolvers),
+                    (Some(field_resolver), None) => Some(field_resolver),
+                    (Some(field_resolver), Some(input_resolvers)) => {
+                        Some(input_resolvers.pipe(field_resolver))
+                    }
+                };
+            };
+
+            Valid::succeed(b_field)
+        },
```
tusharmath, line 105: Its a lot of code to read, lets add some documentation.

src/core/ir/model.rs
karatakis: I refactored it to document it better with code. There is no need to introduce a complex recursive type because I only go up to 1 level deep. That's why it is a (String, String) (type name, field name) => new field name. I use a hashmap for performance gains to check the presence of a rename for a field.

src/core/json/json_like.rs
```diff
@@ -41,6 +43,66 @@ pub trait JsonLike<'json>: Sized {
     fn get_path<T: AsRef<str>>(&'json self, path: &[T]) -> Option<&Self>;
     fn get_key(&'json self, path: &str) -> Option<&Self>;
     fn group_by(&'json self, path: &[String]) -> HashMap<String, Vec<&Self>>;
+
+    ///
+    /// Used to apply `input_transforms` operations on the given `json_like`
+    /// struct. Current supported operations are `modify` - rename.
+    fn handle_input_transforms(
```
tusharmath, line 50: Design Feedback: JsonLike shouldn't make assumptions about how it's being used, in this case — `handle_input_transforms` function has the following problems:
- Depends on a transformation that's defined in IR::model. 
- Only works one level deep.
- Only renames keys, so isn't a general purpose transform.

tusharmath: Try something like — 

```rust
enum Transform {
    
    Rename(String) // Renames the provided key in the object
    // TODO: add more operations as time goes on
}


enum Lens {
    Select(String), // Selects the value at the provided key
    ForEach(Rc<Lens>), // Applies the provided lens to each value in 
    Compose(Rc<Lens>, Rc<Lens>), // Composes two lenses together
    Transform(Rc<Lens>, Transform), // Applies a transformation on the lens
}

impl Lens {
    fn transform<'json>(self, J: &mut JsonLike<'json>) -> J {
        todo!()
    }
}
```

karatakis: * Better organized the codebase
* Added a more complex (deep nested) test case to prove that nesting is working on deep nested objects
* Refactored the code to allow for other operations in the future easily

karatakis:
It is awaiting review

src/core/json/json_like.rs
```diff
@@ -5,7 +5,7 @@ pub trait JsonLikeOwned: for<'json> JsonLike<'json> {}
 impl<T> JsonLikeOwned for T where T: for<'json> JsonLike<'json> {}
 
 /// A trait for objects that can be used as JSON values
-pub trait JsonLike<'json>: Sized {
+pub trait JsonLike<'json>: Sized + Clone {
```
tusharmath, line 8: trait doesn't need to have any clone requirements.

tusharmath:
@karatakis Like discussed, we should more the lens capabilities out of the core transformation capabilities.

PR #2538, opened 2024-07-26
fix: query parameter selection when `batchKey` is used

karatakis:
**Summary:**  
Fix documentation and test correctness to prevent misleading of users.

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/core/config/config.rs
+++ b/src/core/config/config.rs
@@ -590,8 +590,9 @@ pub struct Http {
     /// This represents the query parameters of your API call. You can pass it
     /// as a static object or use Mustache template for dynamic parameters.
     /// These parameters will be added to the URL.
-    /// When `batchKey` is present Tailcall uses the first query parameter as
-    /// the key for the groupBy operator.
+    /// When `batchKey` is present Tailcall uses the last query parameter as
+    /// the key for the groupBy operator. You can overwrite this behavior by
+    /// specifying the `key` property.
     pub query: Vec<KeyValue>,
 }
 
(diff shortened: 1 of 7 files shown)


src/core/config/config.rs
```diff
@@ -589,8 +589,9 @@ pub struct Http {
     /// This represents the query parameters of your API call. You can pass it
     /// as a static object or use Mustache template for dynamic parameters.
     /// These parameters will be added to the URL.
-    /// When `batchKey` is present Tailcall uses the first query parameter as
-    /// the key for the groupBy operator.
+    /// When `batchKey` is present Tailcall uses the last query parameter as
+    /// the key for the groupBy operator. You can overwrite this behavior by
+    /// specifying the `key` property.
```
tusharmath, line 594: What is the `key` property you are talking about here?

karatakis: query: [{key: "id", value: "{{.value.userId}}"}]

The `key` from the `query` setting of `@http` directive.

It still needs to be clarified so the users can understand better. Thanks for pointing this out; any suggestions?

tusharmath: How about:


>Query parameter order is critical for batching in Tailcall. The first parameter referencing a field in the current value using mustache syntax is automatically selected as the batching parameter. Arrange parameters carefully to ensure desired batching behavior.

tusharmath (approved):

amitksingh1490 (approved):

PR #2547, opened 2024-07-27
fix: disable user from adding predefined scalars

### LINKED ISSUE #2544: Validation error should be given when config overrides a predefined scalar
The start/check command should throw a validation error if config overrides a scalar which is predefined
eg: 
```graphql
scalar JSON
type T {
 a: JSON
}
```

If the config tries to redefine or override the JSON scalar, it should result in a validation error. Currently, this does not happen and can lead to unexpected behaviors.

**Expected Behavior:**

- The start/check command should detect attempts to override predefined scalars.
- A clear validation error message should be thrown, indicating the scalar that is being overridden.

ssddOnTop:
**Issue Reference(s):**  
Fixes #2544
/claim 2544


src/core/blueprint/definitions.rs
```diff
@@ -14,14 +14,18 @@ use crate::core::valid::{Valid, Validator};
...
+    }else {
+        Valid::succeed(Definition::Scalar(ScalarTypeDefinition {
+            name: name.to_string(),
+            directive: Vec::new(),
+            description: None,
+            scalar: scalar::Scalar::find(name)
+                .unwrap_or(&scalar::Scalar::Empty)
+                .clone(),
+        }))
+    }
```
tusharmath, line 28: Missing tests. Please add an integration test.

tusharmath (approved):

Issue #2200, opened 2024-06-15
Write a transformer to create better type names for configurations using LLM


tusharmath:
Currently autogenerated configurations can sometimes have not-so-readable type names. We want to use LLMs to produce more meaningful type-names

**Before**

```graphql
type T1 {
  name: String
  age: Int
  gender: Gender
}

type T2 {
  name: String
  address: Address
  reviews: [Review]
}
```

**After**
A simple instruction like below to chat gpt can produce a more meaningful configuration

>Generate better names for T1 and T2 and produce a new schema file with the new names.

```graphql
type Person {
  name: String
  age: Int
  gender: Gender
}

type Business {
  name: String
  address: Address
  reviews: [Review]
}
```

**Technical Requirements**
- It should work with very large configurations.
- It should ensure there are no conflicts.
- No information should be lost, such as comments, directives etc.
- Should be able to plug any LLM API

**Additional Information**
- Transformers are written over here: 	https://github.com/tailcallhq/tailcall/blob/main/src/core/config/transformer/improve_type_names.rs
- Start Locally by using Ollama: https://ollama.com/
- Use APIs documented here https://github.com/ollama/ollama/blob/main/docs/api.md

**Proposed Solution**
- Update `improve_type_names` transformer to also take in additional "hints" for each type name. 
- The hints are going to be in the form of:
	```rust
	type OriginalName: String; // Like T1
	type AIGeneratedName: String; // Like Business or Person
	HashMap<OriginalName, Vec<AIGeneratedName>>
	```
- At the time of name resolution or a name conflict, if a name hint is provided that name is going to be preferred over its own name resolution algorithm.

**Sample Prompt for OLLAMA**

```json
{
  "model": "llama3",
  "system": "Given the GraphQL type definition below, provide a response in the form of a JSONP callback. The function should be named \"callback\" and should return JSON suggesting at least five suitable alternative names for the type. Each suggested name should be concise, preferably a single word, and capture the essence of the data it represents based on the roles and relationships implied by the field names. \n\n```graphql\ntype T {\n  name: String,\n  age: Int,\n  website: String\n}\n```\n\n**Expected JSONP Format:**\n\n```javascript\ncallback({\n  \"originalTypeName\": \"T\",\n  \"suggestedTypeNames\": [\"Person\",\"Profile\",\"Member\",\"Individual\",\"Contact\"\n  ]\n});\n```",
  "prompt": "type T {title: String, body: String, userId: Int}"
}
```

[...]

PR #2578, opened 2024-07-30
feat: LLM generated type names

beelchester:
fixes: #2200 
/claim #2200

Note: LLM service needs to be deployed.


src/core/config/transformer/improve_type_names/improve_type_names_llm.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+impl ImproveTypeNamesLLM {
+    pub fn generate_llm_type_names(
+        config: Config,
+    ) -> Result<HashMap<OriginalTypeName, Vec<AIGeneratedTypeName>>, String> {
+        let mut llm_requests: Vec<LLMRequest> = Vec::new();
+
+        for (type_name, type_info) in config.types.iter() {
+            if type_name == "Query" || type_name == "Mutation" || type_name == "Subscription" {
```
laststylebender14, line 30: can use `is_root_operation_type` method

src/core/config/transformer/improve_type_names/improve_type_names_llm.rs
```diff
@@ -0,0 +1,110 @@
...
+        let mut llm_requests: Vec<LLMRequest> = Vec::new();
+
+        for (type_name, type_info) in config.types.iter() {
+            if config.is_root_operation_type(type_name) {
+                continue;
+            }
+
+            let mut fields: Vec<String> = Vec::new();
+            for (field_name, field_info) in type_info.fields.iter() {
+                fields.push(format!("{}: {}", field_name, field_info.type_of));
```
laststylebender14, line 36: what if present field type is of type list?

src/core/config/transformer/improve_type_names/improve_type_names.rs
```diff
@@ -114,22 +115,47 @@ impl<'a> CandidateGeneration<'a> {
...
-    /// Generates type names based on inferred candidates from the provided
-    /// configuration.
     fn generate_type_names(&self, mut config: Config) -> Config {
+        let ai_generated_type_names = ImproveTypeNamesLLM::generate_llm_type_names(config.clone());
         let finalized_candidates = CandidateGeneration::new(&config).generate().converge();
+        let mut llm_worked = false;
 
         for (old_type_name, new_type_name) in finalized_candidates {
+            // Try to use the llm generated type names first
             if let Some(type_) = config.types.remove(old_type_name.as_str()) {
```
laststylebender14, line 125: i think we should use this in combination with auto infer but assign the LLM generated name bit higher priority.

laststylebender14: what are we doing here is, we are removing the type from config and performing validations.
eg. what would happen if we remove the type from config buton line no 127 it's None instead of Some.

laststylebender14: another problem i see is we are doing same checks for each candidate.

src/core/config/transformer/improve_type_names/improve_type_names.rs
```diff
@@ -114,22 +115,47 @@ impl<'a> CandidateGeneration<'a> {
...
-                        if actual_field.type_of == old_type_name {
-                            // Update the field's type with the new name
-                            actual_field.type_of.clone_from(&new_type_name);
+                if let Ok(ai_generated_types) = &ai_generated_type_names {
+                    if let Some(ai_generated_types) = ai_generated_types.get(&old_type_name) {
+                        // ensure the type name is not already present in the config, else try
+                        // the next llm generated type name
+                        for type_name in ai_generated_types.iter() {
+                            if !config.types.contains_key(type_name) {
+                                llm_worked = true;
```
laststylebender14, line 132: not sure why this flag needed? we make it true only once.

src/core/config/transformer/improve_type_names/improve_type_names.rs
```diff
@@ -114,22 +115,47 @@ impl<'a> CandidateGeneration<'a> {
...
+                                config.types.insert(type_name.to_owned(), type_.clone());
```
laststylebender14, line 133: think about other case, we remove the type from config and we never enter here. i.e all suggested type names are already present in the config.

src/core/config/transformer/improve_type_names/improve_type_names.rs
```diff
@@ -114,22 +115,47 @@ impl<'a> CandidateGeneration<'a> {
...
+                                // Replace all the instances of old name in config.
+                                for actual_type in config.types.values_mut() {
+                                    for actual_field in actual_type.fields.values_mut() {
+                                        if actual_field.type_of == old_type_name {
+                                            // Update the field's type with the new name
+                                            actual_field.type_of.clone_from(type_name);
+                                        }
```
laststylebender14, line 140:
```rust
input PostInput {
    ...
}

post(postInput: PostInput): Post
```

we use AI to generate the name of PostInput, then we'll end up renaming the input type but not in the args. so should also iterate over args present in the field.

src/core/config/transformer/improve_type_names/improve_type_names.rs
beelchester: Hi, thanks for the review.

> i think we should use this in combination with auto infer but assign the LLM generated name bit higher priority.

Isn't it working the same though? 🤔

If anything llm related fails between lines 126 to 143 (in the latest commit)... It will use inferred type names. That's why I have used the flag that is turned true only if llm successfully generated type names + there are no duplications... or else inferred type name is used like before this PR.

src/core/config/transformer/improve_type_names/improve_type_names.rs
beelchester: Again, if it fails it will use the inferred type name.

src/core/config/transformer/improve_type_names/improve_type_names.rs
beelchester: Sure will try. But in this PR, AI is only updating the types that the inference would have updated.

PR #2585, opened 2024-07-31
test: Add GraphQL conformance tests

karatakis:
**Summary:**  
Use [GraphQ spec](https://spec.graphql.org/October2021/) and write conformance tests for Tailcall.

**Build & Testing:**

**Checklist:**


tests/core/snapshots/graphql-conformance-002.md_0.snap
```diff
@@ -0,0 +1,39 @@
...
+source: tests/core/spec.rs
+expression: response
+---
+{
+  "status": 200,
+  "headers": {
+    "content-type": "application/json"
+  },
+  "body": {
+    "data": {
```
meskill, line 11: instead of recreating right snapshots here we can mark failed tests as skip with relevant comment. And for that test we can create an issue to fix this and enable test

tests/execution/graphql-conformance-004.md
```diff
@@ -0,0 +1,49 @@
...
+- method: POST
+  url: http://localhost:8080/graphql
+  body:
+    query: |
+      {
+        user(id: 4) {
+          id
+          name
+          smallPic: profilePic(size: 64)
+          bigPic: profilePic(size: 1024)
```
meskill, line 47: this test is failing because aliases are not handled at all at selection_set for the graphql upstream. It applies both for async_graphql and jit implementation

tests/execution/graphql-conformance-005.md
```diff
@@ -0,0 +1,46 @@
...
+```yml @test
+- method: POST
+  url: http://localhost:8080/graphql
+  body:
+    query: |
+      {
+        customer: user(id: 4) {
+          id
+          name
+          date_of_birth: dob
```
meskill, line 44: seems like it fails also because of the aliases not working as in the previous test

tests/execution/graphql-conformance-006.md
```diff
@@ -0,0 +1,177 @@
...
+        }
+      }
+- method: POST
+  url: http://localhost:8080/graphql
+  body:
+    query: |
+      query withFragments {
+        user(id: 4) {
+          friends(first: 10) {
+            ...friendFields
```
meskill, line 140: fragments are not supported in this moment for graphql upstream. I'm not even sure if we can do this propely with current async_graphql api. For jit we need to desing actually how to handle fragments for the graphql upstream

tests/execution/graphql-conformance-007.md
```diff
@@ -0,0 +1,83 @@
...
+```yml @test
+- method: POST
+  url: http://localhost:8080/graphql
+  body:
+    query: |
+      query FragmentTyping {
+        profiles(handles: ["user-1", "user-2"]) {
+          handle
+          ...userFragment
+          ...pageFragment
```
meskill, line 67: this also failing because of the fragments I believe

karatakis: Yeah, we don't send them over to the target graphql server. Maybe we might need an issue to fix that

tests/execution/graphql-conformance-006.md
karatakis: There is no need to support it at the moment. I just documented the behavior so we can track the functionality that is missing

tests/execution/graphql-conformance-001.md
```diff
@@ -0,0 +1,104 @@
+```graphql @config
```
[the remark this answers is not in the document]
karatakis: Noted, I added names that describe what is the test about on each file

tests/core/snapshots/graphql-conformance-002.md_0.snap
karatakis: That is a good strategy to know which stuff are not currently implemented

tests/execution/graphql-conformance-004.md
karatakis: Noted.

tests/execution/graphql-conformance-005.md
karatakis: Correct

tests/execution/graphql-conformance-002.md
```diff
@@ -0,0 +1,76 @@
...
+type Query {
+  user(id: ID!): User! @graphQL(name: "user", args: [{key: "id", value: "{{.args.id}}"}])
+}
+
+type User {
+  id: ID!
+  name: String!
+  city: String
+  birthday: BirthDay!
+  friends: [User!]!
```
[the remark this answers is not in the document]
karatakis: Yeah, it works okay for JIT, I tested it

tusharmath (approved):

meskill:
I think we still missing some tests for some graphql features (correct me if I'm wrong):
- use interfaces and unions as fragment `on` type
- default values for arguments including default values for nested fields in input type
- validation tests for negative scenarios. So basically for every feature we're testing we should have several negative cases scenarios

karatakis:
@meskill 

- fragments on unions are implemented
- default input values are tested
- All negative paths are failing because we have validation disabled. So the current behavior is ignoring the field altogether

tests/execution/graphql-conformance-012.md
```diff
@@ -0,0 +1,164 @@
...
+      }
+      fragment photoFragment on Photo {
+        height
+        ...metaFragment
+      }
+      fragment metaFragment on photo {
+        meta {
+          iso
+        }
+      }
```
meskill, line 126: is this wrong fragment?

tests/execution/graphql-conformance-012.md
```diff
@@ -0,0 +1,164 @@
...
+
+# Negative: missing fragment
+# - method: POST
```
meskill, line 129: why comment them?

this test is skipped anyway

meskill:
> @meskill
> 
> * fragments on unions are implemented
> * default input values are tested
> * All negative paths are failing because we have validation disabled. So the current behavior is ignoring the field altogether

- what about fragments on interfaces and the union itself? I think we can specify fragments not only on concrete type but on the interface and union itself
- I see default inputs but input could be also a complex object that could contain default fields
- what validation you're talking about? Many validations comes from async_graphql out of the box, like required fields, query validation against schema etc.

meskill:
I found additional case missing - list inside lists. See https://github.com/graphql/graphql-spec/issues/749. It doesn't work in both implementations

karatakis:
@meskill 
- added fragment example (it fails, so it is disabled)
- complex input with optional is achieved
- implemented many negative scenarios (but because we have the validation from async_graphql disabled for performance reasons, they are failing)
- List of lists example is implemented

This is an excellent start to fix many issues regarding the complete GraphQL spec. There is still room for improvement, but the tests highlight the rough edges.

meskill (changes requested):

karatakis:
@meskill 
- added tests for an input field with an array
- added test for an input type that defines default values
- added descriptions of why tests are disabled

tests/execution/graphql-conformance-002.md
```diff
@@ -0,0 +1,109 @@
+---
+skip: true
+---
+
+# Test complex nested query. Skipped because Tailcall does not send the whole query to the remote server. It sends a shallow version of the query.
```
meskill, line 5:
```suggestion
# Test complex nested query. 

TODO: Skipped because Tailcall does not send the whole query to the remote server. It sends a shallow version of the query.
```

Maybe format like this? In my opinion it's more visible and allows to provide more info

tests/execution/graphql-conformance-http-002.md
```diff
@@ -0,0 +1,103 @@
...
+            name
+            birthday {
+              year
+            }
+          }
+        }
+      }
+
+# Negative: invalid selection at nested
+# Disabled because async_graphql::dynamic does not perform validation
```
meskill, line 79:
```suggestion
# TODO: Disabled because async_graphql::dynamic does not perform validation
```

also to increase visibility

tests/execution/graphql-conformance-http-007.md
```diff
@@ -0,0 +1,84 @@
+---
+skip: true
+---
+
+# Test named fragments. Skipped because Tailcall does not construct correctly the interface types based on \_\_typename field.
```
meskill, line 5: I think it's not working because we not using our discriminator https://github.com/tailcallhq/tailcall/blob/f464794bd32abe3224eb96e1de09d6fccd766148/src/core/blueprint/union_resolver.rs#L34 for interfaces and only for unions. The upstreams in most cases won't provide `__typename` info for us anyway and we need to figure out the type on our side, that's the discrininator job

tests/execution/graphql-conformance-http-007.md
```diff
@@ -0,0 +1,84 @@
...
+- request:
+    method: GET
+    url: http://upstream/profiles?handles=user-1&handles=user-2
+  expectedHits: 1
+  response:
+    status: 200
+    body:
+      - id: 1
+        handle: user-1
+        __typename: User
```
meskill, line 50: upstream won't send `__typename` in most cases, especially rest upstream

meskill: could you also check other tests for this? instead of relying on `__typename` from the upstream Tailcall should be able to figure the type on it's own

tests/execution/graphql-conformance-http-012.md
```diff
@@ -0,0 +1,139 @@
...
+        ...metaFragment
+      }
+      fragment metaFragment on Photo {
+        meta {
+          iso
+        }
+      }
+
+# Negative: missing fragment
+# Disabled because async_graphql::dynamic does not perform validation
```
meskill, line 102: could you expand on what validation is expected in those cases?

tests/execution/graphql-conformance-015.md
```diff
@@ -0,0 +1,185 @@
...
+# Disabled because Tailcall does not uses default values from Input Object
+# - method: POST
+#   url: http://localhost:8080/graphql
+#   body:
+#     query: |
+#       query {
+#         user(id: 4) {
+#           id
+#           name
+#           featuredVideoPreview
```
meskill, line 168: is this test right?

the field set args as required and we need to provide that arg (or that is validation missed case?) and we need to test that `hdr` is got defaulted to `true`

tests/execution/graphql-conformance-http-015.md
```diff
@@ -0,0 +1,182 @@
...
+    @expr(body: "video_{{.value.id}}_{{.args.video.width}}_{{.args.video.height}}_{{.args.video.hdr}}")
+  featuredVideoPreview(video: VideoSize!): String!
+    @expr(body: "video_{{.value.id}}_{{.args.video.width}}_{{.args.video.height}}_{{.args.video.hdr}}")
+  searchComments(query: [[String!]!]! = [["today"]]): String! @expr(body: "video_{{.value.id}}_{{.args.query}}")
+}
+
+input VideoSize {
+  width: Int!
+  height: Int!
+  hdr: Boolean
```
meskill, line 29: this test differs from its companion test for graphql because there is no default values inside input type

meskill (changes requested):

tests/execution/graphql-conformance-http-015.md
karatakis: good catch :+1:

tests/execution/graphql-conformance-http-007.md
karatakis: I just removed the __typename. This test can test this behavior

tests/execution/graphql-conformance-015.md
karatakis: changed it to `featuredVideoPreview(video: VideoSize! = {}): String!`. I set up another graphql implementation and verified that hdr is true. I disable the test for now.

meskill (approved):

PR #2609, opened 2024-08-03
chore: introduce include_config

### LINKED ISSUE #2608: chore: Implement `include_config!` macro
A special macro that works like `include_str!` macro but apart from including the file path as string, this macro also parses the string and creates a configuration. 

**Technical Requirements**
- It should be available in test mode only.
- It should produce a `Result<Config, String>`

ssddOnTop:
**Issue Reference(s):**  
Fixes #2608
/claim 2608

### DIFF
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -108,7 +108,7 @@ pub mod tests {
     #[macro_export]
     macro_rules! include_config {
         ($file:literal) => {{
-            use crate::core::config::{Config, Source};
+            use $crate::core::config::{Config, Source};
             let config_str = include_str!($file);
             let result = || {
                 let source = Source::detect($file)?;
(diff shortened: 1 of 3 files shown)


src/core/mod.rs
```diff
@@ -125,4 +138,9 @@ pub mod tests {
             Self(HashMap::from_iter(iter))
         }
     }
+    #[test]
+    fn test_include_config() {
+        let cfg = include_config!("fixtures/helloworld.graphql");
+        assert!(cfg.is_ok());
```
tusharmath, line 144: convert to snapshot.

tusharmath (changes requested):

src/core/mod.rs
```diff
@@ -105,6 +105,19 @@ pub mod tests {
...
+    #[macro_export]
+    macro_rules! include_config {
+        ($file:literal) => {{
+            use $crate::core::config::{Config, Source};
+            let config_str = include_str!($file);
+            let result = || {
+                let source = Source::detect($file)?;
+                Config::from_source(source, config_str)
+            };
+            result().map_err(|e| e.to_string())
```
tusharmath, line 117: just return result instead of converting to string.

tusharmath (approved):

PR #2626, opened 2024-08-07
refactor: store __typename in value instead of separate entity

meskill:
**Summary:**  
Refactor the way the value of `__typename` is stored along the data value itself. Usually, `__typename` is already provided by `async_graphql` library, but in some cases we need to resolve it on our side, like for union types.

Previously, special wrapper type `TypeName` was used to resolve this type and it was stored in the `EvaluationContext` and propagated to the async_graphql/jit implementation either through context or other wrapper types to implement the fragments logic and to provide field `__typename` itself.

Because field `__typename` is actually part of the values in graphQL I think it's reasonable to store that type info inside the property `__typename` for values we have (currently, only for union types since for other cases the typename could be inferred from the schema).  

To implement this change, following code was changed:
- extend JsonLike to make it easier to set the new fields to resolved values
- trait TypedValue to abstract the typename operations on value
- remove additional code and wrappers from graphql engines and utilize newly added trait

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


karatakis (approved):
Looks okay

src/core/json/json_like.rs
```diff
@@ -41,6 +44,29 @@ pub trait JsonLike<'json>: Sized {
     fn get_path<T: AsRef<str>>(&'json self, path: &[T]) -> Option<&Self>;
     fn get_key(&'json self, path: &str) -> Option<&Self>;
     fn group_by(&'json self, path: &[String]) -> HashMap<String, Vec<&Self>>;
+
+    fn map<Err>(self, mut mapper: impl FnMut(Self) -> Result<Self, Err>) -> Result<Self, Err> {
```
tusharmath, line 48: I think this method solves a very specific problem in a very specific way and doesn't need to be placed in JsonLike.

meskill: I moved this to another trait since it could re reused in other places later e.g. jit batching

karatakis (approved):

Issue #2613, opened 2024-08-04
feature: Custom Extensions


tusharmath:
This feature should allows developers to write custom Rust functions to extend Tailcall's ability.
Here is how the user experience going to look like:

## Example
Let's say we want to implement langugage translation in Tailcall.

The extension is loaded using the link operator
```graphql
schema
  # Dynamically Link an Extension
  @link(src: "./translate", type: Extension)
  @upstream(baseURL: "https://jsonplaceholder.typicode.com") {
  query: Query
}

type Query {
  users: [User] @http(path: "/users")
  user(id: String): [User]
    @http(path: "/users")
    # User can use the @extension directive
    @extension(name: "some_other_extension", params: ["{{.args.id}}"])
}

type User {
  id: ID
  name: String
    @extension(
      name: "translate"
      params: ["userName", "{{.value.catchPhrase}}"]
    )
  username: String
  company: Company
}

type Company {
  name: String
  catchPhrase: String
    @extension(
      name: "translate"
      params: ["catchPharse", "{{.value.catchPhrase}}"]
    )
  bs: String
}
```

**Directive Definition**
Should be added automatically to the `.tailcallrc.graphql`
```graphql
directive @extension(name: String!, params: [String!]!) on FIELD
```

**Rust Code**

The extension needs to implement the following trait:

```rust
trait Extension {
    // Initializes the extension
    async fn init(&self);

    // Called before the IR is going to be executed
    fn prepare<Json: JsonLikeOwned>(&self, ir: IR, params: &[Json]) -> IR;

    // Called after a value is produced
    fn process<Json: JsonLikeOwned>(&self, params: &[Json], value: Json) -> Json;
}
```

The trait is exposed via `tailcall::extensions` module that published as a crate. The extension that implements the trait Extension should loaded dynamically into Tailcall.

**Technical Requirements**
- Add tests and documentation wherever possible
- Use lib loading should be async and failures should be nicely displayed to the user

PR #2634, opened 2024-08-09
feat(extensions): Add extensions support

karatakis:
**Summary:**  
Modify Tailcall to allow users to create a version with custom extension functionality. We achieve that by providing a `@extension` directive and an extension trait that developers must implement on their custom extensions before tailcall can load them. An example that showcases the power of extensions is included.

**Issue Reference(s):**  
Fixes https://github.com/tailcallhq/tailcall/issues/2613

**Build & Testing:**
To test it run the following command `cargo run -p extension-i18n` and execute the query

```gql
{
  user(id: 1) {
    id
    name
    company {
      catchPhrase
    }
  }
}
```

**Checklist:**


src/core/blueprint/operators/extension.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+pub trait ExtensionLoader: std::fmt::Debug + Send + Sync {
+    // TODO: signature is not the desired one
+    fn load(&self);
+
+    // TODO: signature is not the desired one
+    fn prepare(&self, ir: Box<IR>, params: ConstValue) -> Box<IR>;
+
+    // TODO: signature is not the desired one
```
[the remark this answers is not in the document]
meskill: you can't use generic method inside trait object. We should redesign the interface.

smth like this could work `pub trait ExtensionLoader<Value: JsonLikeOwned>`

examples/extension-i18n/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+        println!("TranslateExtension...loaded!")
+    }
+
+    fn prepare(
+        &self,
+        ir: Box<tailcall::core::ir::model::IR>,
+        params: ConstValue,
+    ) -> Box<tailcall::core::ir::model::IR> {
+        // TODO: implement function
+        // TODO: figure out why this is not executing
```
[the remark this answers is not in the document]
meskill: yes, you're currently calling the `prepare` method only inside `modify_inner` function. I explained there why it's wrong place to do this. This should be done in the `eval`, see the suggestions on how to call it there

examples/extension-i18n/main.graphql
```diff
@@ -0,0 +1,21 @@
...
+type User {
+  id: ID
+  name: String @extension(name: "translate", params: ["userName", "{{.value.name}}"])
+  username: String
+  company: Company
+}
+
+type Company {
+  name: String
+  catchPhrase: String @extension(name: "translate", params: ["catchPharse", "{{.value.catchPhrase}}"])
```
meskill, line 19:
```suggestion
  catchPhrase: String @extension(name: "translate", params: ["catchPhrase", "{{.value.catchPhrase}}"])
```

meskill (changes requested):

src/core/blueprint/operators/extension.rs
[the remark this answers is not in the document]
karatakis: If I do this, I have to define this generic all over the place, all the way to the ConfigModule. That's why I did not go with that solution.

src/core/ir/eval.rs
```diff
@@ -95,6 +95,20 @@ impl IR {
 
                     Ok(value)
                 }),
+                IR::Extension { plugin, params, ir } => {
+                    // TODO: maybe modify IR before execution
```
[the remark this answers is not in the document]
karatakis: It solved my lifetime issues and the preparation step. That's very insightful, thanks.

src/core/blueprint/operators/extension.rs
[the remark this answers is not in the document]
meskill: it shouldn't be generic everywhere and you can specify some concrete type. 

To simplify it, create an alias type for `ExtensionLoader<async_graphql_value::ConstValue>` for example and reuse it

src/core/blueprint/operators/extension.rs
```diff
@@ -0,0 +1,67 @@
...
+pub trait ExtensionLoader: std::fmt::Debug + Send + Sync {
+    fn load(&self) {}
+
+    fn modify_inner(&self, ir: Box<IR>) -> Box<IR> {
+        ir
+    }
+
+    fn prepare(&self, ir: Box<IR>, params: ConstValue) -> Box<IR>;
+
+    fn process(&self, params: ConstValue, value: ConstValue) -> Result<ConstValue, Error>;
```
meskill, line 66: it could be implemented as async, right?

meskill: I think instead of passing just some values we need to utilize some context object that could be extended and could be used to provide any additional info

src/core/blueprint/operators/extension.rs
```diff
@@ -0,0 +1,67 @@
...
+            }
+            Valid::succeed(b_field)
+        },
+    )
+}
+
+pub trait ExtensionLoader: std::fmt::Debug + Send + Sync {
+    fn load(&self) {}
+
+    fn modify_inner(&self, ir: Box<IR>) -> Box<IR> {
```
meskill, line 60: I think it shouldn't be part of this API

examples/extension-i18n/src/main.rs
```diff
@@ -0,0 +1,276 @@
...
+        &self,
+        context: ProcessContext<ConstValue>,
+    ) -> Result<ConstValue, tailcall::core::ir::Error> {
+        *(self.process_counter.lock().unwrap()) += 1;
+        if let ConstValue::String(value) = context.value {
+            let new_value = match value.as_str() {
+                "Multi-layered client-server neural-net" => {
+                    "Red neuronal cliente-servidor multicapa".to_string()
+                }
+                "Leanne Graham" => "Leona Grahm".to_string(),
```
tusharmath, line 54: Read translations from translation files.

examples/extension-i18n/src/main.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    async fn process(
+        &self,
+        context: ProcessContext<ConstValue>,
+    ) -> Result<ConstValue, tailcall::core::ir::Error> {
+        *(self.process_counter.lock().unwrap()) += 1;
+        Ok(context.value)
+    }
+}
```
tusharmath, line 131: Let's split this file into feature specific stuff ie `translate.rs` and then in `main.rs` just keep the extensions bit.

src/core/config/config.rs
```diff
@@ -604,6 +608,30 @@ pub struct Call {
...
+    DirectiveDefinition,
+)]
+#[directive_definition(locations = "FieldDefinition")]
+pub struct Extension {
+    /// Used to define where the extension is located so tailcall can load it.
+    pub name: String,
+
+    /// Used to define parameters that are calculated and injected into the IR.
+    /// Mustache syntax can be used on the parameters field.
+    pub params: Value,
```
tusharmath, line 632: This should be a `
```suggestion
    pub params: BTreeMap<String, Value>,
```

karatakis: Value is JSONLike. It can act as a superset and accept Maps.
It can accept `["test", "{{.args.test}}", 1]`, `false`, `{ "hello": "world", "foo": 12, "bar": true }`, etc. 

Do we need to force a `{ [Key]: [Value] }` structure? or should we leave the more flexible format?

If we change the format I have to change the function signatures from `JsonLike` to `BTreeMap<String, JsonLike>`

karatakis:
RFC:

The current Extension SDK implementation exposes the user to Tailcall internal API structs, which are subject to change in the future.
```rust
#[async_trait::async_trait]
pub trait ExtensionTrait<Json: JsonLikeOwned>: std::fmt::Debug + Send + Sync {
    fn load(&self) {}

    async fn prepare(&self, context: PrepareContext<Json>) -> Box<IR>;

    async fn process(&self, context: ProcessContext<Json>) -> Result<Json, Error>;
}
```

To solve this issue, I suggest we create new enumerations, `PrepareOperation` and `ProcessOperation`, which give a set of pre-defined operations for the Extension SDK user. Those enums follow a similar logic to `IR` but provide a subset of operations the user can safely use. At the same time, the development team will enforce the rule that it is always backward compatible.

Proposed solution:
```rust
#[async_trait::async_trait]
pub trait ExtensionTrait<Json: JsonLikeOwned>: std::fmt::Debug + Send + Sync {
    fn load(&self) {}

    async fn prepare_steps(&self) -> Box<PrepareOperation>;

    async fn process(&self) -> Box<ProcessOperation>;
}

pub enum PrepareOperation {
  ModifyRequestHeaders(HashMap<String, String>),
  AddRequestCookie(String),
  // TODO add other operations
  Compose(Box<PrepareOperation>, Box<PrepareOperation>),
}

pub enum ProcessOperation {
  ExtractPath(Vec<String>),
  ModifyValue(Fn(ProcessContext<Json>) -> Result<Json>),
  GroupBy(Vec<String>),
  Filter(Vec<String>, Fn(JsonLike) -> bool),
  // TODO add other operations
  Compose(Box<ProcessOperation>, Box<ProcessOperation>),
``` 

The design is not final, and it is open to suggestions

meskill:
> RFC:

Have you considered a middleware pattern for this? This be could be more ergonomic and familiar for users.

Something like this:
```rust
#[async_trait::async_trait]
pub trait ExtensionTrait<Json: JsonLikeOwned>: std::fmt::Debug + Send + Sync {
    fn load(&self) {}

    async fn execute(&self, context: Context<Json>) -> Result<Json, Error> {
        context.set_cookie(...);
        context.set_headers(...);
        let response = context.http(...).await;
        convert_response(response)
    }
}

```

karatakis:
@meskill, that's a good idea. I agree with you that more people will be comfortable with the middleware approach.

karatakis:
We are working on the SDK API

Issue #2640, opened 2024-08-11
Add snapshot tests for CLI's check command


tusharmath:
**Technical Requirements**
- Add the tests in `tests/cli/check.rs`.
- use `include_config!` macro to import configurations from a `test/fixtures` directory where test configurations should be added.
- Use the CLI programatically and create snapshot tests.
- Take inspiration from the [generator.rs](https://github.com/tailcallhq/tailcall/blob/main/src/cli/generator/generator.rs)

PR #2644, opened 2024-08-11
chore: fmt to capture Writer

ssddOnTop:
related to #2640


tusharmath:
Lets connect on this when you are back.

meskill:
Do we really need to capture the output separately?

Consider this:
- instead of calling some function from the lib we can start the binary programmatically and capture it's output.
- also we can use different tracing crates to capture the output for testing, but I don't think making changes to the code to test the cli output is a good approach

ssddOnTop:
>     * instead of calling some function from the lib we can start the binary programmatically and capture it's output.

but is that a good approach?

we couldn't get codecov, if something throws an runtime error, we won't have much information about that.. etc

meskill:
> > ```
> > * instead of calling some function from the lib we can start the binary programmatically and capture it's output.
> > ```
> 
> but is that a good approach?
> 
> we couldn't get codecov, if something throws an runtime error, we won't have much information about that.. etc

I believe it'd better represent end user experience and we will test the actual output instead of testing that some data has been written somewhere that won't mean the actual output (for testing TAILCALL_LOG). 
Also, code coverage is possible for standalone binaries though it's probably not that easy to collect, agree

ssddOnTop:
> I believe it'd better represent end user experience and we will test the actual output instead of testing that some data has been written somewhere that won't mean the actual output (for testing TAILCALL_LOG). Also, code coverage is possible for standalone binaries though it's probably not that easy to collect, agree

Okay.

Closing this PR and I'd raise new PR based on the suggestions.

meskill:
> > I believe it'd better represent end user experience and we will test the actual output instead of testing that some data has been written somewhere that won't mean the actual output (for testing TAILCALL_LOG). Also, code coverage is possible for standalone binaries though it's probably not that easy to collect, agree
> 
> Okay.
> 
> Closing this PR and I'd raise new PR based on the suggestions.

Please discuss first with @tusharmath about the approach

PR #2650, opened 2024-08-11
chore(2649): drop tag from config

### LINKED ISSUE #2649: drop `tag` from `config`
We use `tag` to store meta information about type while generating a configuration using GRPC. With the LLM powered config generation we don't need that anymore. 

**Technical Requirements**
- Just drop it :)

ssddOnTop:
**Issue Reference(s):**  
Fixes #2649
/claim 2649


src/core/generator/from_proto.rs
```diff
@@ -308,8 +308,6 @@ impl Context {
                 }
             }
 
-            ty.tag = Some(Tag { id: msg_type.id() });
```
tusharmath, line 311: We should drop `Namespace::id` method and other unnecessary logic from `Namespace` that isn't being used any more.

ssddOnTop: it is used ig

tusharmath (approved):

PR #2657, opened 2024-08-11
feat(2426): support mustache template on `generate::config`

### LINKED ISSUE #2646: Support mustache templates in all `string` fields `generate::config`
Config should support mustache templates everywhere there is a "String" type. 

**Technical Requirements**
- Templates should be able to resolve env variables.

**Before**

```yml
inputs:
- curl:
    src: https://example.com
    headers:
      user-agent: tailcall-v1
    fieldName: test
output: {}
schema: {}
```

**After**
```yml
inputs:
- curl:
    src: https://example.com
    headers:
      user-agent: "{{.env.USER}}"
    fieldName: test
output: {}
schema: {}
preset:
  inferTypeNames: true
```

ssddOnTop:
**Issue Reference(s):**  
Fixes #2646
/claim 2646


src/cli/generator/generator.rs
```diff
@@ -89,6 +84,16 @@ impl Generator {
...
+
+        if let Ok(mustache) = Mustache::parse(&config_content) {
+            config_content = mustache.render(&reader_context);
+        }
+
+        let config: Config = match source {
+            ConfigSource::Json => serde_json::from_str(&config_content)?,
+            ConfigSource::Yml => serde_yaml::from_str(&config_content)?,
+        };
+
```
tusharmath, line 96: This is an interesting approach @ssddOnTop. I think in this case we don't need a lot of logic that we have implemented right now to resolve headers and secret. Please drop that logic.

src/cli/generator/generator.rs
```diff
@@ -89,6 +84,16 @@ impl Generator {
             vars: &Default::default(),
             headers: Default::default(),
         };
+
+        if let Ok(mustache) = Mustache::parse(&config_content) {
+            config_content = mustache.render(&reader_context);
+        }
```
tusharmath, line 90: Missing tests.

src/cli/generator/config.rs
```diff
@@ -196,24 +195,21 @@ impl Location<UnResolved> {
...
 
-    pub fn as_btree_map(&self) -> &Option<BTreeMap<String, TemplateString>> {
+    pub fn as_btree_map(&self) -> &Option<BTreeMap<String, String>> {
         &self.0
     }
 }
 
 impl Headers<UnResolved> {
-    pub fn resolve(self, reader_context: &ConfigReaderContext) -> Headers<Resolved> {
+    pub fn resolve(self) -> Headers<Resolved> {
```
tusharmath, line 208: I don't think we need this API any more.

src/cli/generator/config.rs
```diff
@@ -60,7 +59,7 @@ pub struct Location<A>(
 #[derive(Deserialize, Serialize, Debug)]
 #[serde(transparent)]
 pub struct Headers<A>(
-    #[serde(skip_serializing_if = "is_default")] Option<BTreeMap<String, TemplateString>>,
+    #[serde(skip_serializing_if = "is_default")] Option<BTreeMap<String, String>>,
```
tusharmath, line 62: We can drop the type parameter from Headers

tusharmath (approved):

PR #2667, opened 2024-08-12
fix(2665): ensure `infer_type_names` sent only once

### LINKED ISSUE #2665: Send system message only once in `infer_type_names`
Infer Type Names sends the system message everytime with user message. Fix it such that the system message is only sent once for all the types who's name needs to be inferred, and the user message is sent one time for each type whose name needs to be inferred.

varshith257:
/claim #2665
fixes #2665


src/core/config/transformer/improve_type_names.rs
```diff
@@ -110,13 +111,26 @@ impl<'a> CandidateGeneration<'a> {
     }
 }
 
-#[derive(Default)]
-pub struct ImproveTypeNames;
+pub struct ImproveTypeNames {
+    system_message_sent: Cell<bool>,
```
tusharmath, line 115: not required.

tusharmath:
Changes are required in `infer_type_name.rs` not here.

PR #2681, opened 2024-08-13
fix: schema validation with GRPC

### LINKED ISSUE #2636: bug: generated gRPC configuration throws validation error ` expected type to be required `.
## Describe the bug
Generated config from gRPC throws following validation error.
error: 
```bash
ERROR Invalid Configuration
Caused by:
  • expected type to be required [at Query.news__NewsService__GetAllNews.@grpc.news__NewsList.news]
```

Above issue occurs due to following lines.
```graphQL

# we say `news__NewsList` is required but it's internals aren't required.
news__NewsService__GetAllNews: news__NewsList! @grpc(method: "news.NewsService.GetAllNews")

type news__NewsList @tag(id: "news.NewsList") {
  news: [news__News]
}
```

## Steps to reproduce
1. use following configuration to generate the tailcall config.
path for news.proto: https://github.com/tailcallhq/tailcall/blob/main/tailcall-fixtures/fixtures/protobuf/news.proto
```json
{
  "inputs": [
    {
      "proto": {
        "src": "../protobuf/news.proto"
      }
    }
  ],
  "output": {
    "path": "./output.graphql",
    "format": "graphQL"
  },
  "schema": {
    "query": "Query"
  }
}
```
2. Run the generated config.

### Technical Requirements
- add unit test for the specific test case

laststylebender14:
**For Input Types**
GraphQL Input Type < gRPC Input Type
eg:
- gRPC Optional: GraphQL optional or required
- gRPC Required: GraphQL required (Unsupported in proto 3)

**Output Types**
GraphQL Output type > gRPC Output type
eg: 
- gRPC Optional: GraphQL optional
- gRPC Required: GraphQL optional or required (Unsupported in proto 3)

**NOTE:**
`A > B` should be read as "A is a super type of B".

**Issue Reference(s):**  
Fixes #2636


src/core/generator/from_proto.rs
```diff
@@ -273,6 +273,12 @@ impl Context {
                 // required only applicable for proto2
                 cfg_field.required = matches!(label, Label::Required);
 
+                if cfg_field.list {
+                    // by default proto3 fields are optional and default values are provided.
+                    // in order to support proto2, we mark these list fields required.
+                    cfg_field.required = true;
```
meskill, line 279: it's acceptable to mark the field as required for output types, but for input types it adds a stricter requirement than defined by protobuf which may be not desired.

we have todo like this https://github.com/tailcallhq/tailcall/blob/64e3c53747c7a6a47eebff74a5b85a6ea17f19e4/src/core/blueprint/operators/grpc.rs#L81-L84
and I generally think that json_schema validation should not enforce the nullability of the fields at all:
- if you define the field as optional in graphql everything is fine because fields in proto3 are optional
- if you define the field as required in graphql you just enforcing rules on the graphql side and everything still works on grpc side
- for proto2 field should be marked as required if it's required and therefore we need to validate that graphql also defines it as required

tests/execution/test-grpc-nested-optional.md
```diff
@@ -1,5 +1,5 @@
 ---
-error: true
+error: false
```
meskill, line 2: isn't it intentional to get the validation error here?

so right now validation fails if we mark fields for proto as required and we need to have a test for this.

meskill (changes requested):

meskill (approved):

tusharmath (approved):

tusharmath (approved):

Issue #2695, opened 2024-08-14
Rewrite the implementation for `config::Resolver` from declarative to procedural derive macro


meskill:
Current implementation in https://github.com/tailcallhq/tailcall/pull/2675 is done with declarative macro that has limitations.

Rewrite it with derive macro. Add new macro to the `tailcall-macros` workspace member

meskill:
fixed by https://github.com/tailcallhq/tailcall/pull/2721

PR #2721, opened 2024-08-16
refactor(2695): derive macro for config::Resolver

mehulmathur16:
/claim #2695


tusharmath:
@meskill LGTM. Can you also check once.

src/core/config/config.rs
```diff
@@ -210,53 +210,11 @@ pub struct RootSchema {
...
-                    }
-                })
-            }
-        }
-    };
-}
-
-create_resolver! {
+#[derive(
+    Serialize, Deserialize, Clone, Debug, PartialEq, Eq, schemars::JsonSchema, CustomResolver,
```
tusharmath, line 214: We are essentially deriving `DirectiveCodec`, can we use that?

tailcall-macros/src/resolver.rs
```diff
@@ -0,0 +1,92 @@
...
+                }
+                if result.is_some() {
+                    return Valid::fail(format!(
+                        "Multiple resolvers detected [{}]",
+                        resolvable_directives.join(", ")
+                    ));
+                }
+                result = Some(Self::#variant_name(resolver));
+            } else if let Err(e) = result_option {
+                return Valid::from_validation_err(e);
```
meskill, line 44: can we preserve original logic about validation errors?

the way it was implemented it combines validation errors instead on failing fast like in your version

tailcall-macros/src/resolver.rs
```diff
@@ -0,0 +1,92 @@
...
+                let directive_name = <#ty>::trace_name();
+                if !resolvable_directives.contains(&directive_name) {
+                    resolvable_directives.push(directive_name);
+                }
+                if result.is_some() {
+                    return Valid::fail(format!(
+                        "Multiple resolvers detected [{}]",
+                        resolvable_directives.join(", ")
+                    ));
+                }
```
meskill, line 41: check `resolvable_directives` length after executing the conversions as it was done before.

That way we show an error about all defined directives instead of the only first two

tailcall-macros/src/resolver.rs
```diff
@@ -0,0 +1,92 @@
...
+            })
+            .collect::<Vec<_>>()
+    } else {
+        panic!("Resolver can only be derived for enums");
+    };
+
+    let variant_parsers = variants.iter().map(|(variant_name, ty)| {
+        quote! {
+            let temp_result = <#ty>::from_directives(directives.iter());
+            let result_option = temp_result.to_result();
```
meskill, line 30: why convert to result and then back to Valid instead of using Valid in the first place?

meskill (changes requested):

src/core/config/config.rs
meskill: we can't do it right now because we need a custom implementation of `DirectiveCodec` here and the default one `impl<'a, A: Deserialize<'a> + Serialize + 'a> DirectiveCodec for A` doesn't allow to write specialized implementation

meskill (approved):

PR #2754, opened 2024-08-25
fix(jit): raise validation error for required fields

karatakis:
**Summary:**  
Fix validation of required fields to cover edge cases for the JIT module

Remember to test it with `cargo test --features force_jit`

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/core/blueprint/blueprint.rs
+++ b/src/core/blueprint/blueprint.rs
@@ -77,6 +77,15 @@ impl Type {
             Type::ListType { non_null, .. } => *non_null,
         }
     }
+
+    // checks if inner type is nullable
+    pub fn is_inner_nullable(&self) -> bool {
+        match self {
+            Type::NamedType { non_null, .. } => !*non_null,
+            Type::ListType { of_type, .. } => of_type.is_nullable(),
+        }
+    }
+
     /// checks if the type is a list
     pub fn is_list(&self) -> bool {
         matches!(self, Type::ListType { .. })
--- a/src/core/blueprint/blueprint.rs
+++ b/src/core/blueprint/blueprint.rs
@@ -78,14 +78,6 @@ impl Type {
         }
     }
 
-    // checks if inner type is nullable
-    pub fn is_inner_nullable(&self) -> bool {
-        match self {
-            Type::NamedType { non_null, .. } => !*non_null,
-            Type::ListType { of_type, .. } => of_type.is_nullable(),
-        }
-    }
-
     /// checks if the type is a list
     pub fn is_list(&self) -> bool {
         matches!(self, Type::ListType { .. })
--- a/src/core/jit/synth/synth.rs
+++ b/src/core/jit/synth/synth.rs
@@ -100,10 +100,10 @@ where
         &'a self,
         node: &'a Field<Nested<Value>, Value>,
     ) -> Result<Value, Positioned<Error>> {
+        // according to GraphQL spec https://spec.graphql.org/October2021/#sec-Handling-Field-Errors
         if node.type_of.is_nullable() {
             Ok(Value::null())
         } else {
-            // link to GraphQL reference https://spec.graphql.org/October2021/#sec-Handling-Field-Errors
             Err(ValidationError::ValueRequired.into()).map_err(|e| self.to_location_error(e, node))
         }
     }
@@ -121,7 +121,7 @@ where
         }
 
         let eval_result = if value.is_null() {
-            // link to GraphQL reference https://spec.graphql.org/October2021/#sec-Handling-Field-Errors
+            // check the nullability of this type unwrapping list modifier
             let is_nullable = match &node.type_of {
                 crate::core::blueprint::Type::NamedType { non_null, .. } => !*non_null,
                 crate::core::blueprint::Type::ListType { of_type, .. } => of_type.is_nullable(),
--- a/src/core/jit/synth/synth.rs
+++ b/src/core/jit/synth/synth.rs
@@ -94,7 +94,7 @@ where
                     Data::Single(result) => {
                         let value = result.as_ref().map_err(Clone::clone)?;
 
-                        if !node.type_of.is_list() == value.as_array().is_some() {
+                        if node.type_of.is_list() != value.as_array().is_some() {
                             return self.node_nullable_guard(node);
                         }
                         self.iter_inner(node, value, data_path)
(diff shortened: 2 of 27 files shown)


karatakis:
Test case `test-required-fields.md_5.snap` needs discussion. I believe it should not fail, but the current implementation fails. This issue is open for discussion.

meskill:
> Test case `test-required-fields.md_5.snap` needs discussion. I believe it should not fail, but the current implementation fails. This issue is open for discussion.

Why do you think it should not fail? You have non-null list of non-null values and one of the values is null - according to spec the parent value should be resolved to null in that case and that is against the schema

karatakis:
My bad, @meskill; I messed up the numbering, I believe. Maybe I fixed the issue while I was working on it yesterday.

tests/execution/test-required-fields.md
```diff
@@ -0,0 +1,193 @@
...
+      id: 1
+      bar: bar_1
+
+# this does not fail
+- request:
+    method: GET
+    url: http://jsonplaceholder.typicode.com/relaxed-missing
+  response:
+    status: 200
+    body: null
```
meskill, line 66: let's add test for `Foo` but with body `id: 1 bar: null`

tests/execution/test-required-fields.md
```diff
@@ -0,0 +1,193 @@
...
+      - id: 2
+        bar: bar_2
+
+# this fails
+- request:
+    method: GET
+    url: http://jsonplaceholder.typicode.com/full-missing
+  response:
+    status: 200
+    body:
```
meskill, line 86: additional test for `body: null`

tests/execution/test-required-fields.md
```diff
@@ -0,0 +1,193 @@
...
+- request:
+    method: GET
+    url: http://jsonplaceholder.typicode.com/inner-missing
+  response:
+    status: 200
+    body:
+      - id: 1
+        bar: bar_1
+      - id: 2
+        bar: null
```
meskill, line 109: for all of the list cases we need probably 4 tests:
- the whole list is null
- the list is not null and some entry is null
- the list is not null and some entry has inner null
- the list with values

wdyt?

tests/core/snapshots/test-required-fields.md_7.snap
```diff
@@ -0,0 +1,28 @@
...
+source: tests/core/spec.rs
+expression: response
+---
+{
+  "status": 200,
+  "headers": {
+    "content-type": "application/json"
+  },
+  "body": {
+    "data": null,
```
meskill, line 11: btw if I understand the spec right it should be `data: { innerMissing: null }` here. But I'm not sure is the problem in async_graphql or tailcall. 

Should we leave some todos related to this?

src/core/jit/synth/synth.rs
```diff
@@ -128,10 +142,22 @@ where
...
+                    } else {
+                        Err(ValidationError::ValueRequired.into())
+                    }
+                }
+                crate::core::blueprint::Type::ListType { of_type, non_null: _ } => {
+                    if of_type.is_nullable() {
+                        Ok(Value::null())
+                    } else {
+                        Err(ValidationError::ValueRequired.into())
+                    }
```
meskill, line 159: this check is everywhere, can we move it somewhere and reuse it?

src/core/jit/synth/synth.rs
```diff
@@ -128,10 +142,22 @@ where
         }
 
         let eval_result = if value.is_null() {
-            if node.type_of.is_nullable() {
-                Ok(Value::null())
-            } else {
-                Err(ValidationError::ValueRequired.into())
+            // link to GraphQL reference https://spec.graphql.org/October2021/#sec-Handling-Field-Errors
+            match &node.type_of {
+                crate::core::blueprint::Type::NamedType { name: _, non_null: _ } => {
```
meskill, line 147: do we need a match here or we can use `is_nullable` on the type_of itself?

meskill (changes requested):

src/core/jit/synth/synth.rs
```diff
@@ -68,9 +68,9 @@ where
         Ok(Value::object(data))
     }
 
-    /// checks if type_of is an array and value is an array
+    /// checks if both are array or both are scalar type
     #[inline(always)]
-    fn is_array(type_of: &crate::core::blueprint::Type, value: &Value) -> bool {
+    fn is_both_array_or_scalar(type_of: &crate::core::blueprint::Type, value: &Value) -> bool {
```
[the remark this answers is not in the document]
karatakis: I renamed it because the original name was not right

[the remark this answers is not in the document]
meskill: makes sense

src/core/jit/synth/synth.rs
karatakis: We want to test the inside of the type in case it is a list.
If we have `[Foo!]` in this scenario, I check if Foo is nullable or not, so I need to unwrap

tests/execution/test-required-fields.md
karatakis: I can do it, it won't take much time. So we can be 100% sure

tests/core/snapshots/test-required-fields.md_7.snap
karatakis: I can check it with the graphql reference implementation and I will reply

karatakis: yeah data is `{ innerMissing: null }`

karatakis: Added a TODO, it can be solved to separate issue

src/core/blueprint/blueprint.rs
```diff
@@ -77,6 +77,15 @@ impl Type {
...
         }
     }
+
+    // checks if inner type is nullable
+    pub fn is_inner_nullable(&self) -> bool {
+        match self {
+            Type::NamedType { non_null, .. } => !*non_null,
+            Type::ListType { of_type, .. } => of_type.is_nullable(),
+        }
+    }
```
tusharmath, line 87: Technically it should return an `Option<bool>` inner isn't available if self is a NamedType. I would rather have this function inlined into its only usage as this seems more like business logic than a helper method on Type.
Action: Inline the function wherever it's used.

meskill (approved):

PR #2757, opened 2024-08-26
fix(grpc): handle union shared fields

meskill:
**Summary:**  
Pr to address differences in async_graphql and jit engines related to selecting shared fields from union type.

I've checked other graphql implementations and it appeared that original test should not work actually based on spec. So this pr:
- generates shared interface for oneOf for proto config generator
- test now uses interface fragment to select shared fields
- improve jit implementation for tests to handle interfaces on fragments

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


tests/execution/grpc-oneof.md
```diff
@@ -169,9 +173,9 @@ type oneof__Response__Var2 @tag(id: "oneof.Response") {
     query: >
       query {
         oneof__OneOfService__GetOneOfVar1(request: { command: { command: "start" } }) {
-          # TODO: check that it's possible to get shared field from Union like that
-          # outside of the fragment
-          usual
```
meskill, line 174: this is wrong way to select the shared field. It doesn't work for graphql-js and static async-graphql and seems like possible due to bug in the dynamic implementation

src/core/blueprint/index.rs
```diff
@@ -66,6 +66,37 @@ impl Index {
...
+            self.map.get(implements),
+            Some((Definition::Interface(_), _))
+        ) {
+            return false;
+        }
+
+        self.is_type_implements_rec(type_name, implements)
+    }
+
+    fn is_type_implements_rec(&self, type_name: &str, implements: &str) -> bool {
```
tusharmath, line 85: I see a potential stackoverflow scenario. Can use a hashset to ensure to handle it. Also please add tests for the stack overflow condition also.

src/core/blueprint/index.rs
```diff
@@ -66,6 +66,37 @@ impl Index {
     pub fn get_mutation(&self) -> Option<&str> {
         self.schema.mutation.as_deref()
     }
+
+    pub fn is_type_implements(&self, type_name: &str, implements: &str) -> bool {
```
tusharmath, line 70:
```suggestion
    pub fn is_type_implements(&self, implements: &str, type_name: &str) -> bool {
```

Nit Pick: In general its a good idea to keep the most changing arguments on the right and the least changing arguments on the left. For eg: `self` doesn't change and stays on the left, similarly we are looking for `implements` in a chain of types, so keeping that as the second parameter seems intuitive.

src/core/blueprint/index.rs
```diff
@@ -66,6 +66,37 @@ impl Index {
...
+        }
+
+        if !matches!(
+            self.map.get(implements),
+            Some((Definition::Interface(_), _))
+        ) {
+            return false;
+        }
+
+        self.is_type_implements_rec(type_name, implements)
```
tusharmath, line 82: I also feel since this is an index, we should pre-populate the implementations of each type so that finding that a type implements an interface is an O(1) operation and not an `O(N)` which is what it is right now.

src/core/blueprint/index.rs
meskill: I've checked the graphql spec again https://spec.graphql.org/October2021/#sec-Interfaces.Interfaces-Implementing-Interfaces and according to it the type should implement all transitive interfaces. So I removed the recursion check and left the only interface present in the list of `obj.implements`

src/core/blueprint/index.rs
meskill: recursion is removed the check is O(1) now

src/core/blueprint/index.rs
meskill: but it also should correspond to the name of the method otherwise it would be confusing.
Also after removal of recursive call both parameters are changing in the same magnitude so I left it unchanged

tusharmath (approved):

Issue #2775, opened 2024-09-01
Push data to PostHog


tusharmath:
We capture some metrics and send it to GA. Add posthog to one of the other collectors of metrics.

**Technical Requirements**
- Ensure there is no code duplication.
- Keep the structure clean by creating a `collect/ga.rs` and `collect/posthog.rs` files that handle collection of data.
- Implement data collection via a trait.

**Additional Links**
- https://posthog.com/
- https://posthog.com/docs/libraries/rust

PR #2776, opened 2024-09-01
fix(2775): send metrics to posthog

beelchester:
fixes: #2775
/claim #2775

Screenshot 2024-09-01 at 5 44 44 PM

---
Steps to test:

1. Add posthog project api key to env `POSTHOG_API_SECRET`.
2. For local testing I had to hardcode `is_tracking` to true.
3. Run cargo test on tailcall-tracker or just use build the project and run any argument.
4. Check the activity section in posthog project events should appear.


tailcall-tracker/src/ga_event.rs
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+}
+
+/// Event structure to be sent to GA
+#[derive(Debug, Serialize, Deserialize)]
+pub struct GaEvent {
+    client_id: String,
+    events: Vec<GaEventValue>,
+}
```
tusharmath, line 45: Make it private to collect/ga.rs

tailcall-tracker/src/tracker.rs
```diff
@@ -1,77 +1,51 @@
...
     fn default() -> Self {
+        let ga_tracker = GaTracker::default();
+        let posthog_tracker = PostHogTracker::default();
         Self {
-            base_url: BASE_URL.to_string(),
-            api_secret: API_SECRET.to_string(),
-            measurement_id: MEASUREMENT_ID.to_string(),
+            collectors: vec![Box::new(ga_tracker), Box::new(posthog_tracker)],
             is_tracking: check_tracking(),
         }
```
tusharmath, line 26: We can set the start time in the Tracker over here instead of passing it from outside via init and dispatch.

tailcall-tracker/src/helpers.rs
```diff
@@ -0,0 +1,31 @@
...
+    System::long_os_version().unwrap_or("Unknown".to_string())
+}
+
+pub fn get_uptime(start_time: DateTime<Utc>) -> String {
+    let current_time = Utc::now();
+    format!(
+        "{} minutes",
+        current_time.signed_duration_since(start_time).num_minutes()
+    )
+}
```
tusharmath, line 31: The functions need not be public, instead we should expose an enum that represents an event.

```rust
struct Event {
    time: Date,
    cores: Option<u8>,
    client_id: String,
    os_name: String,
    up_time: Date,
    path: Option<String>,
    user: Option<String>,
    kind: Kind
}

enum Kind {
    Ping,
    Run { command: String, args: Vec<String> }
}
```

The event should then be passed on to the collectors and it's then the collectors responsibility to ensure that the data send is correctly passed up to GA or PostHog.

tailcall-tracker/src/collect/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod ga;
+pub mod posthog;
```
tusharmath, line 2: Don't expose posthog or GA publicly. I would recommend we create a wrapper type that composes to collectors and dispatches the event to both the collectors.

tusharmath:
- Added comments 
- CI is failing which needs fixing
- Add steps to test the feature

Moving to draft to reduce noise and improve CI efficiency. Once you are ready just mark it as "ready to review". Feel free to give a shoutout on the #contributors channel on Discord if you want immediate attention.

tusharmath:
closing in favour of https://github.com/tailcallhq/tailcall/pull/2879

PR #2793, opened 2024-09-05
fix: added type rename support for enum, union and interface.

laststylebender14:
**Summary:**  
previously rename types transformer supported only renaming in normal types, now extended the support to enum, union and interface types.

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/config/transformer/rename_types.rs
```diff
@@ -65,6 +83,51 @@ impl Transform for RenameTypes {
                         }
                     }
                 }
+
+                // replace in interface.
+                type_.implements = type_
+                    .implements
+                    .iter()
+                    .filter_map(|interface_type_name| {
```
karatakis, line 91: Why are you using `filter_map` when you are always returning `Some`? 

Can a normal `map` with `match` work instead?

src/core/config/transformer/rename_types.rs
```diff
@@ -65,6 +83,51 @@ impl Transform for RenameTypes {
...
+                        lookup
+                            .get(interface_type_name)
+                            .cloned()
+                            .or(Some(interface_type_name.clone()))
```
karatakis, line 95: This is eagerly evaluated, use `or_else`

src/core/config/transformer/rename_types.rs
```diff
@@ -65,6 +83,51 @@ impl Transform for RenameTypes {
...
+                    union_type_.types.insert(type_name);
+                }
+            }
+
+            // replace in union as well.
+            for union_type_ in config.unions.values_mut() {
+                union_type_.types = union_type_
+                    .types
+                    .iter()
+                    .filter_map(|type_name| {
```
karatakis, line 127: 1) why we also use `filter_map`?
2) replace `or` with `or_else` because it lazily evaluates

karatakis (commented):
I found some slips, but overall its a great work.

src/core/config/transformer/rename_types.rs
laststylebender14: yep, replace with map.

src/core/config/transformer/rename_types.rs
laststylebender14: using `unwrap_or`.

karatakis (approved):

PR #2816, opened 2024-09-15
chore: make merge_right return valid

### LINKED ISSUE #2815: refactor: Return a `Valid` from `merge_right`
Currently `merge_right` can never fail, we need to fix that by returning a Valid

**Technical Requirements**
- Change the type signature of merge_right to return a `Valid<Self, String>`

ssddOnTop:
**Issue Reference(s):**  
Fixes #2815


ssddOnTop:
closing in favour of: #2873

PR #2860, opened 2024-09-18
feat: add on response body hook to http directives

### LINKED ISSUE #2857: feat: add `onResponseBody` hook for JS API
The current `http` and `grpc` directives need to be expanded with `onResponseBody` hook. If specified this should be called by the runtime with the complete response body. 

**Example**

```graphql

type Query {
   users: [User] @http(path: "/users", onResponseBody: "onUserResponse")
}

```

**Technical Requirements**
- Ensure the implementation is efficient and produces minimal performance penalty, by sending data as bytes or string to the JS function where they can parse it if they want to and process it if required.
- The return value of the function needs to be used as the final output of the HTTP call.
- Ensure sufficient unit tests and integration tests have been added.

**Additional Notes**
- Take inspiration from the implementation of `onRequest`

laststylebender14:
**Summary:**  
docs: https://github.com/tailcallhq/tailcallhq.github.io/pull/494

**Issue Reference(s):**  
Fixes #2857

**Build & Testing:**

**Checklist:**


src/core/ir/eval_http.rs
```diff
@@ -109,6 +113,24 @@ impl<'a, 'ctx, Context: ResolverContextLike + Sync> EvalHttp<'a, 'ctx, Context>
...
             },
             None => self.execute(request).await,
+        };
+
+        // send the final response to JS script to futher evaluation.
+        if let Ok(resp) = resp {
+            if let Some(on_response) = http_filter.on_response.as_ref() {
+                let js_response = worker::WorkerResponse::try_from(resp.clone())?;
+                let response_event = worker::Event::Response(js_response);
+                let final_command = worker.call(on_response, response_event).await?;
```
tusharmath, line 123: The worker response should not send the response body to the JS function. We have `on_response_body` for that. Event worker response is going to be a little different than the command worker response.

tusharmath: Ignore everything above. I think we don't need any of this code. `on_response_body` should only send the response body to the JS function.

tusharmath (approved):

PR #2863, opened 2024-09-19
fix(gen): read from parent_dir, then try cwd

karatakis:
**Summary:**  
In a discussion on Discord, the user `ben` had an issue regarding the generation of tailcall configuration from a proto file. The issue was that they used relative paths from the project root folder, whereas we used relative to the folder of the proto file during the stitching of proto files. To fix this issue, we now do the following: (1) we check for the proto file we want to stitch if it exists in the folder relative to the parent proto file, and if we do not find we (2) try to load the target file for stitching from the path where the CLI command had been run.

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/core/proto_reader/reader.rs
+++ b/src/core/proto_reader/reader.rs
@@ -156,8 +156,21 @@ impl ProtoReader {
                 .context("Unable to extract content of google well-known proto file")?
                 .to_string()
         } else {
-            let path = Self::resolve_path(path.as_ref(), parent_dir);
-            self.reader.read_file(path).await?.content
+            // try to read the file relative to the parent_dir
+            let parent_file = {
+                let path = Self::resolve_path(path.as_ref(), parent_dir);
+                self.reader.read_file(path).await
+            };
+
+            // if we fail then we read relative to the current working directory of the cli
+            // tool
+            if let Ok(file) = parent_file {
+                file.content
+            } else {
+                let root_dir = std::env::current_dir().unwrap();
+                let path = Self::resolve_path(path.as_ref(), Some(&root_dir));
+                self.reader.read_file(path).await?.content
+            }
         };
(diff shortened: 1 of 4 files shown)


src/core/proto_reader/reader.rs
```diff
@@ -156,8 +156,21 @@ impl ProtoReader {
...
+
+            // if we fail then we read relative to the current working directory of the cli
+            // tool
+            if let Ok(file) = parent_file {
+                file.content
+            } else {
+                let root_dir = std::env::current_dir().unwrap();
+                let path = Self::resolve_path(path.as_ref(), Some(&root_dir));
+                self.reader.read_file(path).await?.content
+            }
```
laststylebender14, line 173: store the config file under: `test/config.json`
i was present in path: `tailcall/src/core`
and then i run following command `cargo run -- gen ../../tests/test.json`

got following error
Failed to read file: `/Users/ranjitmahadik/Desktop/tailcall/tests/tailcall-fixtures/fixtures/protobuf/news_root.proto`: No such file or directory (os error 2)


whereas actual path should be `/Users/ranjitmahadik/Desktop/tailcall/tailcall-fixtures/fixtures/protobuf/news_root.proto`

laststylebender14: found above issue

meskill: good catch.
It's because we always get the absolute path here that was previously resolved. See code at https://github.com/tailcallhq/tailcall/blob/075e1f86749fa9614bafd4b33f99e5ee4d3ea450/src/cli/generator/config.rs#L254

And it's always this way for the first protobuf file defined in the generator config, while any nested protobuf files could be resolved relative to cwd. It seems in tests there is different resolver logic is applied in therefore there is no error comparing to running it locally

PR #2886, opened 2024-09-23
fix: directives name clashes with apollo federation for link

meskill:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/sdl.rs
```diff
@@ -363,48 +339,97 @@ fn print_directive_type_def(directive: &DirectiveDefinition) -> String {
...
-                TypeKind::Union(_) => unions.push(print_type_def(&type_def.node)),
-                TypeKind::InputObject(_) => inputs.push(print_type_def(&type_def.node)),
-            },
-            TypeSystemDefinition::Directive(type_def) => {
-                directives.push(print_directive_type_def(&type_def.node))
-            }
+#[derive(Default)]
+pub struct SdlPrinter {
+    pub federation_compatibility: bool,
+}
```
tusharmath, line 345: This is an interesting approach where we by default add federation specific items if the flag is set but then we are scattering this sub-graph specific logic everywhere in our codebase. My recommendation is to add a top-level `definitions` field in `Config` where we can add any additional definitions, like in this case - `@link` these additional definitions aren't consumed by Tailcall but are outputed as it is.

meskill: not exactly, the approach is not to output the Tailcall directives that conflicts with federation directives. To make the sdl output federation compatible we need to include apollo spec with federation's `@link` and by doing so we can't now leave tailcall's usage and definitions of `@link`

tusharmath (approved):

PR #2747, opened 2024-08-23
feat(config): add support for nested lists

meskill:
**Summary:** 
Refactor the way the type represented for field and arg in order to support any level of list nesting in the sdl.

It's without a fix for the jit implementation since it's already a big pr and jit handling for lists could be implemented as https://github.com/tailcallhq/tailcall/pull/2635

**Build & Testing:**

- [ ] I ran `cargo test` successfully.
- [ ] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [ ] I have added relevant unit & integration tests.
- [ ] I have updated the [documentation] accordingly.
- [ ] I have performed a self-review of my code.
- [ ] PR follows the naming convention of `<type>(<optional scope>): <title>`

[documentation]: https://github.com/tailcallhq/tailcallhq.github.io/tree/develop/docs

PR #2894, opened 2024-09-24
feat: add support for external directives

meskill:
**Summary:**  

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**


src/core/config/config.rs
```diff
@@ -123,6 +123,11 @@ pub struct Type {
     /// skip since it's set automatically by config transformer
     #[serde(skip_serializing)]
     pub key: Option<Key>,
+
+    ///
+    /// Any additional directives
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub directives: Vec<Directive>,
```
tusharmath, line 130: Define directives for Config separately. Config shouldn't depend on blueprint primitives ideally. Blueprint can depend on config though.

src/core/blueprint/blueprint.rs
```diff
@@ -123,11 +124,10 @@ impl FieldDefinition {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, schemars::JsonSchema)]
 pub struct Directive {
```
tusharmath, line 128: drop `schemars` from here.

tusharmath:
@meskill added some comments. Please resolve conflicts.

src/core/config/config.rs
meskill: I'm noticing some contradictory guidance based on your previous comment https://github.com/tailcallhq/tailcall/pull/2747#discussion_r1732148289. Could you please help clarify how to handle this so we can have a consistent approach moving forward? This clarification could also be valuable as part of the contribution guide for others. I believe we can avoid code duplication by reusing the implementation in one entity from another. What are your thoughts on this?

meskill: moved this to core to reuse in both config and blueprint

tests/execution/apollo-federation-entities.md
```diff
@@ -21,7 +21,7 @@ type User @call(steps: [{query: "user", args: {id: "{{.value.id}}"}}]) {
 ```graphql @file:posts.graphql
 type Post @expr(body: {id: "{{.value.id}}", title: "post-title-{{.value.id}}"}) {
   id: Int!
-  title: String!
+  title: String! @override(from: "name")
```
tusharmath, line 24: Lets add a separate test to support extra directives. This feature is independent of federation.

src/core/directive.rs
```diff
@@ -1,18 +1,24 @@
...
     Positioned::new(a, Pos::default())
 }
 
-#[allow(dead_code)]
-fn to_const_directive(directive: &blueprint::Directive) -> Valid<ConstDirective, String> {
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, schemars::JsonSchema)]
+pub struct Directive {
+    pub name: String,
+    pub arguments: HashMap<String, Value>,
+}
```
tusharmath, line 19: What if we wish to attach additional information such as IR to a directive in future? Something that won't be meaningful in Config but useful in blueprint.

tusharmath: In my humble opinion we should keep them separated - the `blueprint::Directive` and `config::Directive`. One change I can see immediately is in `Config` I would like to use a BTreeMap or an IndexMap because I would like to maintain an order, but in Blueprint I could use a HashMap for better performance.

tests/execution/apollo-federation-entities.md
meskill: done

src/core/directive.rs
meskill: applied the suggested change

src/core/config/from_document.rs
```diff
@@ -422,6 +428,22 @@ fn to_add_fields_from_directives(
     )
 }
 
+fn to_unknown_directives(
+    directives: &[Positioned<ConstDirective>],
+) -> Valid<Vec<Directive>, String> {
+    Valid::from_iter(directives.iter(), |directive| {
+        if KNOWN_DIRECTIVES
```
tusharmath, line 435: This potentially is going to introduce a bug when the developer forgets to update known_directives. We should change the logic to detect directives here. Keep a mutable list of directives and keep removing the consumed directives and add the remaining as unknown (or something around those lines)

meskill: I agree this approach is error-prone. While doing this I explored the other possibilities but found it'll will either require some significant refactoring around directives or may introduce other bugs since it'll require more code to work. 

So I went by simple path since I saw already similar approach https://github.com/tailcallhq/tailcall/blob/aab7e07349cfb6421c231260e40bd4a97e7ea80b/src/core/config/config.rs#L681-L733

meskill: What would be your suggestion here? I think directives have also other flaws that could be reworked together (e.g. working only with non repeatable directives, no way to provide custom implementation of trait DirectiveCodec etc.)

tusharmath: Alternative approach: How about we create a white list of additional directives that we can copy. For eg: If `enableFederation` is set to true, we allow the federation specific directives into the configuration.

tusharmath (approved):

PR #2899, opened 2024-09-24
chore: handle possible list types for batched grpc data_loader

### LINKED ISSUE #2891: Handle possible list types for batched data_loader implementation of `@grpc`
In some cases batched request could be resolved to multiple values for every entry of the batched parent value. The logic is already implemented for the `@http` dataloader depending on the type of the field. See https://github.com/tailcallhq/tailcall/blob/23a56b4a980ad0df84cfb7c4b7d19467dccc2e9d/src/core/http/data_loader.rs#L45-L49 for reference.

Implement support for list types for the `@grpc` batched request

ssddOnTop:
**Issue Reference(s):**  
Fixes #2891

### DIFF
--- a/src/core/grpc/data_loader.rs
+++ b/src/core/grpc/data_loader.rs
@@ -18,14 +18,51 @@ use crate::core::http::Response;
 use crate::core::json::JsonLike;
 use crate::core::runtime::TargetRuntime;
 
+fn get_body_value_single(body_value: &HashMap<String, Vec<&ConstValue>>, id: &str) -> ConstValue {
+    body_value
+        .get(id)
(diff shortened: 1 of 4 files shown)


src/core/grpc/data_loader.rs
```diff
@@ -18,14 +18,51 @@ use crate::core::http::Response;
...
+            runtime,
+            operation,
+            group_by,
+            body: if is_list {
+                get_body_value_list
+            } else {
+                get_body_value_single
+            },
+        }
+    }
```
tusharmath, line 64: Add some integration tests for gRPC batching.

PR #2935, opened 2024-09-29
fix: filter out tailcall specific directives

### LINKED ISSUE #2929: hide tailcall directives from `_service`
Currently all the tailcall directives are outputted in the `_service` call. We should only expose the final client SDL with the federation specific directives.

**Why?**
Because of the current implementation everytime `Config` changes, the federation spec tests also need to be updated.

cc: @meskill

ssddOnTop:
**Issue Reference(s):**  
Fixes #2929


src/core/blueprint/operators/apollo_federation.rs
```diff
@@ -96,7 +114,7 @@ pub fn compile_service(config: &ConfigModule) -> Valid<IR, String> {
 
 fn filter_conflicting_directives(sd: ServiceDocument) -> ServiceDocument {
     fn filter_directive(directive_name: &str) -> bool {
-        directive_name != "link"
+        FEDERATION_DIRECTIVES.iter().any(|d| *d == directive_name)
```
meskill, line 117:
```suggestion
        FEDERATION_DIRECTIVES.contains(directive_name)
```

src/core/blueprint/operators/apollo_federation.rs
```diff
@@ -13,6 +15,22 @@ use crate::core::try_fold::TryFold;
 use crate::core::valid::{Valid, Validator};
 use crate::core::{config, Type};
 
+// ref: https://www.apollographql.com/docs/federation/federated-schemas/federated-directives
+const FEDERATION_DIRECTIVES: &[&str] = &[
```
meskill, line 19: we have a similar list of federation directives here https://github.com/tailcallhq/tailcall/pull/2935/files#diff-f40f5478a0ee614381f25d37362c9c9c7779df174a375a166cf70e3e392e5a5dR109 can we reuse the definition?

meskill (changes requested):

tusharmath:
The implementation for generating `_service` needs to change. Closing this PR because the implementation in its base branch itself isn't right.

Issue #2893, opened 2024-09-24
add `dedupe` capability on all IO operations


tusharmath:
Currently dedupe is only support on `upstream`. We would like to provide more low-level support for individual IO calls.

**Example**

```graphql
type Query {
  users: [User] @http(path: "/users", dedupe: true)
}
```

**Technical Requirements**
- Support it for Http, Grpc, Graphql, Call
- Add a field in operation_plan to mark a query as deduped if everything in the query is deduped.

PR #2945, opened 2024-09-30
fix: add dedupe capability on all IO operations

ssddOnTop:
Issue Reference(s):
Fixes https://github.com/tailcallhq/tailcall/issues/2893


src/core/jit/model.rs
```diff
@@ -408,6 +410,14 @@ impl<Input> OperationPlan<Input> {
...
             .map(|f| f.into_nested(&fields))
             .collect::<Vec<_>>();
+        let dedupe = fields.iter().filter(|v| v.ir.is_some()).all(|v| {
+            v.ir.as_ref()
+                .map(|v| match v {
+                    IR::IO(io) => io.dedupe(),
+                    _ => false,
+                })
+                .unwrap_or_default()
+        });
```
tusharmath, line 420: Move this logic to a function and add a unit test.

src/core/jit/model.rs
```diff
@@ -410,7 +410,7 @@ impl<Input> OperationPlan<Input> {
             .filter(|f| f.extensions.is_none())
             .map(|f| f.into_nested(&fields))
             .collect::<Vec<_>>();
-        let dedupe = fields.iter().filter(|v| v.ir.is_none()).all(|v| {
+        let dedupe = fields.iter().filter(|v| v.ir.is_some()).all(|v| {
```
tusharmath, line 413: We don't need `filter` Can pattern match directly inside `all` and return a true or false.

tailcall-fixtures/fixtures/configs/jsonplaceholder.graphql
```diff
@@ -5,7 +5,7 @@ schema
 }
 
 type Query {
-  posts: [Post] @http(path: "/posts?_limit=11")
+  posts: [Post] @http(path: "/posts?_limit=11", dedupe: true)
```
tusharmath, line 8: create a new test file `dedupe.graphql` with different `dedupe` settings.

src/core/jit/model.rs
```diff
@@ -703,4 +714,29 @@ mod test {
...
+        let config = include_config!("../../../tailcall-fixtures/fixtures/configs/jsonplaceholder.graphql").unwrap();
+        let module = ConfigModule::from(config);
+        let bp = Blueprint::try_from(&module).unwrap();
+
+        let request = Request::new(r#"{ users { id } }"#);
+        let jit_request = jit::Request::from(request);
+        let plan = jit_request.create_plan(&bp).unwrap();
+
+        assert!(!plan.dedupe);
+    }
```
tusharmath, line 743: add another test where there are nested calls and some are deduped and some are not.

ssddOnTop: it's checked on flattened form of field

tusharmath (approved):

PR #2953, opened 2024-10-02
chore: move OP resolution logic to `ConstValueExecutor`

### LINKED ISSUE #2952: chore: avoid any resolution to `OperationPlan` with `Variables`
Currently, we use `Variables` to skip adding the fields that are always skipped, and to resolve input data. But ideally, that should not happen while creating `OperationPlan`. 

**Proposal**
Move the logic to `ConstValueExecutor`

ssddOnTop:
**Issue Reference(s):**  
Fixes #2952


src/core/jit/exec_const.rs
```diff
@@ -20,7 +21,27 @@ pub struct ConstValueExecutor {
...
+        // Create a new plan
+        let mut plan = request.create_plan(&app_ctx.blueprint)?;
+        plan.flat.retain(|f| !f.skip(variables));
+        let plan = OperationPlan::new(
+            &plan.root_name,
+            plan.flat,
+            plan.operation_type,
+            plan.index,
+            plan.is_introspection_query,
+        );
```
tusharmath, line 34: Why are recreating this object?

PR #2957, opened 2024-10-04
fix: use variance merge for ConfigModule

meskill:
**Summary:**  
Use variance merge for ConfigModule instead of MergeRight.

- for ConfigModule I changed the implemented trait from Invariant to Covariant/Contravariant to handle cases differently around settings like server/upstream. The reasoning: for multiple configs it's ok to merge such settings while for subgraph they should probably be ignored.
- had to make changes to execution_spec since the error check were executed only on the first config and not on merged config but I tried to test validation errors of merging

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/core/config/config_module/merge.rs
+++ b/src/core/config/config_module/merge.rs
@@ -7,7 +7,7 @@ use crate::core;
 use crate::core::config::{Arg, Config, Enum, Field, Type};
 use crate::core::merge_right::MergeRight;
 use crate::core::valid::{Valid, Validator};
-use crate::core::variance::{Contravariant, Covariant};
+use crate::core::variance::{Contravariant, Covariant, Invariant};
 
 impl core::Type {
     fn merge(self, other: Self, non_null_merge: fn(bool, bool) -> bool) -> Valid<Self, String> {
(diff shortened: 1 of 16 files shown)


tests/core/snapshots/test-multiple-resolvable-directives-on-field.md_error.snap
```diff
@@ -4,19 +4,8 @@ expression: errors
...
-    ],
-    "description": null
-  },
-  {
-    "message": "Multiple resolvers detected [@http, @call]",
-    "trace": [
-      "Query",
-      "user2"
-    ],
+    "message": "Validation Error\n• Multiple resolvers detected [@http, @expr] [Query, user1]\n• Multiple resolvers detected [@http, @call] [Query, user2]\n",
```
tusharmath, line 7: The error structure that separated the message from trace and description is lost over here.

src/core/config/config_module/merge.rs
```diff
@@ -179,12 +179,32 @@ impl Covariant for Enum {
...
-        let mut enums = self.config.enums;
+impl Contravariant for Config {
+    fn shrink(self, _other: Self) -> Valid<Self, String> {
+        Valid::succeed(self)
+    }
+}
+
+impl Covariant for Config {
+    fn expand(self, other: Self) -> Valid<Self, String> {
+        Valid::succeed(self.merge_right(other))
```
tusharmath, line 190: Actually Config can't have bivariance. We are implementing Variance only for types, interfaces, inputs and enums. They represent some structure of data, hence implementing these traits make sense. Config is a pure value and I would classify it as invariant.

tests/core/snapshots/test-multiple-resolvable-directives-on-field.md_error.snap
meskill: fixed

tusharmath (approved):

PR #2978, opened 2024-10-08
fix(2892): consider entity resolvers in n + 1 check

### LINKED ISSUE #2892: Handle entity resolvers in the n+1 checker
Current implementation is checking only the fields while calculating n+1 issue. But after implementing https://github.com/tailcallhq/tailcall/issues/2639 resolvers could now be defined on types as well. 

When considering entity resolver the entity resolver itself should be batched to prevent n+1 issue and that should be applied to any nested fields.

Handle resolvers on types while executing n+1 check

**Example Configuration**

```graphql
schema
  @server(port: 8000)
  @upstream(baseURL: "http://jsonplaceholder.typicode.com", httpCache: 42, batch: {delay: 100}) {
  query: Query
}

type Query {
  user(id: Int!): User @http(path: "/users/{{.args.id}}")
}

type User @http(path: "/users", query: [{key: "id", value: "{{.value.user.id}}"}], batchKey: ["id"]) {
  id: Int!
  name: String!
  username: String!
  email: String!
  phone: String
  website: String
}

type Post {
  id: Int!
  userId: Int!
  title: String!
  body: String!
  user: User @call(steps: [{query: "user", args: {id: "{{.value.userId}}"}}])
}
```

**Actual**
```bash
tailcall check -n ./config.graphql
 INFO File read: ./config.graphql ... ok
 INFO Config ./config.graphql ... ok
 INFO N + 1 detected: 1
query { posts { user } }
```

**Expected**
```bash
tailcall check -n ./config.graphql
 INFO File read: ./config.graphql ... ok
 INFO Config ./config.graphql ... ok
 INFO N + 1 detected: 2
query { posts { user } }
query { _entities { __typename: "User" } }
```

meskill:
**Summary:**  

**Issue Reference(s):**  
Fixes #2892, #3008, #3161
/claim #2892

**Build & Testing:**

**Checklist:**


src/core/config/npo/snapshots/tailcall__core__config__npo__tracker__tests__multiple_type_usage.snap
```diff
@@ -0,0 +1,8 @@
+---
+source: src/core/config/npo/tracker.rs
+expression: actual
+---
+query { t2 { t2 { n } } }
+query { t2_ls { t2 } }
+query { t3 { t3 { n } } }
+query { t3_ls { t3 { n } } }
```
meskill, line 8: on previous version it was 

```
query { t2 { t2 { n } } }
query { t2 { t2 { n } } }
query { t2_ls { t2 } }
query { t2 { t2 { n } } }
```

because of caching the whole paths

src/core/config/npo/snapshots/tailcall__core__config__npo__tracker__tests__cycles_with_resolvers.snap
```diff
@@ -2,5 +2,4 @@
 source: src/core/config/npo/tracker.rs
 expression: actual
 ---
-query { f1 { f1 { f2 } } }
 query { f1 { f2 } }
```
meskill, line 5: here one path is dropped, but it was basically the same path so if batching is added it'd resolve both branches. I don't think there is difference if we show both

src/core/config/npo/tracker.rs
```diff
@@ -91,80 +91,113 @@ impl<'a> FieldName<'a> {
...
-        if let Some(type_of) = self.config.find_type(type_name.as_str()) {
-            for (name, field) in type_of.fields.iter() {
-                let field_name = FieldName::new(name);
-                let path = path.clone().append(field_name);
-                if !visited.contains(&(type_name, field_name)) {
                     if is_list && field.has_resolver() && !field.has_batched_resolver() {
-                        chunks = chunks.append(path.clone());
+                        chunks = chunks.append(path);
                     } else {
-                        let mut visited = visited.clone();
```
meskill, line 137: visited has been cloned every time

src/core/config/npo/snapshots/tailcall__core__config__npo__tracker__tests__multiple_keys.snap
```diff
@@ -2,5 +2,5 @@
 source: src/core/config/npo/tracker.rs
 expression: actual
 ---
-query { f1 { f2 } }
 query { f1 { f3 } }
```
meskill, line 5: just the order change, I'm not sure why it's inconsistent for different test cases

karatakis (changes requested):

src/core/config/npo/fixtures/nested.graphql
```diff
@@ -0,0 +1,26 @@
+schema
+  @link(type: Config, src: "nested/a.graphql")
+  @link(type: Config, src: "nested/b.graphql")
+  @link(type: Config, src: "nested/c.graphql")
```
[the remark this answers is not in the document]
meskill: ah, those are unused links, I removed them.

I've tried first with external files, but the `include_config` doesn't resolve links so I decided just merge it manually for the test without links

src/core/config/npo/tracker.rs
```diff
@@ -91,80 +91,117 @@ impl<'a> FieldName<'a> {
...
-        chunks
+            self.cache.insert((type_name, is_list), chunks.clone());
+
+            chunks
+        };
+
+        // chunks contains only paths from the current type.
+        // Prepend every subpath with parent path
+        chunks.map(&mut |chunk| path.clone().concat(chunk.clone()))
     }
```
meskill, line 173: let me explain how the iter is implemented now. The main goal of this code is to find paths to fields from the Query root that in such paths there is some ancestor field of List type and some of the descendant uses resolver without batching.

That's basically executing a DFS for the type graph while preserving information like path and the condition if some of the ancestors is list during the execution. 

That's was already implemented this way, but the caching wasn't done right:
- the cache was storing the paths starting from the root instead of caching intermediate calculations starting from some inner node. We can cache the calculation but it should be done only on subpath level not the whole path because there could multiple ways to connect to the same type and the whole path from Query root will be different in that case. Previously it was showing duplicates with lost information, now it's fixed, see tests.
- visited was only used to prevent cycles, but the same effect could be achieved by using the cache itself if we set empty value before processing descendants and then updating it after the process. The cycles won't appear in the output.

meskill: @tusharmath please take a look once again. I have added explanation that could help

src/core/config/npo/chunk.rs
```diff
@@ -45,6 +45,14 @@ impl<A> Chunk<A> {
         vec
     }
 
+    pub fn map<B>(&self, f: &mut impl FnMut(&A) -> B) -> Chunk<B> {
+        match self {
+            Chunk::Empty => Chunk::Empty,
+            Chunk::Append(a, rc) => Chunk::Append(f(a), Rc::new(rc.map(f))),
+            Chunk::Concat(a, b) => Chunk::Concat(Rc::new(a.map(f)), Rc::new(b.map(f))),
+        }
+    }
```
tusharmath, line 54: The `map` operator isn't lazy like the rest of the operators. For very large chunks this will allocate a lot of memory.

tusharmath (approved):

src/core/config/npo/chunk.rs
meskill: The current usage is to generate the result of npo and we need this output anyway and laziness is not doing much there. 

But I'm open to the suggestions if we can implement this in a lazy way. I see the issues with Rc wrappers and reference/owned value.

src/core/config/npo/tracker.rs
```diff
@@ -92,80 +92,119 @@ impl<'a> FieldName<'a> {
...
-        chunks
+            self.cache.insert((type_name, is_list), chunks.clone());
+
+            chunks
+        };
+
+        // chunks contains only paths from the current type.
+        // Prepend every subpath with parent path
+        if let Some(path) = parent_name {
+            chunks.transform(move |chunk| Chunk::new(path).concat(chunk))
```
meskill, line 173: this doesn't work since the `'static` constraint on the closure in https://github.com/tailcallhq/tailcall-chunk/blob/9fb3a24c5340c3d340491de6ae35b6ddbac60ece/src/chunk.rs#L220 limiting usage of Borrowed data almost completely. 

I'm not sure how we can resolve it except using owned data in chunk and copy strings

meskill: fixed by using `as_vec`

karatakis (approved):

Issue #2981, opened 2024-10-09
When I query optional Union type I get error


karatakis:
### Prerequisites

- [X] I have checked the existing issues to ensure this is not a duplicate.
- [X] I am running the latest version of Tailcall.

### Describe the bug

Given the schema and the query mentioned bellow I get the following error

```json
{
  "data": null,
  "errors": [
    {
      "message": "IOException: Value expected to be object",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```

### Steps to reproduce

**Query**
```gql
{
  nodes {
    ... on Page {
      id
      slug
    }
    ... on User {
      id
      username
    }
  }
}
```
**Schema/Config**
```gql
schema @server(port: 8003) @upstream(baseURL: "http://localhost:8083") {
  query: Query
}

type Query {
  nodes: Node @expr(body: null)
}

union Node = User | Page

type User {
  id: ID!
  username: String!
}

type Page {
  id: ID!
  slug: String!
}
```

### Expected behavior

Return null

### Actual behavior

I got an error

karatakis:
not present in new discriminator

PR #2982, opened 2024-10-09
fix(discriminator): union nullable

karatakis:
**Summary:**  
See the bug report https://github.com/tailcallhq/tailcall/issues/2981

**Issue Reference(s):**  
Fixes https://github.com/tailcallhq/tailcall/issues/2981

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/core/ir/discriminator.rs
+++ b/src/core/ir/discriminator.rs
@@ -1234,17 +1234,15 @@ mod tests {
         assert_eq!(
             discriminator
                 .resolve_type(&Value::from_json(json!("string")).unwrap())
-                .unwrap_err()
-                .to_string(),
-            "Value expected to be object"
+                .unwrap(),
+            "Test"
         );
 
         assert_eq!(
             discriminator
                 .resolve_type(&Value::from_json(json!(25)).unwrap())
-                .unwrap_err()
-                .to_string(),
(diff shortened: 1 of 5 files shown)


tests/execution/test-union-optional.md
meskill, line 1: let's update the test with list case as well

```graphql
type Query {
  node: Node @expr(body: null)
  nodes: [Node] @expr(body: [{id: 1, username: "user"}, null, {id: 2, slug: "page"}])
}
```

src/core/ir/discriminator.rs
```diff
@@ -238,7 +244,7 @@ impl Discriminator {
 
     pub fn resolve_type(&self, value: &Value) -> Result<&str> {
         let Value::Object(obj) = value else {
-            bail!("Value expected to be object");
+            return Ok(&self.name);
```
meskill, line 247: I think it could hide some errors when we are trying to resolve object type but getting something else. Null is a special case though, what about return `Option<&str>` here and for null use `None`?

meskill (changes requested):

karatakis:
Closing because it is covered from the refactor here https://github.com/tailcallhq/tailcall/pull/3000

PR #3001, opened 2024-10-12
fix: `objects` in args

### LINKED ISSUE #2999: feat: raise validation error when object is passed in query.
```graphql
type Query {
  findEmployees(criteria: Nested): [Employee!]!
    @http(
      baseURL: "http://localhost:8081"
      path: "/family/employees"
      query: [
        { key: "nested", value: "{{.args.criteria}}", skipEmpty: true }
      ]
    )
}

input Nested {
  maritalStatus: MaritalStatus
  hasChildren: Boolean
}
```

`nested` is an object it shouldn't be allowed inside query param, throw an validation error like `

**Expected Error**

```
Invalid query parameter type. Expected a value to be a Scalar but received an Object.
```

technical requirements:
- add unit and integration tests

ssddOnTop:
**Issue Reference(s):**  
Fixes #2999

### DIFF
--- a/src/core/blueprint/mustache.rs
+++ b/src/core/blueprint/mustache.rs
@@ -75,6 +75,10 @@ impl<'a> MustachePartsValidator<'a> {
                         return Valid::fail(format!("can't use list type '{tail}' here"));
                     }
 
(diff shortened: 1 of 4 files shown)


ssddOnTop:
closing in favour of #3086

PR #3086, opened 2024-10-29
fix: objects in args

ssddOnTop:
Issue Reference(s):
Fixes https://github.com/tailcallhq/tailcall/issues/2999


src/core/blueprint/operators/http.rs
```diff
@@ -1,18 +1,79 @@
...
+        return Scalar::is_predefined(cur_ty) || module.find_enum(cur_ty).is_some();
+    }
+    let type_ = type_.unwrap();
+
+    let cur = iter.next();
+    if cur.is_none() {
+        return Scalar::is_predefined(cur_ty) || module.find_enum(cur_ty).is_some();
+    }
+
+    let cur = cur.unwrap();
```
tusharmath, line 28: Think of an alternative design where we don't use unwrap.

src/core/blueprint/operators/apollo_federation.rs
```diff
@@ -36,8 +36,9 @@ pub fn compile_entity_resolver(inputs: CompileEntityResolver<'_>) -> Valid<IR, S
                 Resolver::Http(http) => compile_http(
                     config_module,
                     http,
+                    &Default::default(),
                     // inner resolver should resolve only single instance of type, not a list
-                    false,
```
tusharmath, line 40: We need to guarantee over here that the entity isn't a list. Why are we overriding it inside this the function?

ssddOnTop: I think there is a confusion.

Earlier the boolean was passed to check if the field `is_list`

but now I am passing the entire field and checking `field.type_of.is_list()`. In the case of federation, the boolean is replaced by `Field::default()` so `field.type_of.is_list()` will be false by default.

Now, I added the 4th parameter i.e. `is_federation` which is to skip the checks on args, as there are no args when http directive is defined on a Type.

src/core/blueprint/operators/http.rs
```diff
@@ -30,19 +87,26 @@ pub fn compile_http(
...
+                let mut ans = Valid::succeed((
+                    key_value.key,
+                    key_value.value,
+                    key_value.skip_empty.unwrap_or_default(),
+                ));
+
+                // in the case when directive is defined on type
+                // we don't have any args to check for, so we can
+                // skip this check
+                if !is_federation && !check_scalar(&mustache, config_module, field) {
```
tusharmath, line 101: Need to check for `value.a` where `a` is an object type.

src/core/blueprint/operators/http.rs
```diff
@@ -1,18 +1,75 @@
...
+    }
+    ans
+}
+
 pub fn compile_http(
     config_module: &config::ConfigModule,
     http: &config::Http,
-    is_list: bool,
+    field: &Field,
+    is_federation: bool,
```
tusharmath, line 70: We don't need `is_federation`

src/core/blueprint/operators/http.rs
```diff
@@ -1,18 +1,75 @@
...
+}
+
+// Check if args are scalar
+fn check_args(mut iter: Iter<String>, module: &ConfigModule, field: &Field) -> bool {
+    let cur = iter.next();
+    if cur.is_none() {
+        return Scalar::is_predefined(field.type_of.name())
+            || module.find_enum(field.type_of.name()).is_some();
+    }
+    let cur = cur.unwrap();
```
tusharmath, line 39: Handle these scenarios more gracefully.

src/core/blueprint/operators/http.rs
```diff
@@ -1,18 +1,75 @@
...
+    field.args.contains_key(cur)
+        && check_ty(
+            iter,
+            module,
+            field.args.get(cur).as_ref().unwrap().type_of.name(),
```
tusharmath, line 44: handle the unwrap more gracefully.

src/core/blueprint/operators/apollo_federation.rs
```diff
@@ -37,7 +37,8 @@ pub fn compile_entity_resolver(inputs: CompileEntityResolver<'_>) -> Valid<IR, S
                     config_module,
                     http,
                     // inner resolver should resolve only single instance of type, not a list
-                    false,
+                    &Default::default(),
```
tusharmath, line 40: Pass Type and Args over here. Can create a new struct — `HttpParams {type_of: Option<Type>, args: Option<Args>}`  to set the HttpParams for the compiler to work.

PR #3098, opened 2024-11-04
perf: using one RWLock per thread

tusharmath:
- **trying a new implementation for dedupe**
- **add lrwlock**
- **update dedupe with lrwlock**
- **reset the read flag**
- **use a vector instead of thread local**


tusharmath:
Performance severely degraded.

PR #3124, opened 2024-11-19
feat: support multiple resolvers on fields

### LINKED ISSUE #3120: Support multiple resolvers on fields
Currently, only one resolver at a time is supported on a field. We want to change it to allow multiple resolvers:

**DSL Changes**

```graphql
type Query {
  data: Data
    @http(url: "http://api.com/foo")
    @http(url: "http://api.com/bar")

  dataList: [Data]
    @http(url: "http://api.com/foo")
    @http(url: "http://api.com/bar")
}
```

**Runtime Changes**

- Given multiple resolvers (HTTP gRPC or GraphQL) the expectation is to perform deep merge as follows by creating a new object with the own properties of the first object merged with the own properties of the second object. If a key exists in both objects:
	
	1. and both values are objects, the two values will be recursively merged
	2. otherwise if the value is an array, the two values should be concatenated
	3. otherwise the value from the second object will be used.

- Note the `select` operation should be applied first on the HTTP and only then should it be picked up for merging.

**Technical Requirements**
- Add unit tests for this deep merge algorithm
- Add integration tests to check end to end execution flow is implemented correctly.

meskill:
**Summary:**  

**Issue Reference(s):**  
Fixes #3120

**Build & Testing:**

**Checklist:**


src/core/config/resolver.rs
```diff
@@ -53,3 +56,87 @@ impl Resolver {
         }
     }
 }
+
+#[derive(Default, Clone, Debug, PartialEq, Eq, schemars::JsonSchema)]
+pub struct Resolvers(pub Vec<Resolver>);
```
tusharmath, line 61:
```suggestion
pub struct ResolverSet(pub Vec<Resolver>);
```

I think we can use the untagged feature of serde and convert this into an enum.

```rust
enum ResolverSet {
    One(Resolver)
    Many(Vec<Resolver>)
}
```

src/core/config/resolver.rs
```diff
@@ -53,3 +56,87 @@ impl Resolver {
...
+// to provide compatibility with old tests where
+// the same resolver could be defined in multiple configs
+// leading to duplicated resolver in such tests
+impl MergeRight for Resolvers {
+    fn merge_right(self, other: Self) -> Self {
+        if other.is_empty() {
+            self
+        } else {
+            other
+        }
```
tusharmath, line 140: This is exactly how this feature is going to be extremely useful. While merging we should combine the the two resolvers.

src/core/ir/eval.rs
```diff
@@ -95,6 +96,21 @@ impl IR {
...
+                    }))
+                    .await
+                    .into_iter()
+                    .collect::<Result<_, _>>()?;
+
+                    Ok(results
+                        .into_iter()
+                        .reduce(|acc, result| acc.merge_right(result))
+                        .unwrap_or_default())
+                }
```
tusharmath, line 113:
```suggestion
                    // TODO: This is a very opinionated merge. We should allow users to customize how they would like to merge the values. In future we should support more merging capabilities by adding an additional parameter to `Merge`.
                    Ok(results
                        .into_iter()
                        .reduce(|acc, result| acc.merge_right(result))
                        .unwrap_or_default())
                }
```

src/core/ir/eval.rs
```diff
@@ -150,3 +166,72 @@ impl IR {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    mod merge {
```
tusharmath, line 174: Move tests closer to implementation or vice versa.

tusharmath:
@meskill Thanks! This is a great PR. We need to update the docs for each directive here — https://tailcall.run/docs/tailcall-dsl-graphql-custom-directives/

src/core/ir/eval.rs
karatakis: I moved it closer to implementation

src/core/config/resolver.rs
karatakis: I modified the merging operation to concat unique resolvers. Only one test broke

src/core/ir/eval.rs
meskill: It's already along the implementation. Please, note, it's a test for a `IR::Merge` not merge_right or anything

src/core/config/resolver.rs
meskill: that's what I've tried in the first place and the current implementation it's what I ended up, since we want:
- provide compatibility with previous definition where Resolver was flatten
- provide new option `resolvers: []` where it's not flatten anymore

There are limitations in serde to express this

meskill:
> @meskill Thanks! This is a great PR. We need to update the docs for each directive here — https://tailcall.run/docs/tailcall-dsl-graphql-custom-directives/

added documentation here https://github.com/tailcallhq/tailcallhq.github.io/pull/560

tusharmath (approved):

tusharmath (approved):

PR #3169, opened 2024-11-27
docs(3163): adds documentation for link type

Yash-sudo-web:
**Summary:**  
_Briefly describe the changes made in this PR._

**Issue Reference(s):**  
Fixes https://github.com/tailcallhq/tailcall/issues/3163
/claim https://github.com/tailcallhq/tailcall/issues/3163

**Build & Testing:**

- [x] I ran `cargo test` successfully.
- [x] I have run `./lint.sh --mode=fix` to fix all linting issues raised by `./lint.sh --mode=check`.

**Checklist:**

- [x] I have added relevant unit & integration tests.
- [x] I have updated the [documentation] accordingly.
- [x] I have performed a self-review of my code.
- [x] PR follows the naming convention of `<type>(<optional scope>): <title>`

[documentation]: https://github.com/tailcallhq/tailcallhq.github.io/tree/develop/docs

PR #3173, opened 2024-11-28
fix: check if graphql schema should be updated

meskill:
**Summary:**  

I've noticed that running lint fix locally produces changes in the `generated/.tailcallrc.graphql`. After investigating it I found that the changes in graphql schema is not checked i.e. it's possible to pass the CI even when the mentioned file should be updated. 

Update the `tailcall-typedefs check` in the way to check the graphql file also. 

I though it could be the issue with my local setup, but it shows update in the CI as well and the generation is broken after https://github.com/tailcallhq/tailcall/pull/3169 (not because of the pr but because the new format not handled properly by generator). Has the file been manually updated in the mentioned PR?

**Issue Reference(s):**  

**Build & Testing:**

**Checklist:**

### DIFF
--- a/generated/.tailcallrc.graphql
+++ b/generated/.tailcallrc.graphql
@@ -1027,57 +1027,39 @@ enum Method {
 
 enum LinkType {
   """
-  Points to another Tailcall Configuration file. The imported
-  configuration will be merged into the importing configuration.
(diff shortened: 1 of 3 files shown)


meskill:
updated the pr to fix the generation also

PR #3191, opened 2024-12-03
feat(expr): enhance expr with JQ functionality

karatakis:
**Summary:**  
* Enable jq syntax inside the mustache of `@expr` directives and `select` operations.
* Report template execution (runtime) errors on response
* Optimized handling of mustache: if it is read-only we use the faster Mustache implementation, otherwise we use Jq

**Broken changes:**
* On the `select` argument of `@http, ...` directives we cannot prepend the .args selector. So the users have to manually add it if they use Jq features only.

From:
```graphql
@http(..., select: {id: "{{.id | tostring }}")
```
Becomes
```graphql
@http(..., select: {id: "{{.args.id | tostring }}")
```

**Example:**
```graphql
schema
  @server(port: 8000) {
  query: Query
}

type Query {
  users: [User] @http(url: "http://jsonplaceholder.typicode.com/users")
}

type User {
  id: Int!
  posts: [Post] @http(url: "http://jsonplaceholder.typicode.com/users/{{.value.id}}/posts")
}

type Post {
  id: Int!
  title: String!
  body: String!
  bodyParts: BodyParts! @expr(body: "{{ .value.body | split(\" \") | {first: .[0], second: .[1]} }}")
}

type BodyParts {
  first: String!
  second: String!
}
```

**Build & Testing:**

**Checklist:**


karatakis:
# Benchmarks: before/after JQ Template

```graphql
schema @server(port: 8000, hostname: "0.0.0.0") {
  query: Query
}

type Query {
  args(text: String!): String! @expr(body: "{{.args.text}}")
}
```

```lua
request = function()
    local headers = {}
    headers["Content-Type"] = "application/json"

    local body = [[
    {
        "query": "query { args(text: \"Hello, World!\") }"
    }
    ]]

    return wrk.format("POST", "/graphql", headers, body)
end
```

## Before
```lua
$ wrk -t12 -c400 -d30s -s bench.lua http://127.0.0.1:8000/graphql
Running 30s test @ http://127.0.0.1:8000/graphql
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     0.98ms  704.25us   9.30ms   75.32%
    Req/Sec    32.97k     3.26k   49.32k    68.22%
  11825923 requests in 30.06s, 1.55GB read
Requests/sec: 393459.81
Transfer/sec:     52.91MB
```

## After

```lua
$ wrk -t12 -c400 -d30s -s bench.lua http://127.0.0.1:8000/graphql
Running 30s test @ http://127.0.0.1:8000/graphql
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     0.98ms    0.86ms 207.75ms   87.80%
    Req/Sec    33.21k     3.77k   52.03k    69.14%
  11910311 requests in 30.06s, 1.56GB read
Requests/sec: 396253.29
Transfer/sec:     53.28MB
```

src/core/blueprint/dynamic_value.rs
```diff
@@ -2,12 +2,13 @@ use async_graphql_value::{ConstValue, Name};
...
 
-use crate::core::mustache::Mustache;
+use crate::core::mustache::JqTemplate;
 
 #[derive(Debug, Clone, PartialEq)]
+/// This is used to express dynamic value resolver engine.
 pub enum DynamicValue<A> {
     Value(A),
-    Mustache(Mustache),
+    JqTemplate(JqTemplate),
```
tusharmath, line 11: Rename `JqTemplate` to `Mustache`.

src/core/mustache/jq_template.rs
```diff
@@ -0,0 +1,184 @@
...
+/// Used to represent a mixture of getters mustache, jq transformations and
+/// const values templates
+pub struct JqTemplate(pub Vec<JqTemplateIR>);
+
+#[derive(Debug, Clone, PartialEq, Hash)]
+/// The IR for each part of the template
+pub enum JqTemplateIR {
+    JqTransform(JqTransform),
+    Literal(String),
+    Mustache(Mustache),
```
tusharmath, line 23: JQ needs to be wrapped under mustache not the other way round.

This is valid mustache and valid JQ:
```
{{.a.b.c}}
```
When evaluated the above expression should produce the value on field `c`.

Similarly this is valid mustache and JQ:

```
{{.a.b.c | .d[].e}}
```
When evaluated it produces the a list of values stored in `e`.


This is also valid mustache/JQ however it evaluates to a `String` always

```
Foo {{.a.b.c | .d[].e}} Bar
```

This happens because the output of the JQ template is coerced into a string type because its surrounded with text on the left and right.

src/core/path.rs
```diff
@@ -76,9 +77,31 @@ impl<Ctx: ResolverContextLike> EvalContext<'_, Ctx> {
...
+                        .fold(IndexMap::new(), |mut acc, (k, v)| {
+                            acc.insert(k, v.into());
+                            acc
+                        });
+
+                    Some(ValueString::Value(Cow::Owned(
+                        async_graphql_value::ConstValue::object(arr),
+                    )))
+                }
+                "env" => Some(ValueString::Value(Cow::Owned(ctx.env_vars()))),
```
tusharmath, line 104: Doesn't look like this change is related to this PR. Should we create a separate PR for this?

PR #3206, opened 2024-12-07
feat(gen): support proto_paths for gen graphql through proto file

DLillard0:
**Summary:**  

Currently, when using the `tailcall gen` command to process proto files, the import path is fixed to a relative path, but in the [protobuf import syntax](https://protobuf.dev/programming-guides/proto3/#importing), you can specify the path to find the dependency using the `-I/--proto_path` flag. So, most proto projects are currently organized in this way, and I think we also need to support it.

```jsonc
{
  "inputs": [
    {
      "proto": {
        "src": "./proto/apps/mian.proto",
        "url": "http://localhost:8080",
        // new params
        "protoPaths": ["./proto/apps"]
      }
    }
  ],
  "preset": {
    "mergeType": 1.0
  },
  "output": {
    "path": "./output/main.graphql",
    "format": "graphQL"
  },
  "schema": {
    "query": "Query"
  }
}
```

**Issue Reference(s):**  

None

**Build & Testing:**

**Checklist:**

### DIFF
--- a/src/cli/generator/config.rs
+++ b/src/cli/generator/config.rs
@@ -81,9 +81,12 @@ pub enum Source<Status = UnResolved> {
         is_mutation: Option<bool>,
         field_name: String,
     },
+    #[serde(rename_all = "camelCase")]
     Proto {
         src: Location<Status>,
         url: String,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        proto_paths: Option<Vec<Location<Status>>>,
         #[serde(skip_serializing_if = "Option::is_none")]
         #[serde(rename = "connectRPC")]
         connect_rpc: Option<bool>,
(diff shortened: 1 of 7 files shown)


tusharmath:
Can't we use src to specify the protoPath instead of creating a new field?

DLillard0:
proto_path specifies the search path for dependent files in src file.
Should they mean different things?

src/core/proto_reader/reader.rs
```diff
@@ -248,7 +282,7 @@ mod test_proto_config {
         let reader = ProtoReader::init(ResourceReader::<Cached>::cached(runtime.clone()), runtime);
         let proto_no_pkg =
             PathBuf::from(tailcall_fixtures::configs::SELF).join("proto_no_pkg.graphql");
-        let config_module = reader.read(proto_no_pkg.to_str().unwrap()).await;
+        let config_module = reader.read(proto_no_pkg.to_str().unwrap(), None).await;
```
tusharmath, line 285: Can we add tests for reading from proto path instead of default behavior.

src/cli/generator/config.rs
```diff
@@ -81,10 +81,13 @@ pub enum Source<Status = UnResolved> {
         is_mutation: Option<bool>,
         field_name: String,
     },
+    #[serde(rename_all = "camelCase")]
     Proto {
         src: Location<Status>,
         url: String,
         #[serde(skip_serializing_if = "Option::is_none")]
+        proto_paths: Option<Vec<Location<Status>>>,
```
tusharmath, line 89: Based on your use case would it make sense to put `proto_paths` at root level in the `Config` or should it be set for each `Source::Proto`?

DLillard0: Because `inputs` can configure multiple proto entry files, I don't think it should be applied to all `Source::Proto`. We can maintain this flexibility and let users configure it themselves. I think this configuration is a bit like `connectRPC`, So I think this is reasonable.

src/core/proto_reader/reader.rs
DLillard0: I’ve added tests to `tests/cli/fixtures/generator/gen_proto_with_proto_paths_config.md` that should cover my new code.

tusharmath (approved):

PR #3222, opened 2024-12-14
feat(link): add `proto_paths` support

DLillard0:
**Summary:**  
Previously, this [pr[3206]](https://github.com/tailcallhq/tailcall/pull/3206) only processed the `proto_paths` during gen, but when using `tailcall start`, there will still be path problems, so it is necessary to save the` proto_paths` information on the link to ensure that the path can be processed correctly during `tailcall start`.

Config:
```json
{
  "inputs": [
    {
      "proto": {
        "src": "news.proto",
        "url": "http://localhost:50051",
        "protoPaths": ["tailcall-fixtures/fixtures/protobuf"]
      }
    }
  ],

  ...
}
```

Output:
```graphql
schema
  @server
  @upstream
  @link(
    src: "news.proto"
    # new attr
    proto_paths: ["tailcall-fixtures/fixtures/protobuf"]
    type: Protobuf
  ) {
  query: Query
}
...
```

**Issue Reference(s):**  
None

**Build & Testing:**

**Checklist:**

### DIFF
--- a/generated/.tailcallrc.schema.json
+++ b/generated/.tailcallrc.schema.json
@@ -269,6 +269,16 @@
         "meta": {
           "description": "Additional metadata pertaining to the linked resource."
         },
+        "proto_paths": {
+          "description": "The proto paths to be used when resolving dependencies. Only valid when [`Link::type_of`] is [`LinkType::Protobuf`]",
+          "type": [
+            "array",
+            "null"
+          ],
+          "items": {
+            "type": "string"
+          }
+        },
         "src": {
           "description": "The source of the link. It can be a URL or a path to a file. If a path is provided, it is relative to the file that imports the link.",
           "type": "string"
--- a/src/core/config/directives/link.rs
+++ b/src/core/config/directives/link.rs
@@ -93,4 +93,9 @@ pub struct Link {
     /// Additional metadata pertaining to the linked resource.
     #[serde(default, skip_serializing_if = "is_default")]
     pub meta: Option<serde_json::Value>,
+    ///
+    /// The proto paths to be used when resolving dependencies.
+    /// Only valid when [`Link::type_of`] is [`LinkType::Protobuf`]
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub proto_paths: Option<Vec<String>>,
 }
--- a/src/core/config/reader.rs
+++ b/src/core/config/reader.rs
@@ -79,7 +79,13 @@ impl ConfigReader {
                     });
                 }
                 LinkType::Protobuf => {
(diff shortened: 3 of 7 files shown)


tusharmath:
@DLillard0 Thanks for the PR! Our goal ultimately is to drop `link` from the GraphQL configuration. I have talked about it in detail over here - https://tailcall.run/blog/migrating-to-graphql-configuration-v2/


We basically two options now:
1. Generate a file with just the schema information without any links.
2. Generate two files: a main.graphql and a main.yml (for links).

I prefer Option 1 because its a lot simpler to work with. Let me know what you think?

DLillard0:
@tusharmath I also agree with Option 1. If there is a real need for modular configuration management, can we also implement the current link functionality by providing an additional library? This will decouple it from the current library and reduce the current complexity.
Is there a timeline for removing the `link` from the GraphQL configuration?

tusharmath:
> I also agree with Option 1. If there is a real need for modular configuration management, can we also implement the current link functionality by providing an additional library?

It's always nice to organize configurations into smaller files. In large organizations, these files are owned and maintained by individual teams. Link functionality still exists however it is only available in the `YML/JSON` format.

> Is there a timeline for removing the link from the GraphQL configuration?
This feature has already been [shipped](https://github.com/tailcallhq/tailcall/releases/tag/v1.3.0), you can try it out.

DLillard0:
@tusharmath I have two more questions. 
1. Is there any way to generate a  `YML/JSON` configuration file? like `tailcall gen`?
2. Moving the Link functionality to the configuration file also seems to require dealing with the path problem of GraphQL dependencies. It seems that Link still needs to support proto_paths.

tusharmath:
> Is there any way to generate a YML/JSON configuration file? like tailcall gen?

Not via `tailcall gen`. But I don't think these are very complicated to hand-write.



> Moving the Link functionality to the configuration file also seems to require dealing with the path problem of GraphQL dependencies. It seems that Link still needs to support proto_paths.

You are right, we need to still update `config::link.rs` to support `proto_path`. However, I think the generated file should not contain any `@link` directives.

DLillard0:
Thanks for your answer!

DLillard0:
I deleted the generate related code and only left the support for `proto_paths` in parsing `link config`. Is this ok? Or do you have plans to support `proto_paths` in another way?

tusharmath (approved):

tusharmath:
Thanks @DLillard0 Can you update the docs with the `proto_paths` information also - https://github.com/tailcallhq/tailcallhq.github.io

DLillard0:
Ok, I'll add more information later.

PR #3237, opened 2025-01-03
feat: make dedupe faster at I/O level

### LINKED ISSUE #3028: feat: make dedupe faster at I/O level.
Currently, for all deduplicated requests, we create a clone of the response, but we don't modify the cloned response. This results in redundant copies in memory and unnecessary overhead.

## Proposed Solution:
Instead of cloning the response, we can keep a single copy of the response in memory and use it to construct the final response in synth. This approach will optimize both memory usage and performance by reducing duplication and overhead.

A similar approach has been implemented in [PR #3025](https://github.com/tailcallhq/tailcall/pull/3025).

alpha-ulrich:
**Summary:**  
Uses [Arc](https://doc.rust-lang.org/std/sync/struct.Arc.html) to ensure that dedupe requests do not generate new clones of the response, but rather point to the same response using the Atomically referenced counter.

**Issue Reference(s):**  
Fixes #3028 

/claim #3028 

**Build & Testing:**

**Checklist:**

- [x] I have added relevant unit & integration tests. [Modified relevant tests after code changes]
- [x] I have updated the [documentation] accordingly. [No update required for documentation]

### DIFF
--- a/src/core/data_loader/dedupe.rs
+++ b/src/core/data_loader/dedupe.rs
@@ -8,8 +8,8 @@ use tokio::sync::broadcast;
 pub trait Key: Send + Sync + Eq + Hash + Clone {}
 impl<A: Send + Sync + Eq + Hash + Clone> Key for A {}
 
-pub trait Value: Send + Sync + Clone {}
-impl<A: Send + Sync + Clone> Value for A {}
+pub trait Value: Send + Sync {}
+impl<A: Send + Sync> Value for A {}
 
 ///
 /// Allows deduplication of async operations based on a key.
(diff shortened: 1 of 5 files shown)


tusharmath:
@alpha-ulrich Thanks!
Can you add some e2e benchmarks to see if there is an observed difference in performance? We eventually end up cloning the data in the PR, so I am curious to see if Arc is making a difference in performance.

alpha-ulrich:
@tusharmath sure, I'll see what can be done, later today.
I'll have a look at the [graphql-benchmarks](https://github.com/tailcallhq/graphql-benchmarks) repository, to see if I can reuse some parts from there to accomplish this.

If there are any specific metrics that you would want reported, do let me know.

tusharmath:
check the `/benchmarks` folder.

Issue #3384, opened 2025-05-04
tailcall CLI command not recognized


ASafdari:
Hi, I'm trying to get started using Tailcall by following the documentation page. When running "npm i -g @Tailcallhq/tailcall" the installation is succesful however, when running "tailcall" the command is not recognized. I've  restarted my computer to make sure but no luck. When looking through the  global node_modules folder there is no executable for tailcall so it makes sense I  guess. I'm on Windows 11.

Kaushal-26:
I think for windows, we require MSVC to run it.

PR #3386, opened 2025-05-06
fix(3384): improve Windows installation with MSVC detection and better errors

luffy-orf:
### Fixes issue #3384 where Tailcall CLI isn't recognized on Windows after installation.

- Adds MSVC detection for Windows installations
- Improves error handling with descriptive messages
- Adds `process.exit(1)` on installation failure
- Updates README with Windows requirements

/claim #3384

---

### Build & Testing:

### Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ If you're on Windows, you'll need to install:
 
 1. Microsoft Visual C++ Build Tools - [Download here](https://visualstudio.microsoft.com/visual-cpp-build-tools/)
 2. Make sure your Node.js architecture (x86/x64) matches your Windows architecture
+3. Run the installation command with administrator privileges
 
 ### Yarn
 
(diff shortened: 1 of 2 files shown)


tusharmath:
I am not sure if this will solve the issue. The problem is that the correct binary is not being installed. If msvc was a problem, it would have installed but not executed.

luffy-orf:
> I am not sure if this will solve the issue. The problem is that the correct binary is not being installed. If msvc was a problem, it would have installed but not executed.

I can see the Windows binaries need the .exe extension and there's special handling for this in the build process. Then i will add checks to verify the actual binary exists after installation

luffy-orf:
@tusharmath I have added binary verification for windows

